PR #4, opened 2025-10-01
fix!: fixed error handling

pavelgj:
- **BREAKING CHANGE**: The `response` future returned by the `.stream()` method is now nullable (`Future<O?>`). This change supports improved error handling.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -19,487 +19,133 @@ import 'package:http/http.dart' as http;
 import 'exception.dart';
 import 'flow_response.dart';
 
-/// {@template remote_action}
-/// Represents a remote Genkit action (flow) that can be invoked or streamed.
-///
-/// This class is typically instantiated via [defineRemoteAction].
-/// It encapsulates the URL, default headers, HTTP client, and data conversion logic
-/// for a specific flow.
-///
-/// Type parameters for methods:
-///   - `I`: The type of the input data for the flow (inferred at call/stream time).
-///   - `O`: The type of the output data from a non-streaming flow invocation,
-///          or the type of the final response from a streaming flow (defined by fromResponse).
-///   - `S`: The type of the data chunks streamed from the flow (defined by fromStreamChunk).
-/// {@endtemplate}
-class RemoteAction<O, S> {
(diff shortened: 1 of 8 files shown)


lib/src/client.dart
```diff
@@ -19,26 +19,195 @@ import 'package:http/http.dart' as http;
...
+        },
+        cancelOnError: true,
+      );
+  onSubscription?.call(subscription);
+  setCancelCallback?.call(() {
+    if (!responseCompleter.isCompleted) {
+      responseCompleter.complete(null);
+    }
+  });
+
```
nozomi-koborinai, line 147:
```dart 
unawaited( 
  responseCompleter.future.catchError( 
    (error, stackTrace) => Future<O>.error(error, stackTrace), 
  ), 
); 
```

If we drop that in right before returning the future (and mirror it in RemoteAction.stream() right after streamFlow(...)), the future should stay “observed” even when a caller skips await, while callers that do await it still receive the same rejection. feels closer to what the JS client does—both handles surface the error—without introducing Future<O?>. What do you think?

lib/src/client.dart
```diff
@@ -19,26 +19,195 @@ import 'package:http/http.dart' as http;
...
+              GenkitException('Stream finished without a final result chunk.'),
+            );
+          }
+        },
+        cancelOnError: true,
+      );
+  onSubscription?.call(subscription);
+  setCancelCallback?.call(() {
+    if (!responseCompleter.isCompleted) {
+      responseCompleter.complete(null);
```
nozomi-koborinai, line 144:
```dart
responseCompleter.completeError(GenkitException('Stream cancelled.'));
```

lib/src/client.dart
pavelgj: unfortunately this doesn't seem to work... I already tried so many other options. :(
I think I'm going to merge it like this -- need working error handling. We can iterate at trying to make output throw properly...

lib/src/client.dart
pavelgj: unfortunately this causes a separate uncaught exception during cancellation... so part of the same problem.

PR #5, opened 2025-10-07
feat(stream)!: Replace FlowStreamResponse with GenkitStream

pavelgj:
This commit introduces a breaking change to the streaming API to simplify its usage.

The `.stream()` method on a `RemoteAction` now returns a `GenkitStream` object directly, instead of a `FlowStreamResponse` record.

`GenkitStream` is a `Stream` subclass, so it can be used directly in `await for` loops. It also provides a `finalResult` property to access the flow's final, non-streamed response, replacing the previous `response` future from the record.

This change makes the API more intuitive and reduces boilerplate when handling streaming flows.

BREAKING CHANGE: The `.stream()` method now returns a `GenkitStream` instead of a `FlowStreamResponse` record (`({Future<O?> response, Stream<S> stream})`).

Migration:
- Replace `final (:stream, :response) = myAction.stream(...)` with `final stream = myAction.stream(...)`.
- Access the final result via `stream.finalResult` instead of `await response`.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -17,7 +17,6 @@ import 'dart:convert';
 import 'package:http/http.dart' as http;
 
 import 'exception.dart';
-import 'flow_response.dart';
 
 const _flowStreamDelimiter = '\n\n';
 
@@ -288,7 +287,7 @@ class RemoteAction<O, S> {
   }
 
   /// Invokes the remote flow and streams its response.
-  FlowStreamResponse<O?, S> stream<I>({
+  GenkitStream<S, O> stream<I>({
     required I input,
     Map<String, String>? headers,
   }) {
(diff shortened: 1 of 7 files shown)


lib/src/client.dart
```diff
@@ -357,3 +352,9 @@ class RemoteAction<O, S> {
     }
   }
 }
+
+class GenkitStream<S, F> extends StreamView<S> {
+  F? finalResult;
```
yjbanov: I'd make this non-null, and throw when called if:

* The stream hasn't completed yet.
* The stream completed with an error.

If the stream has completed successfully, the API could guarantee the availability of a non-null `F`.

This field is publicly read-write. Is that intentional? I think you want to hide the setter.

Also nit: do you need the "final" in the name?

lib/src/client.dart
```diff
@@ -357,3 +352,9 @@ class RemoteAction<O, S> {
     }
   }
 }
+
+class GenkitStream<S, F> extends StreamView<S> {
```
yjbanov: For users who do not care about the `Stream` and just want the result, you could also have a convenience getter `Future<F> get onResult`, which drains the stream and returns the result, throwing like a normal future would.

lib/src/client.dart
pavelgj: Done.

lib/src/client.dart
pavelgj: Good idea. Done.

PR #35, opened 2026-01-16
chore: update readmes

kevmoo:


kevmoo:
A title in a readme is redundant on the pub site.
Added pub links everywhere
Dropped duplicate content in the root readme
Added a table

PR #47, opened 2026-01-19
chore: enable and fix a number of helpful lints

kevmoo:


packages/genkit/lib/src/client/flow_response.dart
```diff
@@ -1,23 +0,0 @@
-// Copyright 2024 Google LLC
```
kevmoo, line 1: wasn't used any more!

PR #49, opened 2026-01-19
chore: move to relative imports

kevmoo:


packages/genkit/lib/src/core/action.dart
```diff
@@ -14,11 +14,11 @@
 
 import 'dart:async';
 
-import 'package:genkit/client.dart';
```
kevmoo, line 17: easier to see these bits!

PR #58, opened 2026-01-23
Suggestions

mosuem:
- Add license header via generation instead of extra script
- Implement `anyOf` using an annotation and typed constructors. Typed getters don't make sense IMO, as the user will need to switch on the output anyhow. Happy to be convinced otherwise.
- Use `$` prefix instead of `Schema` suffix, IMO this improves readability and makes it easier to immediately recognize which classes are just an input to the generator and which are to be used.
- Make an analyzer plugin to warn about missing `$` prefixes.

TODOs for a future PR:
- Get the analyzer plugin to work - the analysis server seems to crash when enabling the plugin. For now, it's disabled.
- Implement `@protected` to be auto-applied for `@Schematic`?

### DIFF
--- a/packages/schemantic_analyzer_plugin/lib/main.dart
+++ b/packages/schemantic_analyzer_plugin/lib/main.dart
@@ -109,20 +109,14 @@ class AddSchematicPrefix extends ResolvedCorrectionProducer {
       await builder.addDartFileEdit(file, (builder) {
         // 1. Handle the '$' prefix
         if (!name.startsWith(r'$')) {
-          builder.addInsertion(
-            nameToken.offset,
-            (builder) => builder.write(r'$'),
-          );
+          builder.addSimpleInsertion(nameToken.offset, r'$');
         }
 
         // 2. Handle the 'abstract' keyword
         // The lint enforces both. If 'abstract' is missing, we insert it.
         if (classDeclaration.abstractKeyword == null) {
           final classKeyword = classDeclaration.classKeyword;
-          builder.addInsertion(
-            classKeyword.offset,
-            (builder) => builder.write('abstract '),
-          );
+          builder.addSimpleInsertion(classKeyword.offset, 'abstract ');
         }
       });
     }


packages/schemantic/lib/builder.dart
```diff
@@ -18,5 +18,6 @@ import 'package:source_gen/source_gen.dart';
 import 'src/schema_generator.dart';
 
 Builder schemaBuilder(BuilderOptions options) {
-  return PartBuilder([SchemaGenerator()], '.schema.g.dart');
+  final header = options.config['header'] as String?;
+  return PartBuilder([SchemaGenerator()], '.schema.g.dart', header: header);
```
kevmoo: `PartBuilder` has a HEADER????

Ugh. This is much better. I'm embarrassed.

mosuem: :upside_down_face:

packages/schemantic_analyzer_plugin/lib/main.dart
```diff
@@ -0,0 +1,130 @@
...
+    final classDeclaration = node.thisOrAncestorOfType<ClassDeclaration>();
+
+    if (classDeclaration != null) {
+      final nameToken = classDeclaration.namePart.typeName;
+      final name = nameToken.lexeme;
+
+      await builder.addDartFileEdit(file, (builder) {
+        // 1. Handle the '$' prefix
+        if (!name.startsWith(r'$')) {
+          builder.addInsertion(
```
srawlins: I believe this statement can be replaced a call to `addSimpleInsertion`:

```dart
builder.addSimpleInsertion(nameToken.offset, r'$');
```

packages/schemantic_analyzer_plugin/lib/main.dart
```diff
@@ -0,0 +1,130 @@
...
+            nameToken.offset,
+            (builder) => builder.write(r'$'),
+          );
+        }
+
+        // 2. Handle the 'abstract' keyword
+        // The lint enforces both. If 'abstract' is missing, we insert it.
+        if (classDeclaration.abstractKeyword == null) {
+          final classKeyword = classDeclaration.classKeyword;
+          builder.addInsertion(
```
srawlins: Same here:

```dart
builder.addSimpleInsertion(classKeyword.offset, 'abstract ');
```

PR #60, opened 2026-01-29
feat!: implement Schemantic API redesign with $ prefixed schema definitions and static `$schema` for unified schema access.

pavelgj:
Schemantic API Changes

## 1. Schema Definitions: $ Prefix
Old: abstract class UserSchema New: abstract class $User

All schema classes must now start with `$` (e.g., `$User`, `$Input`, `$Output`).
The generated class will strip the $ (e.g., `$User` -> `User`).
The `*Schema` suffix naming convention is no longer supported/special.

## 2. Instantiation: Default Constructor
Old: `User.from(name: 'Alice')` New: `User(name: 'Alice')`

The generated class now uses the default constructor for creating instances.

## 3. JSON Parsing: Named Factory
Old: `User(jsonMap)` New: `User.fromJson(jsonMap)`

The default constructor `User(this._json)` has been privatized to `User._(this._json)`.
You must now use `User.fromJson` for parsing Maps.

## 4. Schema Access: Static Field
Old: `UserType.parse(...)`, `UserType.jsonSchema() `New: `User.$schema.parse(...)`, `User.$schema.jsonSchema`

The `*Type` utility class (e.g., `UserType`) is no longer generated as a public class.
Access schema metadata and utilities via the static `$schema` field on the generated class.

### DIFF
--- a/packages/genkit/lib/src/core/reflection.dart
+++ b/packages/genkit/lib/src/core/reflection.dart
@@ -301,8 +301,7 @@ class ReflectionServer {
   }
 
   Future<void> _handleRunAction(HttpRequest request) async {
-    final body =
-        jsonDecode(await utf8.decodeStream(request)) as Map<String, dynamic>;
+    final body = await _jsonDecodeStream(request);
     final key = body['key'] as String;
     final input = body['input'];
     final stream = request.uri.queryParameters['stream'] == 'true';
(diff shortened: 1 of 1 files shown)


packages/genkit_firebase_ai/lib/genkit_firebase_ai.dart
```diff
@@ -145,7 +145,9 @@ class _FirebaseGenAiPlugin extends GenkitPlugin {
             responseJsonSchema:
                 req.config?['responseJsonSchema'] as Map<String, dynamic>?,
             thinkingConfig: req.config?['thinkingConfig'] != null
-                ? m.ThinkingConfig(
+                ?
+                  // ignore: deprecated_member_use
```
kevmoo, line 149: @pavelgj – just to make analyzer happy. Not exactly sure what to do here to migrate

PR #76, opened 2026-01-31
samples: introduce a new test application demonstrating various agentic patterns in Dart.

pavelgj:


pubspec.yaml
```diff
@@ -11,6 +11,7 @@ workspace:
   - packages/genkit_google_genai
   - packages/genkit_shelf
   - packages/schemantic
+  - testapps/agentic-patterns
```
kevmoo: Update the bits at the end of the file ensuring build is ran w/ melos!

pubspec.yaml
```diff
@@ -55,5 +55,5 @@ melos:
     build-gen-testapps:
       exec: dart run build_runner build
       packageFilters:
-        scope: "{ai_features,basic_sample}"
+        scope: "{agentic-patterns,ai_features,basic_sample}"
```
kevmoo: the `packageFilters` is the package name – a hyphen won't work I don't think. Double check that it runs!

pavelgj: good catch. renamed the folder.

PR #84, opened 2026-02-03
undo core

pavelgj:


packages/genkit/lib/src/utils_web.dart
```diff
@@ -0,0 +1,23 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+String? getEnvVar(String name) {
+  if (Uri.base.queryParameters.containsKey(name)) {
+    return Uri.base.queryParameters[name];
```
kevmoo, line 17: Again, we should be careful here. This would leak into any production app using this. Should we just punt on it now?

pavelgj: I renamed it to `getConfigVar` and I'm not too worried -- it's only used for telemetry stuff which should be compiled out in prod.

PR #95, opened 2026-02-10
feat(plugins): add OpenAI plugin

cabljac:
This PR adds an OpenAI-compatible plugin, following the patterns of this repo and the existing typescript plugin of the same vein.

It also adds a basic test app for the plugin.

## Testing
- [x] Listing of models
image

- [x] Using a model for chat
image

- [x] Multi-turn chat
image

- [x] System prompt
image

- [ ] Images (does not work)
image

- [x] Stream chat
image

- [x] Genkit Tools
image

- [ ] Web Search Tool (does not work)

- [x] JSON
image

- [x] Config max tokens
image

### DIFF
--- a/packages/genkit_openai/build.yaml
+++ b/packages/genkit_openai/build.yaml
@@ -14,7 +14,7 @@ targets:
             //     http://www.apache.org/licenses/LICENSE-2.0
             //
             // Unless required by applicable law or agreed to in writing, software
-            // distributed on an "AS IS" BASIS,
+            // distributed under the License is distributed on an "AS IS" BASIS,
             // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
             // See the License for the specific language governing permissions and
             // limitations under the License.


testapps/openai_sample/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
+name: openai_compat_sample
+publish_to: none
+description: Genkit Dart OpenAI compatibility test application
+resolution: workspace
+
+environment:
+  sdk: ^3.10.0
```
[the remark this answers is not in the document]
cabljac: Screenshot 2026-02-10 at 16 12 00

Not sure that's correct? But worth checking what versions we should support

packages/genkit_openai_compat/lib/src/converters.dart
```diff
@@ -0,0 +1,207 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:convert';
+
+import 'package:genkit/genkit.dart';
+import 'package:openai_dart/openai_dart.dart';
+
+/// Convert Genkit messages to OpenAI format
+List<ChatCompletionMessage> toOpenAIMessages(
```
kevmoo: I wonder if most of these would be better as extensions – just for readability

@pavelgj ?

cabljac: I think yeah we're planning to change it to a class based method rather than a global method

cabljac: @CorieW

testapps/openai_sample/lib/openai_flows.dart
```diff
@@ -0,0 +1,184 @@
...
+  print('Assistant: ${response.text}\n');
+}
+
+/// Run all examples
+Future<void> runAllExamples(String apiKey) async {
+  await simpleGenerateExample(apiKey);
+  await streamingExample(apiKey);
+  await toolCallingExample(apiKey);
+  await conversationExample(apiKey);
+}
```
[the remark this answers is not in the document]
cabljac: I don't think calling them is necessary for them to be exposed in the dev UI for this test app.

PR #140, opened 2026-02-19
fix(genkit core): don't set streamingRequested when onChunk is not provided

cabljac:
When generate() is called with outputSchema but without onChunk, it incorrectly sets streamingRequested: true in the model plugin context, because the onChunk passed to generateHelper was always a non-null lambda. This forces model plugins into their streaming code path unnecessarily. The fix only passes onChunk when the caller actually provided one.

### DIFF
--- a/packages/genkit/test/genkit_test.dart
+++ b/packages/genkit/test/genkit_test.dart
@@ -330,5 +330,44 @@ void main() {
         );
       },
     );
+
+    test(
+      'generate with outputSchema and onChunk should request streaming',
+      () async {
+        const modelName = 'streamingOutputModel';
+
+        bool? wasStreamingRequested;
+
+        genkit.defineModel(
+          name: modelName,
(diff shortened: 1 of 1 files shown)


packages/genkit/test/genkit_test.dart
```diff
@@ -286,5 +292,43 @@ void main() {
       expect(receivedChunks[1].text, 'chunk2');
       expect(result.text, 'final response');
     });
+
+    test(
```
CorieW, line 296: is there an opposite test anywhere? So, checking that `streamingRequested` is `true`?

cabljac: fair point, will add one :)

pavelgj (approved):

cabljac:
@pavelgj  I think this is now fixed in main from other changes? closing this

PR #151, opened 2026-02-21
fix: fix a couple of dartdoc issues

kevmoo:
Also include a tool to check all of the docs for `dart doc` issues


.github/workflows/dart_ci.yml
```diff
@@ -43,11 +43,8 @@ jobs:
...
+
       - name: Verify clean git tree
-        run: |
-          git diff --exit-code
-          if [ -n "$(git status --porcelain)" ]; then
-            echo "Git directory is not clean. Working tree:"
-            git status
-            exit 1
-          fi
+        run: '[ -z "$(git status --short)" ] || (git status --short && exit 1)'
```
kevmoo, line 50: This is a cleaner way to do the same

PR #154, opened 2026-02-23
feat!: move basic type functions to static creation method on SchemanticType

natebosch:
Top level methods in Dart crowd a global namespace, and do not offer
useful auto-complete help until a few characters are present. With
static method the `.` triggers an autocomplete on a narrow surface area
of static methods for smoother editing.

At usage sites with no context type this will make the usage strictly
longer - `stringSchema` becomes `SchemanticType.string`. However at
usage sites with a context type, such as the arguments to `Action`, the
`SchemanticType` prefix can be omitted which leads to both terse and
auto-completable methods.

- Add static methods for each method from `basic_types.dart`. Where
  these don't conflict with Dart keywords omit the `Schema` suffix and
  include it where required for disambiguation.
- Remove export of old members.

### DIFF
--- a/testapps/agentic_patterns/bin/main.dart
+++ b/testapps/agentic_patterns/bin/main.dart
@@ -28,16 +28,14 @@ import 'package:genkit_google_genai/genkit_google_genai.dart';
 
 void main(List<String> arguments) async {
   // Initialize Genkit and Model
-  final ai = Genkit(
-    plugins: [
-      googleAI(),
-    ],
-  );
+  final ai = Genkit(plugins: [googleAI()]);
   final geminiFlash = googleAI.gemini('gemini-2.5-flash');
 
   // Initialize Flows
-  final iterativeRefinementFlow =
-      defineIterativeRefinementFlow(ai, geminiFlash);
+  final iterativeRefinementFlow = defineIterativeRefinementFlow(
+    ai,
+    geminiFlash,
+  );
   final storyWriterFlow = defineStoryWriterFlow(ai, geminiFlash);
   final marketingCopyFlow = defineMarketingCopyFlow(ai, geminiFlash);
   final routerFlow = defineRouterFlow(ai, geminiFlash);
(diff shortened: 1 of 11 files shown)


packages/genkit/example/flow_basics.dart
```diff
@@ -38,8 +38,8 @@ void main() async {
   // genkit flow:run basic "\"hello\""
   final basic = ai.defineFlow(
     name: 'basic',
-    inputSchema: stringSchema(),
-    outputSchema: stringSchema(),
+    inputSchema: .string(),
```
pavelgj, line 41: What is this sorcery? How does this work...

parlough: [Dot shorthands](https://dart.dev/language/dot-shorthands) introduced in Dart 3.10 :D

The syntax essentially allows you to omit the [context type](https://dart.dev/glossary/context-type) when it can be inferred.

pavelgj: Quick google search resolved my ignorance! I like it.

packages/schemantic/lib/src/basic_types.dart
```diff
@@ -23,6 +23,7 @@ import '../schemantic.dart';
 /// ```dart
 /// nullable(stringSchema()).parse('hello');
 /// ```
+@Deprecated('Use SchemanticType.nullable')
```
pavelgj: let's not do this... nothing is released, so let's not introduce deprecated APIs. Can you just switch the repo to the new syntax?

natebosch: Done. Typically I'd recommend modeling it as a semver breaking change, but I see that we've been running with `0.0.1-dev` versions including breaking changes so no strong reason to stop now.

I had thought this was required for the `dart fix` definitions to work, but it turns out to still work 😄

pavelgj (approved):

PR #157, opened 2026-02-24
feat(openai): add structured outputs

CorieW:
Fixes https://github.com/genkit-ai/genkit-dart/issues/136

This adds both streaming and non-streaming structured outputs and a testapp with flows for testing either scenario.

**Testing:**
- [x] Non-streaming
image

- [x] Streaming
image

### DIFF
--- a/packages/genkit_openai/lib/src/openai_plugin.dart
+++ b/packages/genkit_openai/lib/src/openai_plugin.dart
@@ -14,6 +14,7 @@
 
 import 'package:genkit/plugin.dart';
 import 'package:openai_dart/openai_dart.dart' hide Model;
+import 'package:schemantic/schemantic.dart';
 
 import '../genkit_openai.dart';
 import 'aggregation.dart';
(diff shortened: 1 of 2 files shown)


packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -12,39 +12,37 @@
...
-    return id.isNotEmpty && name.isNotEmpty;
-  }
+/// Builds an OpenAI [ResponseFormat] from a Genkit output schema that uses
+/// `$defs` and a top-level `$ref`. Returns null if [schema] is null or
+/// has no definitions (empty `$defs`).
+ResponseFormat? buildOpenAIResponseFormat(Map<String, dynamic>? schema) {
+  if (schema == null) return null;
+  final defs = schema[r'$defs'] as Map<String, dynamic>? ?? {};
+  if (defs.isEmpty) return null;
+  final schemaShell = defs.values.first as Map<String, dynamic>;
```
cabljac: Just want to double check here

E.g if schema was like:

```
{
  "$ref": "#/$defs/Company",
  "$defs": {
    "Address": {
      "type": "object",
      "properties": { "street": { "type": "string" } }
    },
    "Company": {
      "type": "object",
      "properties": { 
        "name": { "type": "string" },
        "location": { "$ref": "#/$defs/Address" }
      }
    }
  }
}
```

it will grab just `Address` right, and `Company` will not be sent? hmm need to think, just noting here for a sec

cabljac: I _think_ we should be fine just passing the schema through

cabljac: After discussion we'll probably use schemantic `flatten` extension method

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -12,39 +12,35 @@
...
+  if (schema == null) return null;
+  final flattened = Schema.fromMap(schema).flatten().value;
+  return ResponseFormat.jsonSchema(
+    jsonSchema: JsonSchemaObject(
+      name: 'output',
+      schema: {
+        ...flattened,
+        'additionalProperties': false,
+      },
+      strict: true,
```
cabljac, line 41: I've just thought - 

does `flatten` definitely add `additionalProperties` at every level?

cabljac: How vital is `strict: true` here? that's why we have `additionalProperties: false` right?

cabljac: @CorieW worth adding tests maybe:

- a schema deeply nested that itself doesn't have additionalProperties false
- also something with multiple refs and a def just to assert we're handling that case fine

```
{
  "$ref": "#/$defs/Company",
  "$defs": {
    "Address": { ... },
    "Company": { ... }
  }
}
```

a test case like that

PR #163, opened 2026-02-25
feat(openai): initial draft of deepseek support

cabljac:
Resolves https://github.com/genkit-ai/genkit-dart/issues/133

This PR adds deepseek support on openai

Note: still a draft, yet to test or compare to js approach.

### DIFF
--- a/packages/genkit_openai/lib/src/openai_plugin.dart
+++ b/packages/genkit_openai/lib/src/openai_plugin.dart
@@ -50,19 +50,22 @@ class _ToolCallAccumulator {
 /// Core plugin implementation
 class OpenAIPlugin extends GenkitPlugin {
   @override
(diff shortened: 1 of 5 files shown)


packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -12,53 +12,60 @@
...
-  String get name => 'openai';
+  final String name;
 
   final String? apiKey;
   final String? baseUrl;
   final List<CustomModelDefinition> customModels;
   final Map<String, String>? headers;
+  final ModelInfo Function(String)? modelInfoResolver;
 
   OpenAIPlugin({
```
cabljac, line 61: don't think this is correct approach

PR #166, opened 2026-02-25
fix: enforce formatting check in CI

kevmoo:


.github/workflows/dart_ci.yml
```diff
@@ -30,9 +30,9 @@ jobs:
       - name: Bootstrap Workspace
         run: melos bootstrap
 
-      # - name: Check Formatting
-      #   if: matrix.sdk == '3.10'
-      #   run: dart format --output=none --set-exit-if-changed .
+      - name: Check Formatting
+        if: matrix.sdk == 'dev'
```
pavelgj: I'm on stable..... I'n not sure developing on dev is generally a good practice -- most users are usually on older versions of the sdk or stable at best.

kevmoo: Let me hack on this to see if we can just use stable.

packages/genkit_firebase_ai/lib/genkit_firebase_ai.dart
```diff
@@ -526,7 +526,7 @@ Part fromGeminiPart(fai.Part p) {
 
   final metadata = <String, dynamic>{
     if (p.isThought == true) 'isThought': true,
-    if (thoughtSignature != null) 'thoughtSignature': thoughtSignature,
+    'thoughtSignature': ?thoughtSignature,
```
kevmoo, line 529: Now that the lints are setup here, we're getting the right errors!

pubspec.yaml
```diff
@@ -22,11 +24,13 @@ workspace:
   - testapps/chrome_example
   - testapps/openai_sample
   - testapps/mcp_server
+  - testapps/middleware
```
kevmoo, line 27: we missed one!

packages/genkit_middleware/pubspec.yaml
```diff
@@ -14,5 +14,4 @@ dependencies:
   schemantic: ^0.0.1-dev.16
 
 dev_dependencies:
-  lints: ^5.0.0
```
kevmoo, line 17: This was blocking the latest `dart_flutter_team_lints` from everywhere.

pavelgj (approved):

PR #167, opened 2026-02-26
refactor: hide package:json_schema_builder

### LINKED ISSUE #155: Consider deprecating the export of `package:json_schema_builder` from `package:schemantic`
I believe this package is serving 2 roles in the interface. The first is that it is used from generated code, the second is that it uses `Schema` as a return type and field type in a few places.

I recommend moving away from using this code through that library. We can prevent this by deprecating the `export` and adding automated fixes to use those APIs from their package import directly. This will let us still use these APIs in the generated code but prevent their direct use through this package.

We can change existing `jsb.Schema` references in public APIs to `Map<String, Object?>` which is equivalent (`Schema` is an `extension type`) Callers can still use the `Schema` utils on the _argument_ side, but we won't have the packages coupled at the API level.

The major benefit this brings is to avoid getting forced into major version bumps just because `package:json_schema_builder` had a major version bump.

natebosch:
Closes #155

The APIs used by generated code need to be available through the same
import as the annotation and related code because a `part of` file
cannot currently add imports. We do not want users to use the
`json_schema_builder` APIs through this import to avoid coupling these
packages. We do want to expose the validation functionality which is
currently only available through those APIs.

- Add a `typedef $Schema` as an alias for `Schema` to indicate a
  relationship with the generated code and discourage it's use. It is
  not currently marked with `@experimental` or `@deprecated` to more
  strongly discourage use, but the `$` is a convention with precedent in
  Dart and we are using it for other non user facing classes.
- Use `Map<String, Object?>` for APIs that previously used `jsb.Schema`.
- Add a `validate` method next to `parse` on the `SchemanticType` class.
- Add a `ValidationErrorType` enum and `ValidationError` extension type
  wrapping the types from `json_schema_builder` to hide them.
- Update usages of `jsonDecode(schema.toJson())` to use the Map value
  directly.


packages/schemantic/lib/schemantic.dart
```diff
@@ -444,3 +457,114 @@ class SchemaHelpers {
...
+
+  // Number/Integer specific
+  minimumNotMet,
+  maximumExceeded,
+  exclusiveMinimumNotMet,
+  exclusiveMaximumExceeded,
+  multipleOfInvalid,
+}
+
+extension on jsb.ValidationErrorType {
```
natebosch, line 531: @pavelgj - I ended up needing to duplicate/wrap a few error classes to get the validation stuff fully decoupled. I still think this is overall preferable to exporting `Schema` directly and using it in signatures in this package.

cc @kevmoo @lrhn for their opinion on this level of wrapping vs exporting directly.

kevmoo: Interesting take.

Not a small amount to maintain, but it doesn't seem like a huge deal, either.

pavelgj: yeah, can't say I'm loving this... but I guess it's fine.

packages/genkit_openai/lib/src/openai_plugin.dart
pavelgj: the `.flatten()` is coming from https://github.com/genkit-ai/genkit-dart/blob/main/packages/schemantic/lib/src/flatten.dart
need to figure what to do with it... can export just as a utility function, e.g. `flattenSchema(Map<String, Object?>)`

maybe through as separate entry point? `import 'package:schemantic/utils.dart';`

natebosch: Oh I didn't notice that was a local implementation.

We can implement that on `Map<String, Object?>`. Given the more broad type it might make sense to rename to `flattenSchema()` or maybe `get flatSchema`, but I left naming as is for now.

packages/schemantic/lib/schemantic.dart
natebosch: If we wind up wanting to expose these again instead of wrap them, I'm pretty sure this part of it will be non-breaking.

Changing the return types back to `Schema` from `Map<String, Object?>` might be breaking, but I think this part of it is safe to run with and we can re-evaluate if any problems creep up.

PR #170, opened 2026-02-26
chore: migrate skills to genkit-ai/skills

ssbushi:
Migrated in https://github.com/genkit-ai/skills/pull/12


kevmoo:
I think we like the dart-specific skills

PR #180, opened 2026-02-27
refactor!: Remove `SchemaHelpers` utility class

natebosch:
Resolves a violation of
https://dart.dev/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members

Rewrite `buildSchema` and the related private utility methods as
extension methods in private extensions. The `buildSchema` function is
invoked publicly through `SchemanticType.jsonSchema()`, with identical
behavior if the caller passes `uesRefs: true`.

### DIFF
--- a/packages/schemantic/CHANGELOG.md
+++ b/packages/schemantic/CHANGELOG.md
@@ -8,7 +8,6 @@
  - **FIX**: enable and fix a couple of lints (#174).
  - **FIX**: be consistent with String quotes (#164).
  - **BREAKING** **FEAT**: move basic type functions to static creation method on SchemanticType (#154).
-- **BREAKING** **REFACTOR**: Remove `SchemaHelpers` utility class.
 
 ## 0.0.1-dev.17
 


packages/schemantic/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
  - **FIX**: enable and fix a couple of lints (#174).
  - **FIX**: be consistent with String quotes (#164).
  - **BREAKING** **FEAT**: move basic type functions to static creation method on SchemanticType (#154).
+- **BREAKING** **REFACTOR**: Remove `SchemaHelpers` utility class.
```
pavelgj: these are autogenerated and dev.18 release is already out -- https://pub.dev/packages/schemantic/versions/0.0.1-dev.18

the changelog is autogenerated from semantic commit messages upon release

```
refactor!: Remove `SchemaHelpers` utility class.
```

☝️ will result in this message generated upon release

packages/schemantic/lib/schemantic.dart
```diff
@@ -348,17 +347,14 @@ abstract class SchemanticType<T> {
 }
 
 /// Internal utilities for building JSON Schemas.
-abstract final class SchemaHelpers {
-  /// Builds a complete schema for the [root] type, including all `$defs`.
-  static Map<String, Object?> buildSchema(SchemanticType root) {
-    if (root.schemaMetadata == null) {
-      return root.jsonSchema(useRefs: false);
-    }
+extension on SchemanticType {
```
pavelgj, line 350: I'm a little fuzzy on the visibility on extension types... won't these make these methods public?

natebosch: No, unnamed extensions can only be used within the library they are defined in. If we wanted this to be usable from outside the package we'd need something like `extension BuildSchema on SchemanticType`.

PR #188, opened 2026-03-02
feat(openai): support whisper model

CorieW:
Resolves https://github.com/genkit-ai/genkit-dart/issues/169

Adds support for OpenAI's whisper models.

**Testing:**
- [x] Loaded from audio (mp3) file
image

- [x] Loaded from video (mp4) file
image

- [x] base64 data url
image

### DIFF
--- a/testapps/openai_sample/lib/whisper_transcription.dart
+++ b/testapps/openai_sample/lib/whisper_transcription.dart
@@ -135,7 +135,8 @@ Flow<String, String, void, void> defineWhisperVideoTranscriptionFlow(
             role: Role.user,
             content: [
               TextPart(
-                text: 'Transcribe the audio in this MP4 file. Return only the transcript text.',
+                text:
+                    'Transcribe the audio in this MP4 file. Return only the transcript text.',
               ),
               MediaPart(
                 media: Media(url: dataUrl, contentType: mimeType),


packages/genkit_openai/lib/src/converters.dart
```diff
@@ -122,6 +137,71 @@ abstract final class GenkitConverter {
     };
   }
 
+  /// Returns true when media is audio input.
+  static bool _isAudioMedia(Media media) {
+    final contentType = media.contentType?.toLowerCase();
+    if (contentType != null && contentType.startsWith('audio/')) {
```
cabljac, line 143: `content: audio/wav` for a non-data URL

e.g `https://...`

we wouldn't actually check the url in this case right?

I think this would be caught elsewhere but is the early return here correct?

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -359,6 +378,241 @@ class OpenAIPlugin extends GenkitPlugin {
...
+        bodyBytes,
+        boundary,
+        fieldName: 'file',
+        filename: audioPayload.filename,
+        contentType: audioPayload.contentType,
+        bytes: audioPayload.bytes,
+      );
+
+      bodyBytes.addAll(utf8.encode('--$boundary--\r\n'));
+      request.add(bodyBytes);
```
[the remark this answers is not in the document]
cabljac: http is not currently a dependency right? Adopting it would add a new dependency just for this one use case - worth weighing up.

Is there any way with openai_dart to create a raw request like this?

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -359,6 +378,241 @@ class OpenAIPlugin extends GenkitPlugin {
...
+    final media = _extractAudioMedia(req.messages);
+    if (media == null) {
+      throw GenkitException(
+        'Transcription requires at least one audio MediaPart in request messages.',
+      );
+    }
+
+    final audioPayload = _decodeAudioDataUrl(media);
+    final transcriptionUri = _buildOpenAIUri('/audio/transcriptions');
+    final httpClient = HttpClient();
```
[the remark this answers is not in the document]
cabljac: have to be careful - right now we're doing `httpClient.close(force: true)` and we'd have to remove that, it'd destroy the shared client.

There's no cleanup hook in GenkitPlugin either right - 

if we did use the http package they would expose a `Client` class i presume

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -359,6 +378,241 @@ class OpenAIPlugin extends GenkitPlugin {
     );
   }
 
+  Future<ModelResponse> _handleTranscription(
+    ModelRequest req,
+    String modelName,
+    ({
```
cabljac, line 384: is there a typedef for ctx in genkit core

`ActionFnArg<ModelResponseChunk, ModelRequest, void> ctx,`

in action.dart:

```
  typedef ActionFnArg<Chunk, Input, Init> = ({
    bool streamingRequested,
    StreamingCallback<Chunk> sendChunk,  // = void Function(Chunk)
    Stream<Input>? inputStream,
    Init? init,
  });
```

not introduced here, something we should look at fixing

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -359,6 +378,241 @@ class OpenAIPlugin extends GenkitPlugin {
...
+      throw GenkitException(
+        'Audio media must be provided as a base64 data URL for transcription models.',
+      );
+    }
+
+    final contentType = match.group(1)!.toLowerCase();
+    final encodedAudio = url.substring(match.end);
+    final bytes = base64Decode(encodedAudio);
+    final filename = switch (contentType) {
+      'audio/wav' || 'audio/x-wav' || 'audio/wave' => 'audio.wav',
```
cabljac, line 564: Can we check against the API that we support all the formats that Whisper supports?

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -359,6 +378,241 @@ class OpenAIPlugin extends GenkitPlugin {
     );
   }
 
+  Future<ModelResponse> _handleTranscription(
```
cabljac, line 381: Am i correct in thinking right now, only model and prompt are forwarded to the API?

Other transcription parameters (language, temperature, response_format, timestamp_granularities, chunking_strategy)

I think we parse temperature but don't pass it on?

https://developers.openai.com/api/reference/resources/audio/subresources/transcriptions/methods/create

Hmm need to consider where these extra options live

packages/genkit_openai/lib/src/openai_plugin.dart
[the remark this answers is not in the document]
CorieW: I don't think so

packages/genkit_openai/lib/src/openai_plugin.dart
CorieW: Supports flac, mp3, mp4, mpeg, mpga, m4a, ogg, wav, and webm ([source](https://developers.openai.com/api/reference/resources/audio/subresources/transcriptions/methods/create/)).

I'll make some changes to support these.

CorieW: Made some changes

packages/genkit_openai/lib/src/openai_plugin.dart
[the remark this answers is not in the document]
CorieW: Actually, might be able to use `OpenAIClient.makeRequest`. However, that method is protected so it shows a warning. What's the preference?

PR #205, opened 2026-03-06
fix!: removed vertex support to unblock using plugin with flutter

pavelgj:
Will move vertex support into a separate package or via `genkit_vertex_auth`.


pavelgj:
FYI @leehack
We will add vertex support another way... just need to keep the package flutter friendly.

PR #216, opened 2026-03-16
docs: minor improvements to tool calling example

pavelgj:
added temp converter tool to better demonstrate the tool calling loop

### DIFF
--- a/packages/genkit_google_genai/example/example.dart
+++ b/packages/genkit_google_genai/example/example.dart
@@ -97,7 +97,8 @@ void main(List<String> args) async {
   );
   ai.defineTool(
     name: 'temperatureConverter',
-    description: 'Converts temperatures between Celsius (C) and Fahrenheit (F).',
+    description:
+        'Converts temperatures between Celsius (C) and Fahrenheit (F).',
     inputSchema: TemperatureConverterInput.$schema,
     outputSchema: TemperatureConverterOutput.$schema,
     fn: (input, context) async {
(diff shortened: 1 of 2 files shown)


packages/genkit_google_genai/example/example.dart
```diff
@@ -90,21 +90,41 @@ void main(List<String> args) async {
...
-        return 'The weather in Boston is 72 and sunny.';
+        return 'The weather in Boston is 72F and sunny.';
       }
-      return 'The weather in ${input.location} is 75 and cloudy.';
+      return 'The weather in ${input.location} is 75F and cloudy.';
+    },
+  );
+  ai.defineTool(
+    name: 'temperatureConverter',
+    description: 'Converts temperature',
```
kevmoo: Should this be more explanatory? Like:
```suggestion
    description: 'Converts temperatures between Celsius (C) and Fahrenheit (F).',
```

packages/genkit_google_genai/example/example.dart
```diff
@@ -90,21 +90,41 @@ void main(List<String> args) async {
...
+    inputSchema: TemperatureConverterInput.$schema,
+    outputSchema: TemperatureConverterOutput.$schema,
+    fn: (input, context) async {
+      if (input.unit == 'C') {
```
kevmoo: I hate stringy APIs for things that should be enums. Perhaps this is the way...

pavelgj: Found a bug in schemantic while trying to switch to enums, PTAL https://github.com/genkit-ai/genkit-dart/pull/220

PR #217, opened 2026-03-16
feat: Cache plugin action lists using a new adapter to optimize discovery

pavelgj:


packages/genkit/lib/src/core/registry.dart
```diff
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import './action.dart';
-import './plugin.dart';
+import '../../plugin.dart';
```
kevmoo: usually you want to avoid importing the "root" libraries.

pavelgj: good point, fixed

PR #223, opened 2026-03-18
feat: added support for bring-your-own-type and schema with schemantic

pavelgj:


pubspec.yaml
```diff
@@ -30,6 +30,7 @@ workspace:
   - testapps/remote_models
   - testapps/gemini
   - testapps/evals
+  - testapps/schemantic
```
kevmoo: We should rename this package something else – so it's clear in the test logs and the build logs what is what!

pavelgj: in logs it shows up as `schemantic_sample`... but renamed the folder as well to `testapps/schemantic_sample` for consistency.

PR #230, opened 2026-03-20
fix: let list() throw on API errors to preserve retry, graceful init() for custom models

### LINKED ISSUE #229: Failed to start the MCP service
Failed to start the MCP service. I am using an OpenAI-compatible API, and the MCP service fails to start. Here is the code:

If I remove the plugins parameter from OpenAI, it starts normally.

```dart
import 'dart:io';

import 'package:genkit_mcp/genkit_mcp.dart';
import 'package:genkit_openai/genkit_openai.dart';
import 'package:uuid/uuid.dart';

void main() async {
  final ai = Genkit(
    plugins: [
      openAI(
        apiKey: "aws-xxx",
        baseUrl: "http://xxx/v1",
        headers: {
          "Authorization": "aws-xxx",
          "appCode": "appdev-flickr-prod",
          "requestId": const Uuid().v7(),
          "accept": "*/*",
        },
        models: [
          const CustomModelDefinition(
            name: "amazon-anthropic.claude-opus-4-5-20251101-v1:0",
          ),
        ],
      ),
    ],
  );

  ai.defineTool(
    name: 'add',
    description: 'Add two numbers together',
    inputSchema: .map(.string(), .dynamicSchema()),
    fn: (input, _) async {
      final a = num.parse(input['a'].toString());
      final b = num.parse(input['b'].toString());
      return (a + b).toString();
    },
  );

  final server = createMcpServer(ai, McpServerOptions(name: 'my-server'));
  final transport = await StreamableHttpServerTransport.bind(
    address: InternetAddress.loopbackIPv4,
    port: 3000,
  );
  await server.start(transport);
}

```

```
Failed to list actions from plugin openai: GenkitException: Error listing models from OpenAI: ClientException with SocketException: Failed host lookup: 'xxx' (OS Error: nodename nor servname provided, or not known, errno = 8), uri=http://xxx/v1/models (Status: INTERNAL, Code: 13)

    INNER EXCEPTION:
    ClientException with SocketException: Failed host lookup: 'xxx' (OS Error: nodename nor servname provided, or not known, errno = 8), uri=http://xxx/v1/models

    INNER STACK TRACE:
    package:http/src/io_client.dart 227:7                              IOClient.send
    package:openai_dart/src/client/interceptor_chain.dart 180:36       InterceptorChain._executeTransport.executeTransport.sendRequest
    dart:async                                                         Future.timeout.<fn>
    package:openai_dart/src/client/retry_wrapper.dart 85:26            RetryWrapper.executeWithRetry
    package:openai_dart/src/interceptors/error_interceptor.dart 36:22  ErrorInterceptor.intercept
    package:openai_dart/src/resources/models_resource.dart 62:22       ModelsResource.list
    package:genkit_openai/src/openai_plugin.dart 199:24                OpenAIPlugin._fetchAvailableModels
    package:genkit_openai/src/openai_plugin.dart 247:24                OpenAIPlugin.list
    package:genkit/src/core/registry.dart 168:14                       _ListActionsCachingPluginAdapter.list
    package:genkit/src/core/registry.dart 125:31                       Registry.listActions
    package:genkit_mcp/src/server/mcp_server.dart 67:21                GenkitMcpServer.setup
    package:genkit_mcp/src/server/mcp_server.dart 88:5                 GenkitMcpServer.start
    package:agents/src/mcp/flicker_mcp.dart 45:3                       main #0      OpenAIPlugin.list (package:genkit_openai/src/openai_plugin.dart:269:7)
<asynchronous suspension>
#1      _ListActionsCachingPluginAdapter.list (package:genkit/src/core/registry.dart:168:14)
<asynchronous suspension>
#2      Registry.listActions (package:genkit/src/core/registry.dart:125:31)
<asynchronous suspension>
#3      GenkitMcpServer.setup (package:genkit_mcp/src/server/mcp_server.dart:67:21)
<asynchronous suspension>
#4      GenkitMcpServer.start (package:genkit_mcp/src/server/mcp_server.dart:88:5)
<asynchronous suspension>
#5      main (package:agents/src/mcp/flicker_mcp.dart:45:3)
<asynchronous suspension>
```

davidmigloz:
> [!IMPORTANT]
> **Depends on #225** (`pj/openai-httpClient`) — must be merged first.

## Summary
- `init()` now degrades gracefully on API errors (DNS failure, timeouts, invalid key) instead of throwing `GenkitException` — custom models are always registered
- `list()` continues to throw on API errors so the registry's `_ListActionsCachingPluginAdapter` can clear the cache and retry on subsequent calls
- Registry's `listActions()` uses `_logger.warning()` instead of `print()` to avoid scary console output
- Same `list()` fix applied to the Anthropic plugin

Closes #229

## Details

When a user configures the OpenAI plugin with a custom `baseUrl`, `list()` throws on API errors. The `Registry.listActions()` catch block already handles this non-fatally, but it used `print()` with the full stack trace, producing alarming console output: _"Failed to list actions from plugin openai: Error listing models from OpenAI: ClientException with SocketException"_.

**Three things fixed:**

1. **`init()` throws before registering custom models** — When `baseUrl == null` and the API is unreachable, `init()` previously threw at the API call before reaching the custom model loop. Now it catches the error and continues, so custom models are always registered. `resolve()` handles on-demand creation for API models.

2. **Registry uses `print()` for plugin errors** — Replaced with `_logger.warning()` via `package:logging` (already a dependency) so the error goes through proper log levels instead of raw stdout. This is the actual fix for the "scary error" reported in #229.

3. **Anthropic plugin's `list()` returns `[]` on error** — Same bug: returning `[]` causes the `_ListActionsCachingPluginAdapter` to cache the empty result permanently, preventing retry. Changed to throw `GenkitException` (consistent with the OpenAI plugin).

**Design: `list()` must throw, not return `[]`** — The `_ListActionsCachingPluginAdapter` caches successful results and clears the cache on errors to allow retry. Returning `[]` makes the adapter think the call succeeded, permanently caching an empty result even if the API recovers later.

| Scenario | Before | After |
|----------|--------|-------|
| Custom baseUrl, endpoint unreachable | `list()` throws, `print()` dumps scary error | `list()` throws, `_logger.warning()` (proper log level) |
| Default host, API unreachable | `init()` throws, custom models lost | `init()` catches error, registers custom models |
| Invalid API key (401) | Both throw with scary print | `init()` degrades, `list()` throws with proper logging |
| API recovers after initial failure | `list()` retried via cache adapter | **Same behavior preserved** |

## Test Plan
- [x] `dart analyze` — zero warnings across genkit, genkit_openai, genkit_anthropic
- [x] `dart test packages/genkit/test/core/registry_test.dart` — all 16 tests pass (including `list actions retry on failure`)
- [x] `dart test packages/genkit_openai/test/openai_plugin_test.dart` — all 37 tests pass
- [x] `list()` error tests updated to verify throws (not empty list)
- [x] `init()` error tests verify custom models survive API failures

### DIFF
--- a/packages/genkit_openai/lib/src/openai_plugin.dart
+++ b/packages/genkit_openai/lib/src/openai_plugin.dart
@@ -14,10 +14,13 @@
 
 import 'package:genkit/plugin.dart';
 import 'package:http/http.dart' as http;
+import 'package:logging/logging.dart';
 import 'package:openai_dart/openai_dart.dart' as sdk;
 import 'package:schemantic/schemantic.dart';
 
 import '../genkit_openai.dart';
+
+final _logger = Logger('genkit_openai');
 
 /// Returns true when the output config indicates JSON-structured output
 /// (format is 'json' or contentType is 'application/json').
@@ -76,10 +79,12 @@ class OpenAIPlugin extends GenkitPlugin {
           final info = _getModelInfo(modelId);
           actions.add(_createModel(modelId, info));
         }
-      } catch (e) {
-        throw GenkitException(
-          'Error fetching available models from OpenAI: $e',
-          underlyingException: e,
+      } catch (e, s) {
+        _logger.warning(
+          'Failed to fetch available models from OpenAI. '
+          'Only custom models will be available.',
+          e,
+          s,
         );
       }
     }
@@ -271,12 +276,9 @@ class OpenAIPlugin extends GenkitPlugin {
       }
 
       return modelMetadataList;
-    } catch (e, stackTrace) {
-      throw GenkitException(
-        'Error listing models from OpenAI: $e',
-        underlyingException: e,
-        stackTrace: stackTrace,
-      );
+    } catch (e, s) {
+      _logger.warning('Failed to list models from OpenAI', e, s);
+      return [];
     }
   }
 
--- a/packages/genkit_openai/lib/src/openai_plugin.dart
+++ b/packages/genkit_openai/lib/src/openai_plugin.dart
@@ -276,9 +276,12 @@ class OpenAIPlugin extends GenkitPlugin {
       }
 
       return modelMetadataList;
-    } catch (e, s) {
-      _logger.warning('Failed to list models from OpenAI', e, s);
-      return [];
+    } catch (e, stackTrace) {
+      throw GenkitException(
+        'Error listing models from OpenAI: $e',
+        underlyingException: e,
+        stackTrace: stackTrace,
+      );
     }
   }
 
--- a/packages/genkit/lib/src/core/registry.dart
+++ b/packages/genkit/lib/src/core/registry.dart
@@ -12,9 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'package:logging/logging.dart';
+
 import '../ai/generate_middleware.dart';
 import './action.dart';
 import './plugin.dart';
+
+final _logger = Logger('genkit.registry');
 
 class Registry {
   final Map<String, Action> _actions = {};
@@ -130,7 +134,11 @@ class Registry {
           }
         }
       } catch (e, st) {
-        print('Failed to list actions from plugin ${plugin.name}: $e $st');
+        _logger.warning(
+          'Failed to list actions from plugin ${plugin.name}',
+          e,
+          st,
+        );
       }
     }
     return allActions.values.toList();
--- a/packages/genkit_anthropic/lib/src/plugin_impl.dart
+++ b/packages/genkit_anthropic/lib/src/plugin_impl.dart
@@ -72,9 +72,11 @@ class AnthropicPluginImpl extends GenkitPlugin {
           )
           .toList();
     } catch (e, s) {
-      // Fallback or empty if listing fails/not supported as expected
-      print('Failed to list Anthropic models: $e\n$s');
-      return [];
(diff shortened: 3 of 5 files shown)


packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -271,12 +276,9 @@ class OpenAIPlugin extends GenkitPlugin {
...
 
       return modelMetadataList;
-    } catch (e, stackTrace) {
-      throw GenkitException(
-        'Error listing models from OpenAI: $e',
-        underlyingException: e,
-        stackTrace: stackTrace,
-      );
+    } catch (e, s) {
+      _logger.warning('Failed to list models from OpenAI', e, s);
```
pavelgj: The framework will catch errors thrown from list and they are non-fatal and it will retry next time list is called. Also, the genkit registry will cache list responses from plugins... so the difference between throwing an error and just returning an empty list is whether genkit will retry. This change would effectively make it so that we ever only try to list once and never retry. I'm honestly not sure what the correct behaviour should be...

davidmigloz: Indeed, this makes the empty list to be permanently cached killing any retries. I've reverted it to throw, but replaced the scary print the user was reporting with `_logger.warning(...)`. Also fixed the same return `[]` bug in the Anthropic plugin's `list()`.

PR #242, opened 2026-03-27
chore: add a gemini cli hook to run format on file changes

MichaelDoyle:


tools/hooks/format.dart
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+  } catch (e) {
+    // Silent failure on parse errors, will fallback to '.'
+  }
+
+  final formatPath = (filePath != null && filePath.isNotEmpty) ? filePath : '.';
+
+  // Run dart format
+  final result = Process.runSync('dart', ['format', formatPath]);
```
kevmoo: You can use [Process.start](https://api.dart.dev/dart-io/Process/start.html) and have it pipe stdout and stderr usig https://api.dart.dev/dart-io/ProcessStartMode/inheritStdio-constant.html

MUCH cleaner!

kevmoo: @gemini-code-assist - what do you think?

MichaelDoyle: Thanks @kevmoo!

tools/hooks/format.dart
```diff
@@ -0,0 +1,57 @@
...
+      final input = jsonDecode(inputString);
+      if (input case {
+        'tool_input': {'args': {'file_path': final String path}},
+      }) {
+        filePath = path;
+      }
+    }
+  } catch (e) {
+    // Silent failure on parse errors, will fallback to '.'
+  }
```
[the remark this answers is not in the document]
MichaelDoyle: I understand the feedback. We are being deliberate here, because it is a CLI hook.

kevmoo (commented):
I like it, but we should see what @pavelgj says

Also see

https://github.com/kevmoo/personal_dotfiles/commit/0fa84ca422d1238d2f68eeb0b54c64a20fbab433#diff-1f6eec92873f0e01ec265c80f26c4d5cc8be4a1a833f08fa848c815d95b61fbdR18-R27

MUSH be paranoid about partially staged files!

MichaelDoyle:
WDYT @pavelgj 

BTW, on a totally separate, but related not, I see we have `.agent/rules` but this doesn't seem to be gemini/jetski friendly. which tool wants `.agent` and does it support some of the other locations?

PR #261, opened 2026-04-28
feat(genkit_vertexai): support Gemini and multimodal embedders

CorieW:
Handle Vertex AI embedding models with the correct request shape for Gemini, legacy text embedding, and multimodal embedding APIs. Add fallback behavior for older Gemini models and expand tests around embedder listing, request routing, and mock HTTP responses.

## Experienced Problem
Currently, multimodal embedder has a limitation where it can only have one document at one time, due to the following limitation:

> Looks like there is a limitation in the current embedder response shape for multimodal batch embedding in Genkit Dart (maybe others as well?). `embedMany` returns `List<Embedding>`, which works well when each input document produces exactly one embedding, but becomes ambiguous when a single document can produce multiple modality-specific embeddings.
> 
> For example:
> - Document 1: image + text
> - Document 2: text
> - Document 3: text
> 
> A flat result may contain four embeddings, but it does not indicate which embeddings belong to which input document or modality. A clearer response shape for multimodal batch embedding would be `List<List<Embedding>>`, where the outer list maps to input documents and the inner list contains that document's modality-specific embeddings.

### Solutions

1. Allow any number of single modality documents OR a single document with any number of modalities.
2. Add a multimodal-specific embedding to Genkit to avoid breaking changes with non-multimodal embedders.
3. Use metadata property to associate embeddings to documents.

### Implemented Solution

Used metadata property to associate embeddings to documents.

image

## Before these changes, the following things were happening
- Model Parity
  - `text-embedding-005` - “No input variables specified for this action”.
  - `text-multilingual-embedding-002` - “No input variables specified for this action”.
  - `multimodalembedding@001` - Doesn't work.
  - `gemini-embedding-001 ` - “No input variables specified for this action”.

## Testing
- [x] Embedders registered
image

- [x] Gemini embedder
image

- [x] Text embedder
image

- [x] Multimodal embedder
image

## Update

Pushed a few follow-up commits to the branch (`dbd2ebb`, `1805697`, `efe3a09`).

**Fixed the multi-input Gemini embedding path.** Testing live against Vertex, I found the `gemini-embedding` path called `:batchEmbedContents`, which Vertex does not have (the REST reference lists only `embedContent` and `predict` on `publishers.models`, and the endpoint 404s). The mock tests asserted the request URL rather than a real response, so it passed CI but failed against the API. I routed all Gemini embedding models through `:predict` instead (the same shape the text embedders already use), verified it live for single and multiple inputs, and dropped the now-dead `embedContent`/`batchEmbedContents` plumbing.

**Documented the multimodal output contract.** Added a doc comment on the multimodal path spelling out the 1:N shape (one embedding per modality per document, so the result is not positionally aligned with the input docs) and the metadata keys callers group by.

**Added testapp flows.** `multimodalEmbedding` (structured `{caption, imageUri}` input, returns each embedding's metadata + dimensions so the per-modality mapping is visible) and `embeddingWithOptions`. Both smoke-tested end to end against a real project through the dev UI.

**Out of scope here: `gemini-embedding-2`.** It is served on a different surface (`aiplatform.{us,eu}.rep.googleapis.com`, `embedContent`, a multi-region location) that this plugin does not build yet, so it is not wired up in this PR. It was not working before these changes either. I will follow up with a separate PR for the rep endpoint. One thing worth a look in review: `list()` surfaces it as an embedder via the `contains('embedding')` filter even though calling it would fail today, so we may want to filter it out for now.

Verification: 19 mock + 10 live tests green, analyzer clean, format and license passing.


packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,480 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:convert';
+
+import 'package:genkit/plugin.dart';
+import 'package:genkit_google_genai/common.dart' as google;
+// ignore: implementation_imports
+import 'package:genkit_google_genai/src/generated/generativelanguage.dart'
```
cabljac: is this correct to import from the generated like this

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,480 @@
...
+  final mediaParts = doc.content
+      .where((part) => part.isMedia)
+      .map((part) => part.media!)
+      .toList();
+
+  // A document can only contain one input type here.
+  // Text, image, and video use different embedding fields in the Vertex
+  // response, and this code needs one clear field to read for each document.
+  if (text.isNotEmpty && mediaParts.isNotEmpty) {
+    throw GenkitException(
```
cabljac: Do we have test coverage for all of these error cases?

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,480 @@
...
+_VertexEmbedderRequestShape _requestShapeFor(String modelName) {
+  final baseModelName = _baseModelName(modelName);
+  // Check the broad model families in order.
+  if (_isMultimodalEmbeddingFamily(baseModelName)) {
+    return _VertexEmbedderRequestShape.multimodalPredict;
+  }
+  if (_usesLegacyGeminiPredictApi(baseModelName)) {
+    return _VertexEmbedderRequestShape.textPredict;
+  }
+  if (_isGeminiEmbeddingFamily(baseModelName)) {
```
cabljac: this one and the check above seem to overlap right? for `gemini-embedding-001` for example

could we do something table-based like

```dart
const _shapeByExactModel = {
    'gemini-embedding-001': _VertexEmbedderRequestShape.textPredict,
  };

_VertexEmbedderRequestShape _requestShapeFor(String name) {
  final base = _baseModelName(name);
  final exact = _shapeByExactModel[base];
  if (exact != null) return exact;
  ...
}
```

cabljac: just wary that we have two checks, swapping their order looks innocent (without checking what they do) but will break things

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,480 @@
...
+List<double> _multimodalPredictionEmbedding(
+  Map<String, dynamic> prediction, {
+  required _MultimodalOutput expectedOutput,
+}) {
+  // Read the embedding field that matches the input type.
+  final values = switch (expectedOutput) {
+    _MultimodalOutput.text => prediction['textEmbedding'] as List?,
+    _MultimodalOutput.image => prediction['imageEmbedding'] as List?,
+    _MultimodalOutput.video =>
+      ((prediction['videoEmbeddings'] as List?)?.firstOrNull
```
cabljac:
```
  {
    "predictions": [{
      "videoEmbeddings": [
        {"startOffsetSec": 0,  "endOffsetSec": 16, "embedding": [...]},
        {"startOffsetSec": 16, "endOffsetSec": 32, "embedding": [...]},
        {"startOffsetSec": 32, "endOffsetSec": 48, "embedding": [...]}
      ]
    }]
  }
```
can we check - vertex might do something like the above 

https://docs.cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-multimodal-embeddings#video-modes

we'd lose all but the first value right?

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,480 @@
...
+Future<List<Embedding>> _runMultimodalPredictRequests({
+  required google.GenerativeLanguageBaseClient service,
+  required String embedderName,
+  required List<DocumentData> docs,
+  required google.TextEmbedderOptions? options,
+}) async {
+  // Multimodal embedders use a different predict request and response shape.
+  final instances = docs.map(_toMultimodalInstance).toList();
+  final parameters = <String, dynamic>{};
+  if (options?.outputDimensionality != null) {
```
cabljac: nit: can we just add a comment here pointing out the divergence, like

```
// Multimodal predict expects `parameters.dimension` (not
  outputDimensionality)."
```

cabljac: just in case an AI model tries to "clean it up"

cabljac:
the multimodal media-input logic (`embedders.dart:323-434`) has 7 throw sites but only 2 covered by tests. Could add coverage for the remaining failure modes? 

  - Data-URI image input → asserts `bytesBase64Encoded` + `mimeType` in the
  request body.
  - `gs://` image input → asserts `gcsUri` + `mimeType` in the request body.
  - Mixed text + media in one document → throws (constraint called out in the PR
   description, currently unpinned).
  - Multiple media parts in one document → throws.
  - Unsupported MIME type (e.g. `audio/wav`) → throws.
  - Missing MIME (no `contentType`, non-`data:` URL) → throws.

packages/genkit_vertexai/lib/src/embedders.dart
cabljac: Basically importing from `genkit_google_genai/src/generated/...` via `// ignore: implementation_imports` couples this package to an internal path. If the generated file moves or is renamed, this likely will break silently, as far as i understand

Could instead re-export the types we need from common.dart or somewhere else?

packages/genkit_vertexai/lib/src/embedders.dart
CorieW: The legacy model uses a different shape to the newer gemini embedding models. `_usesLegacyGeminiPredictApi()` happens before `_isGeminiEmbeddingFamily()` to avoid the old gemini model (`gemini-embedding-001`) from getting caught under `_isGeminiEmbeddingFamily` and given the newer shape, which it doesn't support.

Yeah, your suggestion is good.

packages/genkit_vertexai/lib/src/embedders.dart
CorieW: Fixed I think

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+        final modelMap = m as Map<String, dynamic>;
+        final name = modelMap['name'] as String?;
+        return name != null && name.contains('embedding');
+      })
+      .map((m) {
+        final modelMap = m as Map<String, dynamic>;
+        final modelName = (modelMap['name'] as String).split('/').last;
+        return _vertexEmbedderMetadata('$pluginName/$modelName');
+      })
+      .toList();
```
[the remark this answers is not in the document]
cabljac: fixed in 67c4c37

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+
+  return rawPredictions.map((prediction) {
+    if (prediction is! Map<String, dynamic>) {
+      throw GenkitException(
+        'Vertex AI returned an invalid prediction payload.',
+        status: StatusCodes.INTERNAL,
+      );
+    }
+    return prediction;
+  }).toList();
```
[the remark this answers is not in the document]
cabljac: fixed in 67c4c37

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+  final instances = [
+    for (var i = 0; i < docs.length; i++)
+      _toMultimodalInstance(docs[i], documentIndex: i),
+  ];
+  final parameters = <String, dynamic>{};
+  if (options?.outputDimensionality != null) {
+    // Multimodal predict expects `parameters.dimension`, not
+    // `outputDimensionality`.
+    parameters['dimension'] = options!.outputDimensionality;
+  }
```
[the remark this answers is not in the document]
cabljac: fixed in 67c4c37

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+}) async {
+  // Older text embedders still use the predict payload shape.
+  final instances = docs.map((doc) {
+    final instance = <String, dynamic>{'content': _documentText(doc)};
+    if (options?.title != null) {
+      instance['title'] = options!.title;
+    }
+    if (options?.taskType != null) {
+      instance['task_type'] = options!.taskType;
+    }
```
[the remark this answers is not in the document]
cabljac: fixed in 67c4c37

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+    return instance;
+  }).toList();
+
+  final parameters = <String, dynamic>{};
+  if (options?.outputDimensionality != null) {
+    parameters['outputDimensionality'] = options!.outputDimensionality;
+  }
```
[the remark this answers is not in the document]
cabljac: fixed in 67c4c37

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+    expectedCount: docs.length,
+  );
+  return predictions.map(_textPredictionEmbedding).toList();
+}
+
+Embedding _textPredictionEmbedding(Map<String, dynamic> prediction) {
+  final embeddingData = prediction['embeddings'];
+  final values = embeddingData is Map<String, dynamic>
+      ? embeddingData['values']
+      : null;
```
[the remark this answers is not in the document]
cabljac: fixed in 67c4c37

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+        },
+      ),
+    );
+  }
+
+  for (var i = 0; i < doc.content.length; i++) {
+    final part = doc.content[i];
+    if (!part.isMedia) continue;
+
+    final mediaField = _toMultimodalMediaField(part.media!);
```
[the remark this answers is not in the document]
cabljac: fixed in 67c4c37

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+        final videoEmbeddings = prediction['videoEmbeddings'] as List?;
+        if (videoEmbeddings == null || videoEmbeddings.isEmpty) {
+          throw GenkitException(
+            'Vertex multimodalembedding did not return a video embedding.',
+            status: StatusCodes.INTERNAL,
+          );
+        }
+
+        for (var i = 0; i < videoEmbeddings.length; i++) {
+          final videoEmbedding = videoEmbeddings[i] as Map<String, dynamic>;
```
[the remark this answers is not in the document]
cabljac: fixed in 67c4c37

packages/genkit_vertexai/lib/src/embedders.dart
```diff
@@ -0,0 +1,474 @@
...
+          service.client.close();
+        }
+      }
+    },
+  );
+}
+
+String _documentText(DocumentData doc) {
+  return doc.content.where((p) => p.isText).map((p) => p.text).join('\n');
+}
```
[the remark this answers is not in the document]
cabljac: `isText` already guarantees a non-null `.text`, so this is redundant. Leaving as is.

packages/genkit_vertexai/lib/src/vertex_api_client.dart
pavelgj, line 1: perhaps worth updating the README to mention support for multimodal embeddings?

pavelgj (approved):

cabljac (approved):

PR #263, opened 2026-04-28
feat(vertexai): integrate Claude model support

CorieW:
Adds support for Claude models in VertexAI plugin

## Testing
image

### DIFF
--- a/packages/genkit_anthropic/lib/src/converters.dart
+++ b/packages/genkit_anthropic/lib/src/converters.dart
@@ -58,6 +58,13 @@ sdk.MessageCreateRequest toAnthropicCreateRequest(
       )
       .toList();
 
+  if (messages.isEmpty) {
+    throw GenkitException(
+      'Anthropic requests must include at least one non-system message.',
+      status: StatusCodes.INVALID_ARGUMENT,
+    );
+  }
+
   final tools =
       req.tools?.map(toAnthropicTool).toList() ?? <sdk.ToolDefinition>[];
 
--- a/packages/genkit_anthropic/lib/src/converters.dart
+++ b/packages/genkit_anthropic/lib/src/converters.dart
@@ -1,0 +1,349 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:convert';
+
+import 'package:anthropic_sdk_dart/anthropic_sdk_dart.dart' as sdk;
+import 'package:genkit/plugin.dart';
+import 'package:schemantic/schemantic.dart';
+
+import 'model.dart';
+
+const _defaultMaxTokens = 4096;
+const _defaultStructuredOutputToolName = 'return_output';
+const _minimumThinkingBudgetTokens = 1024;
+
+typedef AnthropicMediaConverter =
+    List<sdk.InputContentBlock> Function(String url, String? contentType);
+
+typedef AnthropicToolOutputFormatter = String Function(dynamic output);
+
+/// Converts a Genkit model request to an Anthropic message request.
+sdk.MessageCreateRequest toAnthropicCreateRequest(
+  ModelRequest req,
+  String modelName,
+  AnthropicOptions options, {
+  String structuredOutputToolName = _defaultStructuredOutputToolName,
+  AnthropicMediaConverter mediaConverter = convertAnthropicMedia,
+  AnthropicToolOutputFormatter toolOutputFormatter =
+      defaultAnthropicToolOutputFormatter,
+}) {
+  final systemMessage = req.messages
+      .where((message) => message.role == Role.system)
+      .firstOrNull;
+
+  final system = systemMessage != null
+      ? convertSystemMessage(systemMessage)
+      : null;
+
+  final messages = req.messages
+      .where((message) => message.role != Role.system)
+      .map(
+        (message) => toAnthropicMessage(
+          message,
+          mediaConverter: mediaConverter,
+          toolOutputFormatter: toolOutputFormatter,
+        ),
+      )
+      .toList();
+
+  final tools =
+      req.tools?.map(toAnthropicTool).toList() ?? <sdk.ToolDefinition>[];
+
+  sdk.ToolChoice? toolChoice;
+
+  if (req.output?.schema != null) {
+    final schema = Map<String, dynamic>.from(req.output!.schema!);
+    if (!schema.containsKey('type')) {
+      schema['type'] = 'object';
+    }
+    tools.add(
+      sdk.ToolDefinition.custom(
+        sdk.Tool(
+          name: structuredOutputToolName,
+          description: 'Return the structured output.',
+          inputSchema: sdk.InputSchema.fromJson(schema),
+        ),
+      ),
+    );
+    toolChoice = sdk.ToolChoice.tool(structuredOutputToolName);
+  }
+
+  if (req.toolChoice != null) {
+    toolChoice = switch (req.toolChoice) {
+      'auto' => sdk.ToolChoice.auto(),
+      'any' => sdk.ToolChoice.any(),
+      'none' => sdk.ToolChoice.none(),
+      final name => sdk.ToolChoice.tool(name!),
+    };
+  }
+
+  return sdk.MessageCreateRequest(
+    model: modelName,
+    messages: messages,
+    system: system,
+    maxTokens: options.maxTokens ?? _defaultMaxTokens,
+    temperature: options.temperature,
+    topP: options.topP,
+    topK: options.topK,
+    stopSequences: options.stopSequences,
+    tools: tools.isNotEmpty ? tools : null,
+    toolChoice: toolChoice,
+    thinking: mapThinkingConfig(options.thinking),
+  );
+}
+
+/// Converts a Genkit system [Message] to an Anthropic [sdk.SystemPrompt].
+sdk.SystemPrompt? convertSystemMessage(Message message) {
+  final parts = <String>[];
+  for (final part in message.content) {
+    if (part.isText) {
+      parts.add(part.text!);
+    }
+  }
+
+  final text = parts.join('\n');
+  if (text.isEmpty) {
+    return null;
+  }
+  return sdk.SystemPrompt.text(text);
+}
+
+/// Converts a Genkit [Message] to an Anthropic [sdk.InputMessage].
+sdk.InputMessage toAnthropicMessage(
+  Message message, {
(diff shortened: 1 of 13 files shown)


packages/genkit_vertexai/example/example.dart
```diff
@@ -41,6 +41,21 @@ void main(List<String> args) async {
...
   );
 
+  // --- Claude Generate Flow ---
+  ai.defineFlow(
+    name: 'claudeGenerate',
+    inputSchema: .string(defaultValue: 'Explain Model Garden briefly.'),
+    outputSchema: .string(),
+    fn: (input, context) async {
+      final response = await ai.generate(
+        model: vertexAI.claude('claude-sonnet-4-6'),
```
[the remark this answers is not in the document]
CorieW: wrong

packages/genkit_vertexai/README.md
```diff
@@ -33,10 +34,42 @@ void main() async {
...
+    plugins: [
+      vertexAI(
+        projectId: Platform.environment['GCLOUD_PROJECT'],
+        location: Platform.environment['GCLOUD_LOCATION'] ?? 'us-central1',
+      )
+    ],
+  );
+
+  final response = await ai.generate(
+    model: vertexAI.claude('claude-sonnet-4-6'),
```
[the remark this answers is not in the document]
CorieW: wrong

packages/genkit_vertexai/lib/src/claude.dart
```diff
@@ -0,0 +1,520 @@
...
+        ),
+      ),
+    ];
+  }
+
+  throw GenkitException(
+    'Vertex AI Claude models require media URLs to be data URLs with base64 '
+    'image data.',
+    status: StatusCodes.INVALID_ARGUMENT,
+  );
```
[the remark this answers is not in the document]
CorieW: Uncertain on validity

packages/genkit_vertexai/lib/src/vertex_api_client.dart
```diff
@@ -190,4 +243,14 @@ class VertexAiPluginImpl extends CommonGoogleGenPlugin {
...
   }
+
+  String _resolveBaseUrl(String safeLocation) {
+    if (safeLocation == 'global') {
+      return 'https://aiplatform.googleapis.com/';
+    }
+    if (_vertexMultiRegions.contains(safeLocation)) {
+      return 'https://aiplatform.$safeLocation.rep.googleapis.com/';
+    }
+    return 'https://$safeLocation-aiplatform.googleapis.com/';
```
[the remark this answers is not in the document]
CorieW: I checked the `.rep.googleapis.com` endpoint concern against the Vertex AI partner-model docs. For normal regional endpoints, `{location}-aiplatform.googleapis.com` is correct, but Claude partner-model multi-region endpoints are documented separately:

- US multi-region: `aiplatform.us.rep.googleapis.com`
- EU multi-region: `aiplatform.eu.rep.googleapis.com`

Docs: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-partner-models#multi-region
Google Cloud announcement also uses the same host pattern for Claude multi-region requests:
https://cloud.google.com/blog/products/ai-machine-learning/multi-region-endpoints-for-claude-available-on-vertex-ai

So I’m going to keep the `us` / `eu` `.rep.googleapis.com` handling here.

packages/genkit_vertexai/lib/src/claude.dart
```diff
@@ -0,0 +1,520 @@
...
+        .firstOrNull;
+
+    final system = systemMessage != null
+        ? _convertSystemMessage(systemMessage)
+        : null;
+
+    final messages = req.messages
+        .where((m) => m.role != Role.system)
+        .map(_toMessage)
+        .toList();
```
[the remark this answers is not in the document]
CorieW: Works fine

image

packages/genkit_vertexai/lib/src/claude.dart
```diff
@@ -0,0 +1,520 @@
+// Copyright 2025 Google LLC
+//
```
cabljac, line 2: a lot of this code is identical to the code in
https://github.com/genkit-ai/genkit-dart/blob/main/packages/genkit_anthropic/lib/src/plugin_impl.dart

perhaps it could be shared somehow?

packages/genkit_vertexai/lib/src/claude.dart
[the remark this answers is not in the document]
CorieW: I don’t think this suggestion is valid for Vertex AI Claude as currently documented.

Vertex AI Claude uses the Anthropic Messages API request shape, not Gemini’s `fileData`/GCS media shape. Google’s Vertex Claude docs point users to the Claude Messages API and show requests sent to `publishers/anthropic/models/...:rawPredict` with `anthropic_version` and `messages`: [Vertex AI Claude docs](https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/claude/use-claude).

For Claude image inputs, Anthropic documents supported sources as base64 image blocks, URL references, or the Anthropic Files API. I don’t see a documented `gcs_uri` image source type there: [Anthropic vision docs](https://docs.anthropic.com/en/docs/build-with-claude/vision).

The Dart SDK this implementation uses also only exposes `ImageSource.base64` and `ImageSource.url`; no `gcs_uri` source exists: [`anthropic_sdk_dart` ImageSource](https://pub.dev/documentation/anthropic_sdk_dart/latest/anthropic_sdk_dart/ImageSource-class.html).

packages/genkit_vertexai/lib/src/claude.dart
cabljac: Addressed in 1b86965 — shared helpers extracted into `packages/genkit_anthropic/lib/src/converters.dart` and re-exported via `common.dart`. Vertex-specific behaviour preserved via parameters (custom media converter, structured-output tool name, tool output formatter). Thanks!

packages/genkit_anthropic/lib/src/converters.dart
```diff
@@ -0,0 +1,349 @@
...
+}
+
+/// Converts Genkit media to Anthropic media blocks for the Anthropic API.
+List<sdk.InputContentBlock> convertAnthropicMedia(
+  String url,
+  String? contentType,
+) {
+  if (url.startsWith('data:')) {
+    final commaIndex = url.indexOf(',');
+    final base64Data = url.substring(commaIndex + 1);
```
cabljac: Can you verify what happens here for a malformed data URL with no comma?

Looks like if `indexOf(',')` returns `-1`, then `substring(commaIndex + 1)` becomes `substring(0)` and we'd send the entire URL string as base64 image data to Anthropic. A quick test against `convertAnthropicMedia` seemed to confirm:

```dart
final result = convertAnthropicMedia('data:image/png;base64nocomma', null);
final src = (result.first as ImageInputBlock).source as Base64ImageSource;
// src.data == 'data:image/png;base64nocomma'
```

Probably worth validating the comma exists and throwing `GenkitException(... INVALID_ARGUMENT)` if not — unless something upstream already rejects this before it gets here?

Note: same logic exists in `packages/genkit_vertexai/lib/src/claude.dart:234-235`.

packages/genkit_anthropic/lib/src/converters.dart
```diff
@@ -0,0 +1,349 @@
...
+          name: request.name,
+          input: request.input ?? {},
+        ),
+      ];
+    }
+    if (part.isToolResponse) {
+      final response = part.toolResponse!;
+      return [
+        sdk.InputContentBlock.toolResult(
+          toolUseId: response.ref ?? '',
```
cabljac: Is `ref` ever expected to be `null` in practice for tool requests/responses?

Both this site and `:140` do `response.ref ?? ''` / `request.ref ?? ''`. If a null ever slips through (e.g. a model output that didn't populate `ref`, or a tool response constructed without one), we'd emit empty `id` / `toolUseId` strings. Anthropic accepts arbitrary IDs so this won't error outright, but parallel tool calls would all share `""` and `tool_result` couldn't be correlated back to `tool_use`.

Confirmed the coercion happens:

```dart
final block = toAnthropicMessage(msg).blocks.first as ToolUseInputBlock;
// block.id == ''
```

If `ref` is guaranteed non-null by the time it reaches here, the `?? ''` is dead code and can go. If it isn't guaranteed, should we throw instead?

packages/genkit_anthropic/lib/src/converters.dart
```diff
@@ -0,0 +1,349 @@
...
+  final messages = req.messages
+      .where((message) => message.role != Role.system)
+      .map(
+        (message) => toAnthropicMessage(
+          message,
+          mediaConverter: mediaConverter,
+          toolOutputFormatter: toolOutputFormatter,
+        ),
+      )
+      .toList();
```
cabljac, line 59: What's the expected behaviour when the caller passes only system messages?

The `where(... != Role.system)` filter produces an empty list and we still build a `MessageCreateRequest` with `messages: []`. Anthropic's API requires at least one non-system message, so this would 400 at runtime with a fairly opaque error.

Confirmed:

```dart
final req = ModelRequest(messages: [
  Message(role: Role.system, content: [TextPart(text: '...')]),
]);
final out = toAnthropicCreateRequest(req, 'claude-sonnet-4-6', AnthropicOptions());
// out.messages.length == 0
```

Low likelihood, but might be worth an early `GenkitException` with a clearer message. Is there an upstream guard for this I missed?

kevmoo:
Still a draft @cabljac ?

packages/genkit_anthropic/lib/src/converters.dart
CorieW: Addressed. `ToolRequest.ref` and `ToolResponse.ref` are nullable in Genkit’s generated types, so the `?? ''` fallback was not safe. The Anthropic converter now rejects null or empty refs with `GenkitException(... INVALID_ARGUMENT)` instead of emitting empty `id` / `toolUseId` values.

packages/genkit_anthropic/lib/src/converters.dart
CorieW: Addressed. I verified the existing behavior and malformed data URLs like `data:image/png;base64nocomma` were being accepted, with the whole URL string treated as the base64 payload.

Added shared validation for `data:` URLs so they must contain the comma separator before extracting the base64 payload. Missing comma now throws `GenkitException(... INVALID_ARGUMENT)`.

packages/genkit_anthropic/lib/src/converters.dart
CorieW: Addressed. I didn't find an upstream guard for this, so `toAnthropicCreateRequest` now validates the converted message list before building the Anthropic request.

If the request contains only system messages, the converter now throws `GenkitException(... INVALID_ARGUMENT)` with a clearer message instead of sending `messages: []` and relying on Anthropic to reject it.

cabljac:
Sorry! ready for review now @kevmoo

packages/genkit_anthropic/lib/src/converters.dart
```diff
@@ -0,0 +1,378 @@
...
+  }
+  return ref;
+}
+
+/// Converts Genkit media to Anthropic media blocks for the Anthropic API.
+List<sdk.InputContentBlock> convertAnthropicMedia(
+  String url,
+  String? contentType,
+) {
+  if (url.startsWith('data:')) {
```
kevmoo, line 192: We have support for data URLs! https://api.dart.dev/dart-core/Uri/data.html

kevmoo (commented):
Please rebase! we have conflicts

PR #266, opened 2026-04-28
feat(vertexai): support meta's llama models

CorieW:
Support Meta's Llama models in VertexAI.

## Testing

image

image

### DIFF
--- a/packages/genkit_vertexai/lib/src/meta_model.dart
+++ b/packages/genkit_vertexai/lib/src/meta_model.dart
@@ -1,0 +1,351 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:convert';
+
+import 'package:genkit/plugin.dart';
+import 'package:schemantic/schemantic.dart';
+
+final metaModelInfo = ModelInfo(
+  supports: {
+    'multiturn': true,
+    'tools': true,
+    'systemRole': true,
+    'constrained': true,
+  },
+);
+
+base class VertexAiMetaOptions {
+  factory VertexAiMetaOptions.fromJson(Map<String, dynamic> json) =>
+      $schema.parse(json);
+
+  VertexAiMetaOptions._(this._json);
+
+  VertexAiMetaOptions({
+    String? version,
+    double? temperature,
+    double? topP,
+    int? maxTokens,
+    List<String>? stop,
+    double? presencePenalty,
+    double? frequencyPenalty,
+    int? seed,
+    String? user,
+    bool? llamaGuard,
+  }) {
+    _json = {
+      'version': ?version,
+      'temperature': ?temperature,
+      'topP': ?topP,
+      'maxTokens': ?maxTokens,
+      'stop': ?stop,
+      'presencePenalty': ?presencePenalty,
+      'frequencyPenalty': ?frequencyPenalty,
+      'seed': ?seed,
+      'user': ?user,
+      'llamaGuard': ?llamaGuard,
+    };
+  }
+
+  late final Map<String, dynamic> _json;
+
+  static const SchemanticType<VertexAiMetaOptions> $schema =
+      _VertexAiMetaOptionsTypeFactory();
+
+  String? get version => _json['version'] as String?;
+  double? get temperature => (_json['temperature'] as num?)?.toDouble();
+  double? get topP => (_json['topP'] as num?)?.toDouble();
+  int? get maxTokens => _json['maxTokens'] as int?;
+  List<String>? get stop => (_json['stop'] as List?)?.cast<String>();
+  double? get presencePenalty => (_json['presencePenalty'] as num?)?.toDouble();
+  double? get frequencyPenalty =>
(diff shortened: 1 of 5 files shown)


packages/genkit_vertexai/lib/src/meta_model.dart
```diff
@@ -0,0 +1,364 @@
...
+      status: StatusCodes.INVALID_ARGUMENT,
+    ),
+  };
+}
+
+List<Map<String, dynamic>> _toMetaToolMessages(Message message) {
+  final toolResponses = message.content
+      .where((p) => p.isToolResponse)
+      .map((p) => p.toolResponse!)
+      .toList();
```
Ehesp: I believe you can remove this `toList()`, as the contents of the iterable doesn't need to be materialised into a List, so for really big messages this would be faster and less memory overhead.

packages/genkit_vertexai/lib/src/vertex_api_client.dart
```diff
@@ -190,4 +244,86 @@ class VertexAiPluginImpl extends CommonGoogleGenPlugin {
...
+  }) async {
+    try {
+      final res = await service.listPublisherModels(
+        projectId: _getResolvedProjectId,
+        publisher: publisher,
+      );
+      return (res['publisherModels'] as List?) ?? [];
+    } catch (e, stack) {
+      if (!warnOnFailure) rethrow;
+      logger.fine('Failed to list $publisher models: $e', e, stack);
```
cabljac: should this be logger.warn or something?

packages/genkit_vertexai/lib/src/meta_model.dart
```diff
@@ -0,0 +1,364 @@
...
+    usage: _fromMetaUsage(response['usage'] as Map<String, dynamic>?),
+  );
+}
+
+ModelResponse fromMetaChatCompletionChunks(List<Map<String, dynamic>> chunks) {
+  final content = StringBuffer();
+  var finishReason = FinishReason.unknown;
+  GenerationUsage? usage;
+
+  for (final chunk in chunks) {
```
cabljac, line 223: do we need to handle tool calls in here?

packages/genkit_vertexai/lib/src/meta_model.dart
```diff
@@ -0,0 +1,364 @@
...
+      'tools': req.tools!.map(_toMetaTool).toList(),
+    'temperature': ?options.temperature,
+    'top_p': ?options.topP,
+    'max_tokens': ?options.maxTokens,
+    'stop': ?options.stop,
+    'presence_penalty': ?options.presencePenalty,
+    'frequency_penalty': ?options.frequencyPenalty,
+    'seed': ?options.seed,
+    'user': ?options.user,
+    if (isJsonMode && req.output?.schema != null)
```
cabljac: if someone omits a schema but specifies json mode (is this possible?) do we handle that case?

packages/genkit_vertexai/lib/src/meta_model.dart
```diff
@@ -0,0 +1,364 @@
...
+
+import 'dart:convert';
+
+import 'package:genkit/plugin.dart';
+import 'package:schemantic/schemantic.dart';
+
+const _unsupportedGoogleCloudStorageMediaUrlMessage =
+    'Meta models on Vertex AI do not support gs:// media URLs. Use a publicly accessible HTTPS URL or a data URI instead.';
+
+final metaModelInfo = ModelInfo(
```
Ehesp, line 23: Could this be const (if ModelInfo is a const class - not sure)

packages/genkit_vertexai/lib/src/meta_model.dart
CorieW: I think we can improve this, but removing only this `toList()` has a tradeoff.

Without `toList()`, `toolResponses` becomes a lazy `Iterable`. The `isEmpty` check then has to iterate it to find the first matching item, and the later `toolResponses.map(...).toList()` iterates it again from the start to build the returned list. So this avoids the intermediate allocation, but can repeat the filter/map work for non-empty messages.

I think the better version is to map directly to the final returned list, then check whether that result is empty. That removes the intermediate `ToolResponse` list while keeping the conversion to a single pass.

packages/genkit_vertexai/lib/src/meta_model.dart
CorieW: I checked this, but `ModelInfo` does not currently have a `const` constructor, so this needs to stay `final`.

packages/genkit_vertexai/lib/src/vertex_api_client.dart
```diff
@@ -77,15 +78,51 @@ class VertexAiPluginImpl extends CommonGoogleGenPlugin {
     );
   }
 
+  Future<GenerativeLanguageBaseClient> getOpenModelApiClient() async {
```
cabljac, line 51: this seems like a lot of the same logic as in `getApiClient` above no?  Could extract some shared private helper? Otherwise two places to keep in sync

packages/genkit_vertexai/lib/src/meta_model.dart
```diff
@@ -0,0 +1,364 @@
...
+  return result;
+}
+
+Map<String, dynamic> _toMetaMessage(Message message) {
+  return switch (message.role.value) {
+    'system' => {'role': 'system', 'content': message.text},
+    'user' => {'role': 'user', 'content': _toMetaContent(message.content)},
+    'model' => {
+      'role': 'assistant',
+      'content': message.text,
```
cabljac, line 336: if it's only tool_calls, what happens here? will this be `content: ''` or something?

packages/genkit_vertexai/lib/src/vertex_api_client.dart
CorieW: Yeah agree 100%

packages/genkit_vertexai/lib/src/meta_model.dart
CorieW:
```js
{
  'role': 'assistant',
  'content': '',
  'tool_calls': [...]
}
```

PR #268, opened 2026-04-28
feat(vertexai): tuned models (endpoints/YOUR_ENDPOINT_ID)

CorieW:
This pull request adds support for Vertex AI tuned models (custom endpoints) in the `genkit_vertexai` package, allowing users to call models deployed to Vertex AI endpoints. It introduces new methods for creating and resolving tuned model actions, refactors API client creation for flexibility, and includes tests to validate the new functionality.

## Testing
image

### DIFF
--- a/packages/genkit_google_genai/lib/src/common_plugin.dart
+++ b/packages/genkit_google_genai/lib/src/common_plugin.dart
@@ -40,13 +40,19 @@ final commonModelInfo = ModelInfo(
 abstract class CommonGoogleGenPlugin extends GenkitPlugin {
   Future<GenerativeLanguageBaseClient> getApiClient([String? requestApiKey]);
 
+  Future<GenerativeLanguageBaseClient> clientForModel(
+    String modelName,
+    String? apiKey, {
+    String? apiModelName,
+  }) {
+    return getApiClient(apiKey);
+  }
+
   Model createModel(
     String modelName,
     SchemanticType customOptions, {
     String? actionName,
     String? apiModelName,
-    Future<GenerativeLanguageBaseClient> Function(String? apiKey)?
-    getApiClientOverride,
   }) {
     return Model(
       name: '$name/${actionName ?? modelName}',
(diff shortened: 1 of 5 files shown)


packages/genkit_vertexai/lib/src/vertex_api_client.dart
```diff
@@ -77,6 +83,20 @@ class VertexAiPluginImpl extends CommonGoogleGenPlugin {
     );
   }
 
+  @override
+  Future<GenerativeLanguageBaseClient> getApiClient([
+    String? requestApiKey,
+  ]) async {
+    return _createVertexApiClient(
+      apiVersion: 'v1beta1',
```
cabljac: nit: we're using v1beta1 a lot across the whole thing, can we make this a const

packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -40,9 +40,16 @@ final commonModelInfo = ModelInfo(
...
   Future<GenerativeLanguageBaseClient> getApiClient([String? requestApiKey]);
 
-  Model createModel(String modelName, SchemanticType customOptions) {
+  Model createModel(
+    String modelName,
+    SchemanticType customOptions, {
+    String? actionName,
+    String? apiModelName,
+    Future<GenerativeLanguageBaseClient> Function(String? apiKey)?
+    getApiClientOverride,
```
cabljac: getApiClientOverride is vertex specific right? it's because of the `/endpoints/...` so a different client is needed.

I think we could have an overridable `clientForModel` method (or similar) instead of a callback param here

packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -92,7 +99,9 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
           toolConfig = toGeminiToolConfig(options.functionCallingConfig);
         }
 
-        final service = await getApiClient(apiKey);
+        final service =
```
cabljac: here we could do 

`final service = await clientForModel(...)`

and then vertex can override it. 

If this makes sense? I can provide a PR into this branch to demonstrate if i'm not being clear

CorieW:
- [ ] Tested changes

PR #277, opened 2026-05-06
feat(schemantic): widen analyzer dependency range to include v13

pavelgj:


kevmoo:
@pavelgj – make sure `pkg:analyzer` actually resolves to latest! There might be other packages blocking it and there ARE breaking changes!

pavelgj:
> @pavelgj – make sure `pkg:analyzer` actually resolves to latest! There might be other packages blocking it and there ARE breaking changes!

Good call. Tested with 13.0.0, all tests pass!

kevmoo (approved):

PR #278, opened 2026-05-06
feat(vertexai): add Virtual Try-On model support

CorieW:
Adds Virtual Try On model to VertexAI plugin.

## Testing
**The following puts a shirt (image A) onto a person (image B):**

Image A
image

Image B
image

Result
image

Params
image

### DIFF
--- a/packages/genkit_vertexai/lib/src/virtual_try_on.dart
+++ b/packages/genkit_vertexai/lib/src/virtual_try_on.dart
@@ -1,0 +1,239 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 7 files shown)


packages/genkit_vertexai/lib/src/virtual_try_on.dart
```diff
@@ -0,0 +1,233 @@
...
+
+ActionMetadata virtualTryOnModelMetadata(String pluginName, String modelName) {
+  return modelMetadata(
+    '$pluginName/$modelName',
+    modelInfo: virtualTryOnModelInfo,
+    customOptions: VirtualTryOnOptions.$schema,
+  );
+}
+
+Model createVirtualTryOnModel({
```
cabljac: I think this should probably be an instance method rather than a top level function?

cabljac: on VertexAiPluginImpl probably

CorieW:
- [ ] Test changes

PR #282, opened 2026-05-11
fix(ai): preserve middleware in generation

MichaelDoyle:
Middleware is current missing in generate options on traces, which led to a few different fixes in this PR: 

*   **`applyFormat`**: Fixed to ensure that when a request is formatted (e.g., adding instructions), the `use` property (middleware references) is copied over to the new options object.
 *   **Multi-turn Tool Loop**: Fixed the recursive generation calls during tool execution to ensure the `use` property is preserved in the options passed to subsequent model turns.
 *   **Resume/Interrupt Flow**: Fixed the logic that restarts tools after an interrupt to preserve the original middleware references when the flow resumes.
 *   **Action Resolution**: Corrected the top-level `generate` action to properly resolve and initialize middleware when they are provided as references within the `GenerateActionOptions` object.

### DIFF
--- a/packages/genkit/lib/src/ai/formatters/formatters.dart
+++ b/packages/genkit/lib/src/ai/formatters/formatters.dart
@@ -124,6 +124,7 @@ GenerateActionOptions applyFormat(
     returnToolRequests: request.returnToolRequests,
     maxTurns: request.maxTurns,
     stepName: request.stepName,
+    use: request.use,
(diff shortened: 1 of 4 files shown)


pavelgj (approved):

pavelgj:
overall LGTM. Gemini comments are not unreasonable... but minor.

PR #285, opened 2026-05-20
fix(genkit_chrome): handle `params` not being defined

kevmoo:
Removes a chrome warning at runtime


tomayac (approved):
LGTM, thanks for the fix

pavelgj (approved):

PR #289, opened 2026-05-27
feat(ai): add top-level `system` parameter to generate()

### LINKED ISSUE #288: Add top-level `system` parameter to `generate()`
## Summary

`generate()` in genkit-dart does not accept a top-level `system` parameter, despite this being a first-class option in the [Genkit JS](https://github.com/firebase/genkit/blob/main/js/ai/src/generate.ts) and [Genkit Go](https://github.com/firebase/genkit/blob/main/go/ai/option.go) APIs. Users currently have to manually construct a `Message` with `Role.system` and prepend it to the `messages` list, which is verbose and easy to get wrong.

## Current state

Neither [`Genkit.generate`](https://github.com/genkit-ai/genkit-dart/blob/main/packages/genkit/lib/src/genkit_class.dart#L578) nor [`lite.generate`](https://github.com/genkit-ai/genkit-dart/blob/main/packages/genkit/lib/lite.dart#L41) exposes a `system` parameter. The same is true for `generateStream` on both surfaces, and for `generateHelper` in `src/ai/generate.dart`.

The infrastructure already supports system messages, just not as a top-level convenience param:
- `Role.system` is defined in [`src/types.dart`](https://github.com/genkit-ai/genkit-dart/blob/main/packages/genkit/lib/src/types.dart#L296)
- Dotprompt rendering already injects system messages internally (see [`src/ai/prompt.dart`](https://github.com/genkit-ai/genkit-dart/blob/main/packages/genkit/lib/src/ai/prompt.dart))

## Cross-SDK comparison

**JS** ([`generate.ts`](https://github.com/firebase/genkit/blob/main/js/ai/src/generate.ts#L132)):
```ts
system?: string | Part | Part[];
```
Implementation prepends a system-role message before `messages` and `prompt`.

**Go** ([`option.go`](https://github.com/firebase/genkit/blob/main/go/ai/option.go#L474)):
```go
ai.WithSystem("You are a helpful assistant")
ai.WithSystemFn(func(ctx, input) (string, error) { ... })
```

**Dart (today)** — no equivalent. You must do:
```dart
await ai.generate(
  messages: [
    Message(role: Role.system, content: [TextPart(text: 'You are...')]),
    Message(role: Role.user,   content: [TextPart(text: 'Hello')]),
  ],
);
```

## Proposed API

Add a `String? system` parameter (matching the existing `String? prompt` style) to:
- `Genkit.generate` and `Genkit.generateStream`
- `lite.generate` and `lite.generateStream`
- `generateHelper`

Behavior: when `system` is provided, a `Message(role: Role.system, content: [TextPart(text: system)])` is prepended to the resolved message list, before any `messages` and before the synthesized `prompt` user message — matching JS semantics.

```dart
await ai.generate(
  model: googleAI.model('gemini-2.5-flash'),
  system: 'You are a helpful pirate.',
  prompt: 'Tell me about Dart.',
);
```

## Scope

Text-only (`String?`) for this first pass, mirroring how `prompt` is currently typed in Dart. Multi-part system content (`Part | Part[]` like JS) can be a follow-up if needed.

## Acceptance criteria

- [ ] `system` param added to `Genkit.generate`, `Genkit.generateStream`, `lite.generate`, `lite.generateStream`, and `generateHelper`
- [ ] System message is prepended to the message list before `messages`/`prompt`
- [ ] Tests cover: `system` alone with `prompt`, `system` combined with explicit `messages`, ordering of system → messages → prompt user message
- [ ] Existing tests still pass

chrisraygill:
## Summary

- Adds a `String? system` parameter to `Genkit.generate`, `Genkit.generateStream`, `lite.generate`, `lite.generateStream`, and `generateHelper`.
- When provided, prepends a `Message(role: Role.system, content: [TextPart(...)])` before any explicit `messages` and the synthesized `prompt` user message, matching the [JS](https://github.com/firebase/genkit/blob/main/js/ai/src/generate.ts#L132) and [Go](https://github.com/firebase/genkit/blob/main/go/ai/option.go#L474) semantics.
- Callers no longer need to construct a `Role.system` message by hand.

Closes #288

## Example

```dart
await ai.generate(
  model: googleAI.model('gemini-2.5-flash'),
  system: 'You are a helpful pirate.',
  prompt: 'Tell me about Dart.',
);
```

## Implementation notes

- The validation in `generateHelper` is relaxed to also allow `system` alone (previously required `prompt` or `messages`).
- `prompt.dart`'s dotprompt flow is unchanged — it already injects system messages via `_renderSystem` and passes the full message list to `generateHelper`.
- Text-only (`String?`) for this first pass, matching how `prompt` is typed. Multi-part system content (the JS `Part | Part[]` form) can be a follow-up.

## Test plan

- [x] New `system parameter` group in `test/ai/generate_test.dart` covering: `system` + `prompt`, `system` + explicit `messages`, ordering of `system` → `messages` → `prompt` user message, and `system` alone.
- [x] New lite-level test in `test/lite_test.dart` asserting the prepended system message reaches the model.
- [x] `dart test` — all 363 tests pass.
- [x] `dart analyze` clean on changed files.
- [x] `dart format` applied.


packages/genkit/lib/lite.dart
```diff
@@ -39,6 +39,7 @@ export 'src/schema_extensions.dart';
 export 'src/types.dart';
 
 Future<GenerateResponseHelper> generate<C>({
+  String? system,
```
kevmoo, line 42: Is this documented anywhere?

chrisraygill:
Docs PR for the new parameter: genkit-ai/docsite#281.

packages/genkit/lib/lite.dart
chrisraygill: Have a PR to fix a docs bug related to it: https://github.com/genkit-ai/docsite/pull/281

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -621,8 +622,9 @@ Future<GenerateResponseHelper> generateHelper<CustomOptions>(
   /// List of tool requests to restart during an interrupted generation session.
   List<ToolRequestPart>? restart,
 }) async {
-  if (messages == null && prompt == null) {
-    throw ArgumentError('prompt or messages must be provided');
+  final hasSystem = system != null && system.isNotEmpty;
```
pavelgj: is there a reason for special treatment of `system.isNotEmpty`? We don't do that for `prompt`, so it's inconsistent. I think that if there's a good reason to handle empty strings, then we should probably do the same for `prompt`.

pavelgj (approved):

packages/genkit/lib/src/ai/generate.dart
chrisraygill: It's for consistency with the JS implementation - though I think JS does it for both system and prompt. Would you prefer I take that out of this PR or do it for both in this PR?

pavelgj: I'd say take it out. Dart doesn't have the same "falsiness" coersion as JS.

PR #292, opened 2026-06-03
refactor: split schemantic into runtime and schemantic_builder packages

pavelgj:
Extract code generation logic from `schemantic` into a new `schemantic_builder` package, keeping `schemantic` as a lightweight runtime-only dependency. Add `schemantic_builder` as a dev dependency across all consuming packages and update build configuration and docs accordingly.


packages/schemantic/pubspec.yaml
```diff
@@ -8,14 +8,8 @@ environment:
...
-  code_builder: ^4.10.0
-  dart_style: ^3.1.4
   json_schema_builder: ^0.1.3
-  source_gen: ^4.2.0
 
 dev_dependencies:
-  build_test: ^3.5.5
-  logging: ^1.3.0
+  schemantic_builder: ^0.1.3
   test: ^1.29.0
```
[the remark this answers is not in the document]
pavelgj: it needs it for examples

packages/schemantic_builder/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
+name: schemantic_builder
+description: A code generator for schemantic that produces type-safe data classes and JSON schemas from abstract definitions.
+repository: https://github.com/genkit-ai/genkit-dart
+license: Apache-2.0
+resolution: workspace
+version: 0.1.3
```
jakemac53: Should this not just start at `0.1.0` given it's a new package?

kevmoo (approved):
I did a the full scan. Seems all reasonable. So glad we're doing this.

packages/schemantic_builder/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
...
+environment:
+  sdk: ^3.10.0
+
+dependencies:
+  analyzer: '>=10.0.0 <14.0.0'
+  build: ^4.0.4
+  code_builder: ^4.10.0
+  dart_style: ^3.1.4
+  schemantic: ^0.1.3
```
jakemac53: So this is actually a bit tricky, but you probably want to actually have a tighter constraint than this.

Lets say you release a new version of schematic which adds a new annotation or field to an annotation, or whatever. That is not breaking for schematic, but the builder package at the old version will not support it.

So, the builder package should most likely depend on the _feature release_ range of schematic, not the breaking change release range. In this case because of the pre-release version shifting its something like the following:
```suggestion
  schemantic: ">=0.1.3 <0.1.4"
```

jakemac53: This does mean a new release of schemantic_builder whenever you have a new feature release of schematic.  But that is very likely what you require anyways.

packages/schemantic_builder/pubspec.yaml
pavelgj: I was thinking we release `schemantic` and `schemantic_builder` under the same version, always in sync.

packages/schemantic_builder/pubspec.yaml
pavelgj: As I commented above, I was thinking we release schemantic and schemantic_builder under the same version, always in sync. To avoid all this.

jakemac53: If you do that, then you will need to pin this version (remove the `^`)

packages/schemantic_builder/pubspec.yaml
jakemac53: That may force breaking changes in one package or the other that are not necessary, so it's probably not what I would do. And just in general it could mean versions that don't really follow semver properly.

You can pin without keeping exactly in sync versions as well.

pavelgj: Ah, interesting... TIL -- the practice of locking the versions of multiple packages is common in NPM ecosystem, but apparently not so much in Dart... so maybe my thinking is not idiomatic. I think your proposal is the right approach.

PR #297, opened 2026-06-09
fix(genkit_openai): send non-image media as OpenAI file content parts

irangarcia:
The content-part converter mapped every MediaPart to an image_url part, so PDFs and other non-image documents were rejected by the OpenAI API with "Invalid MIME type. Only image types are supported."

Branch non-image data URIs to ContentPart.fileData (the OpenAI `file` content part). OpenAI infers the document type from the filename extension, which Genkit's Media does not carry, so synthesize one from the MIME type. Image and remote-URL handling are unchanged.

### DIFF
--- a/packages/genkit_openai/lib/src/converters.dart
+++ b/packages/genkit_openai/lib/src/converters.dart
@@ -101,15 +101,29 @@ abstract final class GenkitConverter {
     }
     if (part.isMedia) {
       final media = part.media!;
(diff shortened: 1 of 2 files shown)


pavelgj:
Thanks for PR.  Overall LGTM. Could you please fix the formating?

pavelgj (approved):

PR #310, opened 2026-06-25
feat(agents): agent & session schema types (1/8)

pavelgj:
First branch in the stacked **agents** feature for Genkit Dart (see the stack overview at the bottom).

This PR adds the foundational schema/type layer only — no runtime behavior.

### What's in this PR
- Agent + session wire types in `packages/genkit/lib/src/types.dart`:
  `$AgentInit`, `$AgentInput`, `$AgentResume`, `$AgentOutput`, `$AgentResult`,
  `$AgentStreamChunk`, `$Artifact`, `$JsonPatchOperation`, `$SessionSnapshot`,
  `$SessionState`, plus `AgentFinishReason` / `SnapshotStatus` enums.
- Abort/metadata wire types (`$AgentAbortRequest`, `$AgentAbortResponse`, `$AgentMetadata`).
- `ToolRequest.input` widened to `any` (`z.unknown().optional()`) to match the
  finalized JS source of truth; plugin call sites cast back to `Map` for the model SDKs.
- Renamed generated `AgentError` → `AgentErrorInfo` to avoid a collision with the
  client-side `AgentError` exception.
- Schema-generator support for local-file / CLI schema sources, pointed at `main`.
- Regenerated `types.g.dart`.

### Testing
- `dart analyze` clean (core + all plugins).
- Generated code verified in sync (`melos run build-gen` produces no diff).

### Stack
main → **pj/agent-types** → pj/agent-session → pj/agent-file-store → pj/agent-core →
pj/agent-agent → pj/agent-remote → pj/agent-conformance → pj/agent-sample-app

Review/merge bottom-to-top.


packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -265,7 +265,7 @@ sdk.InputMessage toAnthropicMessage(Message m) {
         sdk.InputContentBlock.toolUse(
           id: req.ref ?? '',
           name: req.name,
-          input: req.input ?? {},
+          input: (req.input as Map<String, dynamic>?) ?? {},
```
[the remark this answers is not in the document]
pavelgj: Done — replaced the unchecked cast with an `is Map` guard that falls back to an empty `<String, dynamic>{}` map, so a non-map `input` no longer throws.

packages/genkit_firebase_ai/lib/genkit_firebase_ai.dart
```diff
@@ -492,15 +492,15 @@ fai.Part toGeminiPart(Part p) {
     try {
       return fai.FunctionCall.forTest(
         toolRequest.name,
-        toolRequest.input ?? {},
+        (toolRequest.input as Map<String, Object?>?) ?? {},
```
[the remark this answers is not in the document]
pavelgj: Done — guarded with `toolRequest.input is Map ? (toolRequest.input as Map).cast<String, Object?>() : <String, Object?>{}`.

packages/genkit_firebase_ai/lib/genkit_firebase_ai.dart
```diff
@@ -492,15 +492,15 @@ fai.Part toGeminiPart(Part p) {
...
         id: toolRequest.ref,
         isThought: isThought,
         thoughtSignature: thoughtSignature,
       );
     } catch (_) {
       return fai.FunctionCall(
         toolRequest.name,
-        toolRequest.input ?? {},
+        (toolRequest.input as Map<String, Object?>?) ?? {},
```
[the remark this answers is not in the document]
pavelgj: Done — same `is Map` guard applied to this fallback `FunctionCall` path as well.

packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -448,8 +448,9 @@ gcl.Part toGeminiPart(Part p) {
       functionCall: gcl.FunctionCall(
         id: p.toolRequest!.ref ?? '',
         name: _toGeminiToolName(p.toolRequest!.name),
-        args: p.toolRequest!.input, // already a map
+        args: p.toolRequest!.input as Map<String, Object?>?, // always a map
```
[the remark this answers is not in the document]
pavelgj: Done — now `p.toolRequest!.input is Map ? (...).cast<String, Object?>() : null`.

packages/genkit/lib/src/types.dart
```diff
@@ -430,3 +433,178 @@ abstract class $ReflectionStreamChunkParams {
...
+}
+
+/// Agent capability metadata placed under `metadata.agent` on an agent's action
+/// descriptor.
+///
+/// Lets the Dev UI and other reflective callers render the right surface (e.g.
+/// hide the Abort button when the configured store doesn't support it) without
+/// round-tripping through the reflection API.
+@Schema()
+abstract class $AgentMetadata {
```
cabljac: This file is generated by the generate.dart right, I think when you re-run the generate it removes them? If they're Dart-only and not in the _allowlist

cabljac: could we move them to a non-generated file?

packages/genkit/lib/src/types.dart
```diff
@@ -430,3 +433,178 @@ abstract class $ReflectionStreamChunkParams {
...
+  bool get abortable;
+
+  /// JSON schema for the agent's custom session state (the `custom` field of
+  /// [$SessionState]). Omitted when the state type carries no schema to infer.
+  Map<String, dynamic>? get stateSchema;
+}
+
+@Schema()
+abstract class $JsonPatchOperation {
+  JsonPatchOp get op;
```
cabljac, line 555: if i'm understanding things correctly, shouldn't typeOverrides be forcing these to `String` instead of extension types? I think there is drift between this committed file and the overrides? Re-running the generator changes these to String

packages/genkit/lib/src/types.dart
pavelgj: I think this was a bad merge... some changes to generator got reverted. This file is correct, updated the generator now.

packages/genkit/lib/src/types.dart
pavelgj: generator is fixed.

packages/_schema_generator/lib/src/class_generator.dart
```diff
@@ -24,6 +24,20 @@ const typeOverrides = {
...
+  // `input` is `z.unknown().optional()` on the wire; keep it `Object?` so any
+  // JSON value (map/list/scalar) round-trips instead of being cast to a map.
+  'ToolRequest': {'input': 'Object'},
+  // Agent types: inline objects promoted to named defs, the JsonPatch array
+  // alias, and open `{}` schemas that should be `Object?`.
+  'AgentInput': {'resume': '\$AgentResume'},
+  'AgentOutput': {'error': '\$AgentErrorInfo'},
+  'SessionSnapshot': {'error': '\$AgentErrorInfo'},
+  'AgentStreamChunk': {'customPatch': 'List<\$JsonPatchOperation>'},
+  'SessionState': {'custom': 'dynamic'},
```
apascal07: `custom` is required in SessionState's schema, but the `dynamic` override path a few lines down returns before the isRequired handling, so it comes out as an optional constructor arg that's dropped when null. A SessionState built without it then emits JSON missing a field its own schema marks required, which a strict validator would reject. The Object-typed overrides are all optional so only this dynamic+required one is affected.

apascal07 (approved):

cabljac (approved):

packages/_schema_generator/lib/src/class_generator.dart
pavelgj: Good catch — you're right. Because `custom` was a non-nullable `dynamic` getter, the schemantic builder added it to the required list even though upstream `genkit-schema.json` (and JS `z.any().optional()`) treat it as optional, so a `SessionState` serialized without `custom` emitted schema-invalid JSON. Switched the override to `Object?` (which the builder treats as optional, like the other `Object` overrides) and regenerated; `required: ['custom']` is now gone. Fixed in 05489d93 and cascaded up the stack.

PR #311, opened 2026-06-25
feat(agents): session & snapshot storage + JSON Patch (2/8)

pavelgj:
Builds on #310. Adds the browser-safe session/snapshot storage layer,
ported from the finalized JS `session.ts`, plus a JSON Patch implementation.

### What's in this PR
- `lib/src/ai/agents/session.dart`: `Session`, `InMemorySessionStore`, snapshot
  chaining (linear + branched leaf resolution), sessionId/snapshotId helpers,
  `onSnapshotStateChange` notifications.
- `lib/src/ai/agents/json_patch.dart`: RFC 6902 diff/apply.
- Aligned the session store with the finalized JS model and threaded request
  context through `SessionStore` reads/writes.
- Hardening: deep-clone session init state so it can't alias the caller, and
  reject prototype-pollution tokens (`__proto__` etc.) in JSON Pointer paths.

### Testing
- `dart test test/ai/agents/session_test.dart` and `json_patch_test.dart` green.
- `dart analyze` clean.

> Stacked PR — base branch is `pj/agent-types`. Review after the layer below.


packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -0,0 +1,532 @@
...
+/// Zone key under which the active [Session] is stored during an agent turn.
+const Object _sessionZoneKey = #ai.session;
+
+/// Deep-clones a JSON-serializable value (`Map`, `List`, or primitive).
+Object? _deepClone(Object? value) {
+  if (value is Map) {
+    return <String, dynamic>{
+      for (final entry in value.entries)
+        entry.key as String: _deepClone(entry.value),
+    };
```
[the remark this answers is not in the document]
pavelgj: Done — switched to `entry.key.toString()` so non-string keys no longer trigger a runtime `TypeError`.

packages/genkit/lib/src/ai/agents/json_patch.dart
```diff
@@ -0,0 +1,360 @@
...
+
+/// Deep clone of a JSON-serializable value.
+T _clone<T>(T value) => _cloneValue(value) as T;
+
+Object? _cloneValue(Object? value) {
+  if (value is Map) {
+    return <String, dynamic>{
+      for (final entry in value.entries)
+        entry.key as String: _cloneValue(entry.value),
+    };
```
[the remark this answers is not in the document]
pavelgj: Done — using `entry.key.toString()` instead of `as String`.

packages/genkit/lib/src/ai/agents/json_patch.dart
```diff
@@ -0,0 +1,360 @@
...
+  if (_deepEqual(from, to)) return;
+
+  // Both plain objects - recurse member by member.
+  if (_isObject(from) && _isObject(to)) {
+    final fromMap = from as Map;
+    final toMap = to as Map;
+    final keys = <String>{
+      ...fromMap.keys.cast<String>(),
+      ...toMap.keys.cast<String>(),
+    };
```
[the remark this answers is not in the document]
pavelgj: Done — switched to `whereType<String>()` so non-string keys are skipped rather than throwing.

packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -0,0 +1,542 @@
...
+  }
+
+  /// Overwrites the session messages.
+  void setMessages(List<Message> messages) {
+    _json['messages'] = messages.map((m) => m.toJson()).toList();
+    _version++;
+  }
+
+  /// Retrieves the custom state of the session (plain JSON).
+  dynamic getCustom() => _json['custom'];
```
cabljac, line 243: `getCustom()` returns the live internal map rather than a copy.

Matches JS (`session.ts` does the same, so correct parity), but i think leaves a quiet back door: mutating the returned value bypasses the version bump and `customChanged`, so no `customPatch` gets emitted upstack and client state silently diverges.

```dart
final custom = session.getCustom() as Map;
custom['count'] = 999; // state changed, no version bump, no event
```

Given `getState()`/`getMessages()` both return copies, my guess is this deserves a doc note rather than a behaviour change, something like "returns the live value, treat as read-only, mutate via `updateCustom`". Thoughts?

packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -0,0 +1,542 @@
...
+      'leaves), so there is no single latest snapshot. This happens when a '
+      'conversation is branched (e.g. regenerate). Resume by snapshotId '
+      'instead.',
+      status: StatusCodes.FAILED_PRECONDITION,
+    );
+  }
+
+  // Default: pick the most-recently created leaf. `createdAt` is an ISO-8601
+  // timestamp, so lexicographic comparison matches chronological order.
+  return leaves.reduce(
```
cabljac: I think the lexicographic `createdAt` comparison here can pick the wrong leaf in a couple of edge cases. I wrote some quick tests locally to check I wasn't imagining it:

1. On an exact timestamp tie, `compareTo > 0` keeps the first-inserted leaf, so "latest" resolves to the older branch. Probably unhittable on the Dart VM? but since this module is browser-safe and web `DateTime.now()` is millisecond precision, two quick saves (regenerate spam etc.) could plausibly collide there.

2. If writers use different ISO-8601 flavours the ordering breaks down, e.g. `'...T10:00:00.000+05:00'` (05:00 UTC) sorts above `'...T09:00:00.000Z'` (09:00 UTC) despite being four hours earlier. Since `SessionStore` is a public interface and `createdAt` comes from whoever builds the snapshot, mixed formats feel possible once third-party stores exist.

Neither case throws, which is what made me want to flag it, the wrong branch would just load quietly. Would it be worth a note on `SessionSnapshot.createdAt` pinning the expected format (UTC, fixed precision)? And maybe an insertion-order tie-breaker in `InMemorySessionStore`? May well be fine as-is if branched sessions plus same-millisecond saves are considered too rare to matter, happy to go with your read.

packages/genkit/lib/src/ai/agents/session.dart
pavelgj: Good call — added a doc note on `getCustom()` spelling out that it returns the live internal value and must be treated as read-only, and that custom state should be mutated via `updateCustom` (otherwise the version bump / `customChanged` / `customPatch` are skipped and client state diverges). Kept the behavior as-is for JS parity.

packages/genkit/lib/src/ai/agents/session.dart
pavelgj: Nice find on both cases — fixed in code rather than just documenting:

- **Mixed offsets:** `selectLeafSnapshot` now parses `createdAt` with `DateTime.tryParse(...).toUtc()` and compares the resulting instants, so `...+05:00` vs `...Z` order chronologically instead of lexicographically. If a timestamp is unparseable it falls back to a lexicographic compare for that pairing, so we never throw on malformed input.
- **Exact tie:** switched the reduce to keep the *later* element in iteration order (effectively `>= 0`). Since `owned` is built by iterating the insertion-ordered `_snapshots` map, a same-millisecond tie now resolves to the most-recently saved leaf rather than the oldest.

I also documented the expected `createdAt` format (UTC ISO-8601) at the point of use in `selectLeafSnapshot`. (`SessionSnapshot.createdAt` itself lives in the generated `types.dart`, which is regenerated from the upstream schema, so a dartdoc there would be overwritten.)

Added tests covering mixed-offset ordering, the exact-tie insertion-order tie-break, and the unparseable-timestamp no-throw path.

packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -0,0 +1,572 @@
...
+      cb(arg);
+    }
+  }
+
+  /// Returns a deep copy of the current session state.
+  SessionState getState() =>
+      SessionState.fromJson(_deepClone(_json) as Map<String, dynamic>);
+
+  /// Retrieves all messages associated with the session.
+  List<Message> getMessages() =>
```
apascal07, line 219: `getMessages()` hands back Message objects wrapping the live internal maps, and Message's setters write through, so a caller can mutate session state here with no version bump or event. JS returns a structuredClone. A deep clone before decoding would match (getState already clones, and getArtifacts is intentionally live per its doc, so it's just this one that drifted).

packages/genkit/lib/src/ai/agents/json_patch.dart
```diff
@@ -0,0 +1,371 @@
...
+List<String> _parsePointer(String pointer) {
+  if (pointer == '') return [];
+  if (!pointer.startsWith('/')) {
+    throw ArgumentError(
+      'Invalid JSON Pointer: "$pointer" must start with "/".',
+    );
+  }
+  final tokens = pointer.substring(1).split('/').map(_unescapeToken).toList();
+  for (final token in tokens) {
+    if (_forbiddenTokens.contains(token)) {
```
apascal07, line 68: Since `diff` doesn't filter these tokens but `applyPatch` rejects them, a legit user key named `constructor` or `prototype` in custom state round-trips fine through diff and then throws here on apply. Matches JS, but this is arbitrary user JSON, so worth confirming it's intended and documenting the reserved keys. (Putting the check on the apply/untrusted side is the right call for the pollution defense itself.)

apascal07 (approved):

packages/genkit/lib/src/ai/agents/session.dart
pavelgj: Good catch, fixed in 701c2e48. `getMessages()` now deep-clones the underlying maps before decoding (mirroring the JS `structuredClone`), so the returned `Message` objects can no longer write through to session state. Left `getCustom`/`getArtifacts` live per their docs. Added a regression test asserting a mutation on a returned message does not leak back and does not bump the version.

packages/genkit/lib/src/ai/agents/json_patch.dart
pavelgj: Confirmed intended: it matches JS, and keeping the check on the apply/untrusted side is deliberate. Documented it in 701c2e48 - the `applyPatch` doc comment now explicitly calls out that `__proto__`/`prototype`/`constructor` are reserved and rejected as a prototype-pollution defense, and notes the diff-vs-apply asymmetry (a doc that legitimately contains such a key round-trips through `diff` but throws on `applyPatch`).

PR #312, opened 2026-06-25
feat(agents): dart:io FileSessionStore (3/8)

pavelgj:
Builds on #311. Adds a filesystem-backed session store, ported from JS
`session-stores.ts`.

### What's in this PR
- `lib/src/ai/agents/session_io.dart`: `FileSessionStore` (persists snapshots as
  `<prefix>/<id>.json`, sessionId leaf resolution, multi-tenant `snapshotPathPrefix`,
  `maxPersistedChainLength` trimming, snapshotId path-traversal safety, file-watch
  `onSnapshotStateChange`).
- Exposed via a new browser-unsafe `package:genkit/io.dart` entrypoint so the main
  `genkit.dart` import stays web-safe.

### Testing
- `dart test test/ai/agents/session_io_test.dart` green.
- `dart analyze` clean.

> Stacked PR — base branch is `pj/agent-session`.


packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -0,0 +1,402 @@
...
+      try {
+        final snap = SessionSnapshot.fromJson(
+          jsonDecode(await entry.readAsString()) as Map<String, dynamic>,
+        );
+        if (snapshotSessionId(snap) == sessionId) snapshots.add(snap);
+      } on FileSystemException {
+        // Concurrently removed mid-scan; skip it.
+        continue;
+      }
+    }
```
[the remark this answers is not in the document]
pavelgj: Done — the per-file `on FileSystemException` was widened to a bare `catch (_)` (skips corrupt/malformed/concurrently-deleted files), and the whole directory scan is now wrapped in its own `try/catch` so a listing/access error degrades gracefully instead of throwing.

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -0,0 +1,402 @@
...
+    int maxChain,
+    Map<String, dynamic>? context,
+  ) async {
+    final chain = <String>[];
+    SessionSnapshot? cur = head;
+    while (cur != null) {
+      chain.add(cur.snapshotId);
+      final parentId = cur.parentId;
+      cur = parentId != null ? await _snapshotById(parentId, context) : null;
+    }
```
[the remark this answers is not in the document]
pavelgj: Done — added a `visited` set; the loop is now `while (cur != null && visited.add(cur.snapshotId))`, so a cyclic parent chain breaks instead of looping forever.

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -0,0 +1,402 @@
...
+          jsonDecode(contents) as Map<String, dynamic>,
+        );
+      } catch (_) {
+        // Partially written file mid-rewrite: skip without updating
+        // lastSerialized so the next event/poll re-reads the complete file.
+        return;
+      }
+      lastSerialized = contents;
+      callback(snapshot);
+    }
```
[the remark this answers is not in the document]
pavelgj: Done — `emitIfChanged` now uses an `isReading`/`needsRecheck` non-overlapping loop so concurrent events/poll ticks are processed sequentially without dropping a change.

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -0,0 +1,402 @@
...
+
+    // Watch the directory (not the file) so this still works before the file
+    // exists and survives atomic rename-replace writes that swap the inode.
+    StreamSubscription<FileSystemEvent>? watcher;
+    try {
+      watcher = dir.watch().listen((event) {
+        if (event.path.endsWith(fileName)) unawaited(emitIfChanged());
+      });
```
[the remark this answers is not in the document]
pavelgj: Done — the watcher now compares `event.path == file.path` instead of `endsWith(fileName)`, so sibling files with the same suffix no longer match.

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -0,0 +1,402 @@
...
+/// A `dart:io` file-system backed session snapshot store.
+///
+/// Snapshots are stored as flat JSON files keyed by their `snapshotId`, under an
+/// optional per-tenant sub-directory prefix:
+///
+///     dirPath/<prefix>/<snapshotId>.json
+///
+/// `getSnapshot(sessionId: ...)` scans the prefix directory and selects the
+/// single leaf snapshot whose `state.sessionId` matches - there is no separate
+/// grouping directory, the `sessionId` lives in each snapshot's state.
```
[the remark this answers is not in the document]
pavelgj: Done — documented this on the `FileSessionStore` class doc, noting it is intended for local development/testing and that `getSnapshot` performs an O(N) scan that can bottleneck at scale.

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -0,0 +1,574 @@
...
+    // Maintain the per-session pointer so `sessionId` lookups stay fast (one
+    // pointer read plus one snapshot read). Advance the pointer for a new leaf,
+    // refresh it when rewriting the current leaf, and leave it untouched when
+    // upserting an older (non-leaf) snapshot. Snapshots without a `sessionId`
+    // are not addressable by session, so skip the pointer.
+    final sessionId = snapshotSessionId(result);
+    if (sessionId != null) {
+      final isNew = current == null;
+      final pointer = await _readPointer(sessionId, context);
+      if (isNew || pointer == null || pointer.currentSnapshotId == id) {
```
cabljac: I think the pointer-advance here diverges from JS and also from its own comment. The comment says "leave it untouched when upserting an older (non-leaf) snapshot", and JS only advances on a brand-new snapshot (`if (sessionId && !current)`), but the `pointer == null` branch will write the pointer to a non-leaf's id. If the pointer file is absent (crash between the snapshot write and the best-effort `writePointer`, or a legacy store) and the save is rewriting a non-leaf (e.g. a status/heartbeat update on an older snapshot), it points the session at that older snapshot, so a later `getSnapshot(sessionId:)` fast-path returns it.

Quick repro: chain `s1 ← s2`, delete `.pointers/<sessionId>.json`, rewrite `s1`, then `getSnapshot(sessionId:)` comes back as `s1` instead of `s2`. It self-heals on the next new save, and it's the same staleness family as the lingering-pointer limitation JS documents, though JS's pointer always names a former leaf whereas this can name one that was never a leaf.

Narrow (needs an absent pointer plus a non-leaf rewrite), so maybe not worth chasing, but was the `pointer == null` branch deliberate or should it track JS's `!current`?

pavelgj: Good catch, that was not deliberate. Fixed it to track JS: the pointer is now only advanced for a brand-new snapshot (`current == null`); rewriting an existing snapshot (leaf or not) leaves the pointer untouched, so a non-leaf rewrite with an absent pointer can never aim the session at a non-leaf. Added a regression test covering exactly your repro (`s1 ← s2`, delete the pointer, rewrite `s1`, assert `getSnapshot(sessionId:)` still returns `s2`).

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -0,0 +1,569 @@
...
+  ///   on an interval, backstopping any events the watcher drops.
+  ///
+  /// Callbacks are de-duplicated by serialized content, so the noisy/duplicate
+  /// events the watcher emits collapse into one callback per real change.
+  /// Transient read errors (e.g. a partially written file mid-rewrite, or a
+  /// not-yet-created file) are swallowed; the next event/poll re-reads.
+  ///
+  /// Returns an unsubscribe function that stops watching and polling.
+  @override
+  void Function()? onSnapshotStateChange(
```
apascal07, line 493: This is the one path that builds the watched/polled file straight from `snapshotId` without the `_assertSafeSnapshotId` check every other FS op here goes through, so a `../otherTenant/id` reaches across the prefix. A one-line assert at the top of this method would keep it consistent. JS has the same gap, so probably worth flagging upstream too.

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -0,0 +1,569 @@
...
+  /// store / not yet written) or unreadable / corrupt - callers fall back to a
+  /// full directory scan in that case. Best-effort: any error reading the
+  /// pointer resolves to `null` so the optimization can never make a lookup (or
+  /// save) fail where the scan-only baseline would have succeeded.
+  Future<_PointerDoc?> _readPointer(
+    String sessionId,
+    Map<String, dynamic>? context,
+  ) async {
+    try {
+      final file = _pointerFileFor(sessionId, context);
```
apascal07: `_pointerFileFor` (which validates sessionId) is called inside the try, so an invalid sessionId gets swallowed and silently falls back to a scan instead of failing fast. JS deliberately resolves this outside the try for exactly that reason. Same shape in _writePointer. Hoisting it above the try restores the fail-fast.

apascal07 (approved):

packages/genkit/lib/src/ai/agents/session_io.dart
pavelgj: Good catch, fixed. `onSnapshotStateChange` now calls `_assertSafeSnapshotId(snapshotId)` as its first line, so it goes through the same check as every other FS op and a `../otherTenant/id` can no longer reach across the prefix. Added a regression test covering the unsafe-id cases. Will flag the same gap upstream in JS.

packages/genkit/lib/src/ai/agents/session_io.dart
pavelgj: Fixed. Hoisted the `_pointerFileFor(...)` call (which validates `sessionId`) above the `try` in both `_readPointer` and `_writePointer`, so an invalid `sessionId` now fails fast instead of being swallowed into a silent scan fallback, matching JS. Added a `sessionId safety` regression test.

PR #313, opened 2026-06-25
feat(agents): transport-agnostic client core (4/8)

pavelgj:
Builds on #312. Adds the transport-agnostic agent client core, ported
from JS `agent-core.ts`. Shared by both the in-process server agent and the remote
HTTP client.

### What's in this PR
- `lib/src/ai/agents/agent_core.dart`: `AgentChat`, `createAgentApi`/`AgentApi`,
  `AgentTransport`, `AgentTurn`/`AgentResponse`/`AgentChunk`/`AgentInterrupt`,
  `DetachedTask`, and `CancellationToken` (the Dart stand-in for `AbortSignal`).
- Single-message input model + sessionId tracking/restore from init.
- Response fallbacks and failed/aborted-turn message rollback aligned with JS.
- Adapted to optional snapshot state.
- Pre-aborted bail: `send()`/`sendStream()` short-circuit with `finishReason: aborted`
  when the token is already cancelled, leaving `messages` untouched.
- `AgentInterrupt.restart()` re-issues the original tool input verbatim (supports
  non-Map/list/scalar inputs).

### Testing
- `dart test test/ai/agents/agent_core_test.dart` green.
- `dart analyze` clean.

> Stacked PR — base branch is `pj/agent-file-store`.


packages/genkit/lib/src/ai/agents/agent_core.dart
```diff
@@ -0,0 +1,810 @@
...
+      .toList();
+
+  List<Message> get messages => _messages;
+
+  GenerationUsage get usage {
+    final raw = _raw.toJson()['usage'];
+    return GenerationUsage.fromJson(
+      raw is Map ? raw.cast<String, dynamic>() : <String, dynamic>{},
+    );
+  }
```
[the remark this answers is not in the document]
pavelgj: Done — removed the dead `usage` getter entirely since `AgentOutput` has no `usage` field (it would always have been null).

packages/genkit/lib/src/ai/agents/agent_core.dart
```diff
@@ -0,0 +1,810 @@
...
+      if (snap != null) {
+        yield snap;
+        final status = snap.status;
+        if (status != null && _terminalStatuses.contains(status.value)) {
+          return;
+        }
+      }
+      await Future<void>.delayed(Duration(milliseconds: intervalMs));
+    }
+  }
```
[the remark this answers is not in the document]
pavelgj: Done — `poll` is now bounded by `maxConsecutiveMisses` (default 10) and throws a `StateError` after persistent nulls instead of looping forever.

packages/genkit/lib/src/ai/agents/agent_core.dart
```diff
@@ -0,0 +1,810 @@
...
+  /// Aborts the current snapshot.
+  Future<String?> abort() async {
+    final id = snapshotId;
+    if (id == null) return null;
+    return _transport.abort(id);
+  }
+
+  AgentError _toAgentError(Object e) {
+    if (e is AgentError) return e;
+    final message = e is Error ? e.toString() : e.toString();
```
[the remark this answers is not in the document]
pavelgj: Done — collapsed the redundant ternary to `final message = e.toString();`.

packages/genkit/lib/src/ai/agents/agent_core.dart
```diff
@@ -0,0 +1,806 @@
...
+    int messageCountBeforeTurn,
+  ) async {
+    AgentOutput raw;
+    try {
+      raw = await output;
+    } catch (e) {
+      if (cancel.isCancelled) {
+        raw = AgentOutput(finishReason: AgentFinishReason.aborted);
+      } else {
+        throw _toAgentError(e);
```
apascal07, line 557: On a thrown transport error or non-200 (anything that isn't a structured failed/aborted result), this rethrows before the rollback below runs, so the user message pushed at turn start is left orphaned in `messages` and the next send stacks another one after it. This is the path every remote turn hits on a server error, and it undermines the rollback the failed-result branch does do. The rollback also uses removeRange by count, which is fine for a sequential chat but would clobber a second turn's message if two ever overlap.

packages/genkit/lib/src/ai/agents/agent_core.dart
```diff
@@ -0,0 +1,806 @@
...
+import '../../types.dart';
+import 'json_patch.dart';
+
+// ---------------------------------------------------------------------------
+// Cancellation (Dart has no AbortSignal/AbortController).
+// ---------------------------------------------------------------------------
+
+/// A minimal cooperative cancellation primitive, the Dart stand-in for the
+/// Web's `AbortSignal`/`AbortController`.
+class CancellationToken {
```
apascal07, line 39: No way to unregister a cancellation listener (only the whenCancelled future), so the remote transport's whenCancelled.then handler lingers until the token is GC'd. Harmless for turn-scoped tokens since the guarded ops no-op after completion, but a caller-supplied token reused across turns accumulates handlers.

apascal07 (approved):

packages/genkit/lib/src/ai/agents/agent_core.dart
pavelgj: Good catch — fixed. The non-cancelled catch branch in `_buildResponse` now rolls back the eagerly-pushed user message (`removeRange(messageCountBeforeTurn, ...)`) before rethrowing, so a thrown transport error / non-200 no longer leaves it orphaned or stacks another on the next send. Added regression tests covering both the rollback and a clean follow-up send.

packages/genkit/lib/src/ai/agents/agent_core.dart
pavelgj: Fixed — `CancellationToken` now has `onCancel(callback)` that returns a disposer to unregister, and `cancel()` fans out to registered listeners. A caller-supplied token reused across turns can now detach its per-turn handler instead of accumulating them. Registering on an already-cancelled token runs the callback synchronously with a no-op disposer. `whenCancelled` is retained for existing consumers; the remote transport (later PR) will use `onCancel` + disposer.

PR #315, opened 2026-06-25
feat(agents): HTTP / remote agent client (6/8)

pavelgj:
Builds on #314. Adds the remote agent client and server-side `genkit_shelf`
support, reusing the shared Genkit client.

### What's in this PR
- `lib/src/ai/agents/remote_agent.dart`: `remoteAgent(...)` over the shared client
  (streaming SSE + getSnapshot/abort), with per-request header resolution.
- `init` parameter support on remote actions/agents; `SnapshotEvent` exported.
- Typed `AgentAbortRequest`/`AgentAbortResponse` over the wire.
- Opt out of the unary `run` fast-path; renamed the `state` URL to `getSnapshot`.
- `remoteAgent` takes named args instead of a `RemoteAgentOptions` object.
- `genkit_shelf` server-side agent serving.

### Testing
- `dart test test/ai/agents/remote_agent_test.dart` green.
- `dart analyze` clean (incl. `genkit_shelf`).

> Stacked PR — base branch is `pj/agent-agent`.


packages/genkit/lib/src/ai/agents/remote_agent.dart
```diff
@@ -0,0 +1,222 @@
...
+class _HttpAgentTransport extends AgentTransport {
+  _HttpAgentTransport({
+    required String url,
+    String? getSnapshotUrl,
+    String? abortUrl,
+    HeadersResolver? headers,
+    String? stateManagement,
+    http.Client? httpClient,
+  }) : _headers = headers,
+       _httpClient = httpClient ?? http.Client() {
```
apascal07, line 84: This creates an owned http.Client but there's no close()/dispose() on the transport or AgentApi, so a remoteAgent() that isn't passed a client leaks the connection pool. A server building one per request would slowly exhaust sockets. Only close it when we're the ones who created it (a caller-passed client stays caller-owned).

packages/genkit/test/ai/agents/remote_agent_test.dart
```diff
@@ -0,0 +1,200 @@
...
+      });
+
+      final agent = remoteAgent(url: 'http://host/agent', httpClient: client);
+      final status = await agent.abort('s_z');
+
+      expect(seenUrl, 'http://host/agent/abort');
+      expect(status, 'done');
+    });
+
+    test('resolves headers per request', () async {
```
apascal07: This asserts the resolver was called but not that the resolved header actually reached the request, the mock captures url/body but not headers. A regression that resolves the header then drops it would still pass here. Capturing request.headers and asserting on it (and calling twice to prove it re-resolves per request) would make this meaningful.

apascal07 (approved):

packages/genkit/lib/src/ai/agents/remote_agent.dart
pavelgj: Good catch. Fixed: the transport now tracks whether it created the client (`_ownsClient = httpClient == null`) and `close()` only closes a client we created, so a caller-passed client stays caller-owned. Added `AgentTransport.close()` (no-op default) and delegating `AgentApi.close()` at the core layer (`pj/agent-core`, #313) so callers can `await agent.close()` to release the internal connection pool; the in-process agent is a no-op. Commit d88e0041 (remote) + 6af0bd45 (core).

packages/genkit/test/ai/agents/remote_agent_test.dart
pavelgj: Strengthened the test. The mock client now captures `request.headers`, and the test (`re-resolves headers per request and sends them on the wire`) sends twice on the same chat with a resolver that returns a distinct value per call (`Bearer token-$calls`). It asserts the resolver ran twice, both requests were recorded, and each request carried the correct header on the wire (`requests[0]` -> `token-1`, `requests[1]` -> `token-2`), proving per-request re-resolution rather than a cached header. Commit d88e0041.

PR #318, opened 2026-06-25
feat(google-cloud): add genkit_google_cloud with a FirestoreSessionStore

pavelgj:
Adds a new `genkit_google_cloud` plugin package whose first feature is
`FirestoreSessionStore` — a Firestore-backed `SessionStore` for Genkit agents,
ported from the Genkit JS `FirestoreSessionStore` for full feature parity.

It persists session snapshots as incremental [RFC 6902 JSON Patch](https://datatracker.ietf.org/doc/html/rfc6902)
diffs anchored to periodic, sharded full-state checkpoints, so it scales to
arbitrarily long sessions (long-lived chatbots, coding agents, …) without any
single document approaching Firestore's 1 MiB per-document limit.

Built on [`google_cloud_firestore`](https://pub.dev/packages/google_cloud_firestore) `^0.5.2`.

> Stacked on top of `pj/agent-sample-app` (depends on the agent session APIs).
> Review/merge after its parent.

## What's included

- **`packages/genkit_google_cloud/`** — new workspace plugin (pubspec, LICENSE,
  CHANGELOG, README), added to the root pub workspace.
- **`FirestoreSessionStore`** (`lib/firestore_session_store.dart`):
  - Checkpoint + JSON-Patch-diff chain with configurable `checkpointInterval`
    (default 25) and `shardSize` (default 512 KiB).
  - Read-only transactions for strongly-consistent reconstruction; per-session
    pointer documents for one-round-trip `sessionId` lookups.
  - Oversized-diff promotion to sharded checkpoints; shrinking re-checkpoint
    shard pruning.
  - Per-tenant isolation via `snapshotPathPrefix` (derived from call `context`).
  - Honors `FIRESTORE_EMULATOR_HOST` out of the box.
- **`onSnapshotStateChange` via polling.** The Dart Firestore client has no
  real-time `onSnapshot` listener (the one API gap vs. JS), so the change
  notifier polls and de-dupes by serialized content, with a configurable
  `snapshotWatchPollInterval` (default 2s).
- **Exported session-store authoring helpers from `genkit.dart`**
  (`normalizeGetSnapshotOptions`, `resolveSnapshotId`, `snapshotSessionId`,
  `selectLeafSnapshot`, `assertValidSessionId`, `NormalizedGetSnapshot`) — these
  were package-private, so external `SessionStore` implementations couldn't be
  authored without them. Purely additive.

## Testing

- Emulator-gated integration tests (`test/firestore_session_store_test.dart`,
  17 tests): CRUD, diff chains, checkpoint-interval boundaries, sharding,
  oversized-diff promotion, shard pruning, `sessionId`-via-pointer resolution,
  multi-tenant isolation, argument validation, and the polling notifier.
- The suite is gated on `FIRESTORE_EMULATOR_HOST` and cleanly **skips** when
  unset, so the default `melos run test` needs no infrastructure.
- New **`Firestore Emulator Tests`** CI job in `.github/workflows/dart_ci.yml`
  runs them against a real emulator via
  `firebase emulators:exec --only firestore "dart test"` (separate from the
  existing matrix; adds Java + the Firebase CLI). Verified locally: 17/17 pass.
- `dart analyze --fatal-infos` clean; formatted; license headers applied.

## Notes / behavioral differences vs. JS

- `onSnapshotStateChange` is polling-based (see above) rather than push-based.
- `context` is a `Map<String, dynamic>?` (Dart `SessionStore` convention)
  instead of JS's `SessionStoreOptions`.


packages/genkit_google_cloud/lib/firestore_session_store.dart
```diff
@@ -0,0 +1,870 @@
...
+      tx.set(
+        _snapshotsCol(context).doc(id),
+        (_sanitize(doc.toData()) as Map).cast<String, Object?>(),
+      );
+
+      // Advance the pointer when this is a new leaf, or refresh it when we just
+      // rewrote the current leaf. Upserts of older, non-leaf snapshots leave
+      // the pointer untouched.
+      final isNew = existing == null;
+      if (isNew || pointer == null || pointer.currentSnapshotId == id) {
```
apascal07: The pointer always advances to the new snapshot and _PointerDoc doesn't store createdAt, so a backdated or concurrently-committed older leaf can win and stick, with no scan fallback to recover like InMemory/File have. The Go store gates this on (createdAt, id) for exactly this case. As written it can violate the 'most recently created leaf' contract under concurrency or clock skew.

packages/genkit_google_cloud/lib/firestore_session_store.dart
```diff
@@ -0,0 +1,870 @@
...
+  final int shardSize;
+
+  /// Derives the per-tenant prefix from the call's `context`.
+  final String Function(Map<String, dynamic>? context)? snapshotPathPrefix;
+
+  final Duration _snapshotWatchPollInterval;
+
+  /// Resolves the (per-tenant) prefix for the given call context.
+  String _prefixFor(Map<String, dynamic>? context) =>
+      snapshotPathPrefix?.call(context) ?? _defaultPrefix;
```
apascal07: The tenant prefix from context goes straight into .doc() with no validation. Firestore has no parent traversal so it's not a cross-tenant escape, but an empty prefix throws opaquely mid-transaction and a slash silently nests into a deeper path instead of a clean INVALID_ARGUMENT. The Go store rejects empty / `/` / `.` / `..` here, worth matching.

packages/genkit_google_cloud/lib/firestore_session_store.dart
```diff
@@ -0,0 +1,870 @@
...
+        // state/metadata change. Callers must only upsert the *leaf* -
+        // rewriting a non-leaf snapshot's state would invalidate its
+        // descendants' diffs.
+        if (existing.doc.kind == 'checkpoint') {
+          chain = _writeCheckpoint(
+            tx,
+            id,
+            newState,
+            context,
+            existing.doc.checkpointShardCount,
```
apascal07, line 568: Re-checkpointing prunes trailing shards in place (and promotion nulls a diff's statePatch), which corrupts descendants if a non-leaf snapshot is ever upserted: a later read of a descendant hits deleted shards (DATA_LOSS) or applies an empty patch and silently drops that turn. The 'callers only upsert the leaf' rule is a comment, nothing enforces it. Rejecting an upsert that still has descendants would make a violation fail loud.

apascal07 (approved):

packages/genkit_google_cloud/lib/firestore_session_store.dart
pavelgj: Good catch. Fixed in 92f372d2: `_PointerDoc` now stores `createdAt`, and `saveSnapshot` only advances the pointer to a new leaf when it is strictly newer by `(createdAt, snapshotId)` — the same ordering `selectLeafSnapshot` and the Go store use — so a backdated or concurrently-committed older leaf can no longer clobber a newer one and stick. An in-place leaf refresh still updates. Kept the single pointer-read design (no scan fallback) with a comment explaining why. Added a regression test (`pointer advancement does not regress to a backdated older leaf`).

packages/genkit_google_cloud/lib/firestore_session_store.dart
pavelgj: Fixed in 92f372d2: `_prefixFor` now validates the resolved prefix and throws `INVALID_ARGUMENT` for empty/blank, `/`-containing, and `.`/`..` values (matching the Go store), so an invalid prefix fails fast with a clear error instead of throwing opaquely mid-transaction or silently nesting a deeper path. Added tests covering blank, slash, `.` and `..`.

packages/genkit_google_cloud/lib/firestore_session_store.dart
pavelgj: Fixed in 92f372d2. The upsert path now rejects an upsert of a snapshot in a terminal state (`completed`/`failed`/`aborted`/`expired`) with `FAILED_PRECONDITION`, so the shard-pruning / statePatch-nulling corruption you describe can never reach a snapshot that has descendants. Rationale: the runtime only ever resumes (and thus parents a child onto) a `completed` snapshot, so "has descendants" implies "terminal" — rejecting any terminal-state upsert is a superset of "rejecting an upsert that still has descendants", and needs no extra reads (we already have `existing.doc.status`). Every legitimate upsert (detached `pending → completed`, abort `pending → aborted`) targets a non-terminal snapshot, so this only fails genuine misuse — loudly, before any shard is pruned. Added tests for both the rejection and the allowed `pending → completed` upsert.

PR #319, opened 2026-06-26
feat!: pass middleware context with GenkitAI to factories

pavelgj:
Middleware factories created via `defineMiddleware` now receive a __context__ argument carrying an ephemeral `GenkitAI` instance, in addition to their config. This lets a middleware run its own AI operations (`generate`, `generateStream`, `embed`, etc.) at request time.

The primary motivation is to enable middleware that needs to call the model itself — for example an AI risk/safety classifier that runs a `generate` to score or filter a request or response, a summarization/guardrail middleware, or a routing middleware that asks a model to pick a downstream model. Previously the factory only received config, so a middleware had no way to run AI operations.

As a secondary benefit, the same `GenkitAI` exposes the active registry (`ctx.ai.registry`), so a middleware can also resolve other registered actions (models, tools, agents, etc.) by name — e.g. porting the JS `agents` middleware, which resolves sub-agents from the registry by name (`/agent/<name>`).

### `GenkitAI`: a registry-only AI veneer

To make the context useful without handing middleware the full framework entry point, this PR introduces a new `GenkitAI` base class that mirrors the Genkit JS `GenkitAI` pattern:

- `GenkitAI` holds only a `Registry` and exposes the model-orchestration veneer: `generate`, `generateStream`, `generateBidi`, `embed`, `embedMany`, and `run`.
- Because it only needs a registry (which holds all the state), `GenkitAI` is cheap to construct as an ephemeral, throwaway instance.
- `Genkit` now `extends GenkitAI`, adding the framework features on top: plugin loading, the reflection server, and the various `define*` methods. The model-orchestration methods are inherited from `GenkitAI`, so the public `Genkit` API is unchanged.
- The generate pipeline constructs an ephemeral `GenkitAI(registry)` from the active registry and passes it to middleware factories via the context.

### What changed

- New `base class GenkitAI` (`src/genkit_ai.dart`) holding a `Registry` and the model-orchestration veneer (`generate`, `generateStream`, `generateBidi`, `embed`, `embedMany`, `run`).
- `Genkit` now `extends GenkitAI`; the veneer methods moved out of `Genkit` into `GenkitAI`.
- New record type `GenerateMiddlewareContext = ({GenkitAI ai})`.
- `GenerateMiddlewareDef.create` and the `defineMiddleware(create: ...)` factory signature now take a second positional argument of that type.
- The generate pipeline passes an ephemeral `GenkitAI(registry)` through when instantiating middleware.
- `GenkitAI` and `GenerateMiddlewareContext` are exported from `package:genkit/genkit.dart`.
- Updated built-in middlewares (`retry`, `skills`, `tool_approval`, `filesystem`) and tests.

### ⚠️ BREAKING CHANGE

The `create` callback passed to `defineMiddleware` now takes a __second positional parameter__ (the context). Any custom middleware that calls `defineMiddleware` must update its `create` callback signature, or it will fail to compile.

Note: `config` also changes from an *optional positional* (`[config]`) to a *required positional* (`config`), since Dart cannot declare a required positional after an optional one. You don't need to pass it — the framework always does — but the lambda must declare it positionally.

### Migration

Add a second parameter to your `create` callback. If you don't need it, name it `ctx` (or `_`) and ignore it.

__Before__

```dart
defineMiddleware<MyOptions>(
  name: 'my_mw',
  configSchema: MyOptions.$schema,
  create: ([MyOptions? config]) => MyMiddleware(config),
);
```

__After__

```dart
defineMiddleware<MyOptions>(
  name: 'my_mw',
  configSchema: MyOptions.$schema,
  create: (config, ctx) => MyMiddleware(config),
);
```

__Using the context__ (the reason for this change) — e.g. an AI risk classifier that calls `generate`:

```dart
defineMiddleware<MyOptions>(
  name: 'risk_classifier',
  configSchema: MyOptions.$schema,
  create: (config, ctx) => RiskClassifierMiddleware(config, ai: ctx.ai),
);

class RiskClassifierMiddleware extends GenerateMiddleware {
  RiskClassifierMiddleware(this.config, {required this.ai});
  final MyOptions? config;
  final GenkitAI ai;

  @override
  Future<GenerateResponseHelper> generate(envelope, ctx, next) async {
    // Run a nested generate to classify the request before continuing.
    final verdict = await ai.generate(
      prompt: 'Classify the risk of this request: ${envelope.request}',
    );
    if (verdict.text.contains('BLOCK')) {
      throw GenkitException('Request blocked by risk classifier');
    }
    return next(envelope, ctx);
  }
}
```

The same `ctx.ai` also exposes the registry if you need to resolve actions by name:

```dart
final agent = await ctx.ai.registry.lookupAction('agent', 'researcher');
```

### DIFF
--- a/packages/genkit/lib/src/ai/generate_middleware.dart
+++ b/packages/genkit/lib/src/ai/generate_middleware.dart
@@ -15,7 +15,7 @@
 import 'package:schemantic/schemantic.dart';
 
 import '../core/action.dart';
-import '../core/registry.dart';
+import '../genkit_ai.dart';
 import '../types.dart';
 import 'generate_types.dart';
 import 'tool.dart';
@@ -84,10 +84,12 @@ abstract class GenerateMiddleware {
 
 /// Ambient dependencies handed to a middleware factory at instantiation time.
 ///
-/// The `registry` field is the active action registry, letting a middleware
-/// resolve other registered actions (models, tools, agents, etc.) by name when
-/// it is created.
-typedef GenerateMiddlewareContext = ({Registry registry});
+/// The `ai` field is an ephemeral [GenkitAI] instance backed by the active
+/// action registry. It lets a middleware run nested AI operations (e.g.
+/// [GenkitAI.generate], [GenkitAI.embed]) and resolve other registered actions
+/// (models, tools, agents, etc.) by name when it is created. The underlying
+/// registry is available via `ai.registry`.
+typedef GenerateMiddlewareContext = ({GenkitAI ai});
 
 abstract interface class GenerateMiddlewareDef<CustomOptions> {
   String get name;
--- a/packages/genkit/lib/src/ai/generate_middleware.dart
+++ b/packages/genkit/lib/src/ai/generate_middleware.dart
@@ -15,6 +15,7 @@
 import 'package:schemantic/schemantic.dart';
 
 import '../core/action.dart';
+import '../core/registry.dart';
 import '../types.dart';
 import 'generate_types.dart';
 import 'tool.dart';
@@ -81,12 +82,22 @@ abstract class GenerateMiddleware {
   }
 }
 
+/// Ambient dependencies handed to a middleware factory at instantiation time.
+///
+/// The `registry` field is the active action registry, letting a middleware
+/// resolve other registered actions (models, tools, agents, etc.) by name when
+/// it is created.
+typedef GenerateMiddlewareContext = ({Registry registry});
+
 abstract interface class GenerateMiddlewareDef<CustomOptions> {
   String get name;
   SchemanticType<CustomOptions>? get configSchema;
   Map<String, Object?>? get configJsonSchema;
 
-  GenerateMiddleware create([CustomOptions? config]);
+  GenerateMiddleware create(
+    CustomOptions? config,
+    GenerateMiddlewareContext ctx,
+  );
 }
 
 class _GenerateMiddlewareDef<CustomOptions>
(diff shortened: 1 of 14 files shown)


MichaelDoyle:
closed/re-opened just to tickle the GH actions

packages/genkit/lib/src/ai/generate_middleware.dart
```diff
@@ -81,12 +82,24 @@ abstract class GenerateMiddleware {
...
+/// action registry. It lets a middleware run nested AI operations (e.g.
+/// [GenkitAI.generate], [GenkitAI.embed]) and resolve other registered actions
+/// (models, tools, agents, etc.) by name when it is created. The underlying
+/// registry is available via `ai.registry`.
+typedef GenerateMiddlewareContext = ({GenkitAI ai});
+
 abstract interface class GenerateMiddlewareDef<CustomOptions> {
   String get name;
   SchemanticType<CustomOptions>? get configSchema;
   Map<String, Object?>? get configJsonSchema;
```
apascal07, line 98: This is the breaking bit: the `create` callback goes from `[config]` to `(config, ctx)`, so every existing `defineMiddleware` caller has to change. Worth a CHANGELOG note, and `docs/generate-middleware.md` still shows the old single-arg form, so that example won't compile as written anymore.

MichaelDoyle (approved):

apascal07 (approved):

packages/genkit/lib/src/ai/generate_middleware.dart
pavelgj: updated docs/generate-middleware.md.
changelog is autogenerated.

PR #322, opened 2026-07-09
feat(anthropic): curated known-model metadata for Claude models

cabljac:
## What

Adds a curated known-model catalog to the `genkit_anthropic` plugin, mirroring the treatment merged for `googleAI`/`vertexAI` (#320) and refactored to the enum shape in #323. Known Claude model names resolve with a typed `ModelInfo` (label + `stage: 'stable'` + capability `supports`); unknown names still resolve dynamically via the shared `commonModelInfo` fallback.

Curated models (verified against the Anthropic models docs):
- `claude-fable-5`, `claude-opus-4-8`, `claude-opus-4-7`, `claude-sonnet-5`, `claude-sonnet-4-6`, `claude-sonnet-4-5`, `claude-haiku-4-5`

## Changes

- **`lib/src/known_models.dart`** (new) — `KnownClaudeModel` enum: each value pairs a bare model `id` with its `label` and exposes the shared Claude capability preset via an `info` getter. `knownClaudeModels` is derived from the enum (`{ for m in KnownClaudeModel.values: m.id: m.info }`). Same shape as `KnownGeminiModel` in #323.
- **`lib/src/plugin_impl.dart`**
  - `knownModels` (`UnmodifiableMapView`) + `modelInfoFor()` fallback, wired into `resolve()` and model creation.
  - `list()` enriches discovered models with per-model `modelInfo`, appends curated entries missing from discovery (dedup), and falls back to the curated catalog when discovery fails.
  - Dated-snapshot ids (e.g. `claude-haiku-4-5-20251001`) are normalized to their alias for enrichment and dedup.
  - `print` replaced with `Logger('genkit_anthropic')`, matching the sibling plugins.
  - Optional `httpClient` (custom HTTP transport) seam, `@visibleForTesting`, used to inject a mock transport in the offline unit tests.
- **`test/known_models_test.dart`** (new) — resolve (curated + fallback), `list()` merge / no-dup / dated-id / discovery-failure, plus enum `info` shape, unmodifiable supports, and enum->map derivation with a literal-name assertion. Offline via a mocked HTTP client.
- **`pubspec.yaml`** — adds `logging` and `http` dependencies.

## Notes

- Go's Anthropic plugin has no per-Claude catalog (a single `defaultClaudeOpts`), so this deliberately goes a step beyond Go to bring Claude to the same autocompletable-catalog bar as the Gemini/Vertex work.
- `claude-mythos-5` / `claude-mythos-preview` are intentionally excluded (invitation-only, Project Glasswing).
- Depends conceptually on the enum pattern in #323; generate/stream paths are unchanged.

## Testing

`dart analyze` clean; tests pass (`dart test`). Verified end-to-end in the Genkit Dev UI: with no valid API key the discovery call fails and the plugin falls back to the curated catalog, registering all curated models with their labels + `stage: stable`.

### DIFF
--- a/packages/genkit_anthropic/lib/src/known_models.dart
+++ b/packages/genkit_anthropic/lib/src/known_models.dart
@@ -1,0 +1,69 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:genkit/plugin.dart';
+
+/// Claude models the Anthropic plugin curates capability metadata for.
+///
+/// Each value pairs a bare model [id] (no plugin prefix) with a display
+/// [label]; [info] builds the shared Claude capability preset. Other model
+/// names still resolve dynamically via the plugin's `commonModelInfo`
+/// fallback, so this enum only enriches the names listed here.
+enum KnownClaudeModel {
+  fable5('claude-fable-5', 'Claude Fable 5'),
+  opus48('claude-opus-4-8', 'Claude Opus 4.8'),
+  opus47('claude-opus-4-7', 'Claude Opus 4.7'),
+  sonnet5('claude-sonnet-5', 'Claude Sonnet 5'),
+  sonnet46('claude-sonnet-4-6', 'Claude Sonnet 4.6'),
+  sonnet45('claude-sonnet-4-5', 'Claude Sonnet 4.5'),
+  haiku45('claude-haiku-4-5', 'Claude Haiku 4.5');
+
+  const KnownClaudeModel(this.id, this.label);
+
+  /// Bare model name (no plugin prefix).
+  final String id;
+
+  /// Human-readable label surfaced in listings.
+  final String label;
+
+  /// The capability profile shared by every current Claude model: multiturn
+  /// chat, vision (media input), tool calling with tool choice, a system role,
+  /// and native constrained generation. Mirrors the single `defaultClaudeOpts`
+  /// profile the Go plugin applies to every Claude model.
+  ModelInfo get info => ModelInfo(
+    label: label,
+    // Unmodifiable: curated entries are shared across every resolution of the
+    // model, so accidental mutation through action metadata must fail loudly.
+    supports: Map.unmodifiable({
+      'multiturn': true,
+      'media': true,
+      'tools': true,
+      'toolChoice': true,
+      'systemRole': true,
+      'constrained': true,
+    }),
+    stage: 'stable',
+  );
+}
+
+/// Curated capability metadata for known Claude models, keyed by bare model
+/// name (no plugin prefix).
+///
+/// Derived from [KnownClaudeModel]; other model names still resolve dynamically
+/// with the shared `commonModelInfo` fallback. This map only enriches known
+/// names with a typed label and stable stage, and ensures they appear in
+/// listings even when the Anthropic models endpoint omits them.
+final knownClaudeModels = <String, ModelInfo>{
+  for (final model in KnownClaudeModel.values) model.id: model.info,
+};
--- a/packages/genkit_anthropic/lib/src/plugin_impl.dart
+++ b/packages/genkit_anthropic/lib/src/plugin_impl.dart
@@ -12,13 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'dart:collection';
 import 'dart:convert';
 
 import 'package:anthropic_sdk_dart/anthropic_sdk_dart.dart' as sdk;
 import 'package:genkit/plugin.dart';
+import 'package:http/http.dart' as http;
+import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:schemantic/schemantic.dart';
 
+import 'known_models.dart';
 import 'model.dart';
+
+final _logger = Logger('genkit_anthropic');
 
 /// Default model capabilities shared by all Anthropic Claude models.
 final commonModelInfo = ModelInfo(
@@ -37,6 +44,7 @@ final commonModelInfo = ModelInfo(
 ///
 /// Automatically discovers available models from the Anthropic API and
 /// registers them in the Genkit action registry.
+@visibleForTesting
 class AnthropicPluginImpl extends GenkitPlugin {
   /// The static API key used to authenticate requests.
   final String? apiKey;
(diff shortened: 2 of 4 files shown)


cabljac:
Addressed in `4e2b923`:

- **Add more Claude models** — added `claude-sonnet-4-6` and `claude-sonnet-4-5` to the curated catalog.
- **Extract `RegExp` to `static final`** — done (`_datedSuffixRegExp`).
- **Side-effect inside `Iterable.map()`** — done; `discovered` and `coveredAliases` are now computed as two separate passes, no side-effect in `map`.

Not applied:

- **Banner grouping comments in the test file** — skipped. No test file in the repo uses banner comments (including the sibling `genkit_vertexai/test/known_models_test.dart` this PR mirrors), so adding them here would diverge from the established convention. The existing `group(...)` blocks already provide the categorization.

packages/genkit_anthropic/lib/src/known_models.dart
```diff
@@ -0,0 +1,50 @@
...
+/// Any other model name still resolves dynamically with the shared
+/// `commonModelInfo` fallback; this map only enriches known names with a typed
+/// label and stable stage, and ensures they appear in listings even when the
+/// Anthropic models endpoint omits them.
+final knownClaudeModels = <String, ModelInfo>{
+  'claude-opus-4-8': claudeModelInfo('Claude Opus 4.8'),
+  'claude-opus-4-7': claudeModelInfo('Claude Opus 4.7'),
+  'claude-sonnet-5': claudeModelInfo('Claude Sonnet 5'),
+  'claude-haiku-4-5': claudeModelInfo('Claude Haiku 4.5'),
+};
```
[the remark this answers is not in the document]
cabljac: Added `claude-sonnet-4-6` and `claude-sonnet-4-5`, then confirmed `claude-fable-5` against the Anthropic models docs (GA on the Claude API) and added that too. Catalog now covers Fable 5, Opus 4.8/4.7, Sonnet 5/4.6/4.5, and Haiku 4.5. Skipped `claude-mythos-5`/`claude-mythos-preview` (invitation-only, Project Glasswing).

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -50,11 +52,41 @@ class AnthropicPluginImpl extends GenkitPlugin {
...
+  /// Names absent here still resolve; they fall back to [commonModelInfo].
+  final Map<String, ModelInfo> knownModels = UnmodifiableMapView(
+    knownClaudeModels,
+  );
+
+  /// Strips a trailing dated-snapshot suffix (e.g.
+  /// `claude-haiku-4-5-20251001` -> `claude-haiku-4-5`) so dated ids returned
+  /// by the models endpoint map onto the curated aliases.
+  static String _aliasOf(String modelName) =>
+      modelName.replaceFirst(RegExp(r'-\d{8}$'), '');
```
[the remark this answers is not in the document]
cabljac: Done, pulled it out to `static final _datedSuffixRegExp`.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -70,23 +102,41 @@ class AnthropicPluginImpl extends GenkitPlugin {
...
-          )
-          .toList();
+      // Track curated aliases already covered by discovery. The endpoint may
+      // return dated snapshot ids (e.g. `claude-haiku-4-5-20251001`), so match
+      // on the alias to both enrich and dedup against the curated catalog.
+      final coveredAliases = <String>{};
+      final discovered = response.data.map((m) {
+        coveredAliases.add(_aliasOf(m.id));
+        return _curatedMetadata(m.id, modelInfoFor(m.id));
+      }).toList();
```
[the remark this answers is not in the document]
cabljac: Good catch. Split into two passes so `map` stays pure: build `discovered`, then compute `coveredAliases` separately. It only worked before because of the immediate `.toList()`.

packages/genkit_anthropic/test/known_models_test.dart
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  /// Offline plugin — no request is issued when only [resolve] is exercised.
+  AnthropicPluginImpl plugin() =>
+      AnthropicPluginImpl(client: sdk.AnthropicClient());
+
+  Map<String, dynamic> modelInfoOf(Action action) =>
+      (action.metadata['model'] as Map).cast<String, dynamic>();
+
+  group('known model resolution', () {
```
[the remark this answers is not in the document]
cabljac: Leaving the banners off: no test file in the repo uses them, including the sibling `genkit_vertexai/test/known_models_test.dart` this mirrors, so adding them would diverge from the existing convention. The `group(...)` blocks already categorise.

packages/genkit_anthropic/test/known_models_test.dart
```diff
@@ -0,0 +1,178 @@
...
+      expect(info['supports'], commonModelInfo.supports);
+      expect(info.containsKey('stage'), isFalse);
+    });
+
+    test('non-model action types do not resolve', () {
+      expect(plugin().resolve('embedder', 'claude-opus-4-8'), isNull);
+    });
+  });
+
+  group('list', () {
```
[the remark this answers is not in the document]
cabljac: Same as the sibling group above: skipping the banner comments to stay consistent with the rest of the repo, which does not use them. The `group(...)` blocks already provide the categorisation.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -50,11 +55,43 @@ class AnthropicPluginImpl extends GenkitPlugin {
   sdk.AnthropicClient? _client;
 
   /// Creates an [AnthropicPluginImpl].
-  AnthropicPluginImpl({this.apiKey, this.headers, this.baseUrl});
+  ///
+  /// [client] injects a preconstructed Anthropic client, bypassing credential
```
cabljac: hmm not a fan of this... having a look

packages/genkit_anthropic/lib/src/known_models.dart
```diff
@@ -0,0 +1,53 @@
+// Copyright 2025 Google LLC
```
Lyokone: 2026?

packages/genkit_anthropic/lib/src/known_models.dart
```diff
@@ -0,0 +1,53 @@
...
+  supports: Map.unmodifiable({
+    'multiturn': true,
+    'media': true,
+    'tools': true,
+    'toolChoice': true,
+    'systemRole': true,
+    'constrained': true,
+  }),
+  stage: 'stable',
+);
```
Lyokone: Feels a bit weird to have that as a top level function. Would use a builder inside of the ModelInfo class.
Something like ModelInfo.fromLabel(String label) or something like that

packages/genkit_anthropic/lib/src/known_models.dart
```diff
@@ -0,0 +1,53 @@
...
+/// Anthropic models endpoint omits them.
+final knownClaudeModels = <String, ModelInfo>{
+  'claude-fable-5': claudeModelInfo('Claude Fable 5'),
+  'claude-opus-4-8': claudeModelInfo('Claude Opus 4.8'),
+  'claude-opus-4-7': claudeModelInfo('Claude Opus 4.7'),
+  'claude-sonnet-5': claudeModelInfo('Claude Sonnet 5'),
+  'claude-sonnet-4-6': claudeModelInfo('Claude Sonnet 4.6'),
+  'claude-sonnet-4-5': claudeModelInfo('Claude Sonnet 4.5'),
+  'claude-haiku-4-5': claudeModelInfo('Claude Haiku 4.5'),
+};
```
Lyokone, line 69: Should probably be an enum, would be cleaner and you can iterate on it more easily

packages/genkit_anthropic/test/known_models_test.dart
```diff
@@ -0,0 +1,173 @@
+// Copyright 2025 Google LLC
```
Lyokone: 2026

Lyokone (changes requested):

packages/genkit_anthropic/lib/src/known_models.dart
cabljac: Done in 3b5c559: the curated set is now a `KnownClaudeModel` enum (bare id + label) with the capability preset on an `info` getter, so the top-level builder is gone. Same shape applied to the Gemini catalogs in #323. Note: a literal `ModelInfo.fromLabel` would need a core change (ModelInfo is a generated type and Dart 3.10 extensions cannot add constructors), so the builder lives on the enum instead.

packages/genkit_anthropic/lib/src/known_models.dart
cabljac: Done in 3b5c559: `knownClaudeModels` is now an enum (`KnownClaudeModel`), derived into the map. Matches the `KnownGeminiModel` refactor in #323.

packages/genkit_anthropic/lib/src/known_models.dart
cabljac: Fixed to 2026 (year of authorship). Note the rest of the repo, including the #320 files added this week, still uses 2025, so these two new files are currently the only 2026 headers, a repo-wide sweep may be worth a follow-up.

packages/genkit_anthropic/test/known_models_test.dart
cabljac: Fixed to 2026, same as the source file above.

Lyokone (approved):

packages/genkit_anthropic/lib/src/plugin_impl.dart
cabljac: Resolved: the injected sdk.AnthropicClient param is gone; the seam is now an http.Client transport (httpClient), matching the googleai plugin in #329.

pavelgj (approved):

Issue #303, opened 2026-06-15
feat(dart/plugins): register P0 parity models (Gemini 3.x, embedders, Claude Opus)


cabljac:
Dart↔JS parity gap: register P0 model entries that JS already exposes but the Dart plugins do not. Purely additive — model definitions + embedder registration.

Related: #302 — Dart currently resolves models from raw strings with no curated known-model maps and no `EmbedderInfo` type. The registration pattern (and where embedder `input`/`dimensions` metadata lives) depends on that design decision, so embedder rows below may be blocked on #302.

## GoogleAI
- [ ] `gemini-embedding-2` (embedder) — #328 (metadata blocked on #327)
- [ ] `gemini-3-pro-image` — #326
- [ ] `gemini-3.1-flash-image` — #326
- [ ] `gemini-3.5-flash` — #326

## VertexAI — Embedders (landed in #261)
- [x] `text-embedding-005`
- [x] `text-multilingual-embedding-002`
- [x] `multimodalembedding@001`
- [x] `gemini-embedding-001`

## VertexAI — Gemini (landed in #320, refactored in #323)
- [x] `gemini-3.5-flash`
- [x] `gemini-3.1-flash-lite`
- [x] `gemini-3.1-flash-image`
- [x] `gemini-3-pro-image`

## Anthropic (pending #322 — approved, CI green)
- [ ] `claude-opus-4-7`
- [ ] `claude-opus-4-8`

Mirrors the go/py registrations (genkit-ai/genkit#5496 for py). Note #263 adds Claude via the VertexAI plugin; these rows are for the `genkit_anthropic` plugin itself.

## Notes
- P0 model-registration parity.
- #302 is closed; the model-side catalog landed via #320/#323 (`KnownGeminiModel`/`KnownClaudeModel`). The embedder-metadata half of that discussion is now tracked in #327.
- The `gemini-*-image` models need no special media `ModelInfo`: Go gives them the plain `Multimodal` profile (media out via `generateContent` inlineData); `Output:["media"]` is Imagen/Veo/TTS only.

cabljac:
Triage 2026-07-13: ticked the VertexAI rows (embedders landed in #261; Gemini 3.x in #320/#323). Remaining work split into sub-issues:

- #326 — wire the curated Gemini 3.x catalog into the googleai plugin (covers the three GoogleAI Gemini rows; the enum entries exist but only vertexai consumes them today)
[...]

PR #329, opened 2026-07-13
feat(google_genai): wire curated Gemini catalog into googleai plugin

### LINKED ISSUE #326: feat(googleai): wire curated Gemini 3.x catalog into the googleai plugin
Part of #303 (GoogleAI rows: gemini-3.5-flash, gemini-3-pro-image, gemini-3.1-flash-image).

## Context

#320/#323 added the curated `KnownGeminiModel` enum to `genkit_google_genai` with entries for all four Gemini 3.x models, but only the Vertex AI plugin consumes it (`vertex_api_client.dart` overrides `knownModels`). The googleai plugin itself never wires it up:

- `GoogleGenAiPluginImpl` does not override `CommonGoogleGenPlugin.knownModels` (default `const {}`), so `modelInfoFor()` always falls back to `commonModelInfo`.
- `list()` hardcodes `modelInfo: commonModelInfo`, does not append curated entries missing from discovery, and has no curated fallback when discovery fails.
- No typed model refs are exposed for IDE autocompletion (#302).

Note: the `-image` models need no special capability treatment. Go's `models.go` gives `gemini-3-pro-image` and `gemini-3.1-flash-image` the plain `Multimodal` profile (they are full chat models that emit media as inlineData through `generateContent`); the `Media`/`Output:["media"]` profile is for Imagen/Veo/TTS only. The existing `KnownGeminiModel.info` preset already matches.

## Scope

1. Override `knownModels => knownGeminiModels` in `GoogleGenAiPluginImpl` so `resolve()` picks up curated `ModelInfo` via the existing `modelInfoFor()`.
2. `list()`: enrich discovered models with `modelInfoFor(bareName)`, append curated entries missing from discovery (dedup), and fall back to the curated catalog when discovery fails. Mirror the implementation shape of #322.
3. Expose typed refs (`GeminiModels.gemini35Flash`, etc.) derived from `KnownGeminiModel.values` rather than re-typed strings.
4. Offline tests with a mocked HTTP client (same shape as the #322 test suite: curated resolve + fallback, list merge / no-dup / discovery-failure).
5. Live verification: image-bytes round-trip for the two `-image` models through `generateContent` -> `fromGeminiPart` inlineData.

Pattern already established by #320 (vertexai), #322 (anthropic), #323 (enum refactor).

cabljac:
Closes #326. Part of #303.

## What

Wires the curated `KnownGeminiModel` catalog (added in #320, refactored in #323) into the googleai plugin itself. Until now only the Vertex AI plugin consumed it: `GoogleGenAiPluginImpl` never overrode `knownModels`, and `list()` hardcoded `commonModelInfo` for every discovered model.

## Changes

- `knownModels` override in `google_api_client.dart`, so `resolve()` and `list()` pick up curated `ModelInfo` (label, stage, capability `supports`); unknown names still fall back to `commonModelInfo`. Curated entries missing from discovery are appended (deduped), and a discovery failure falls back to the curated catalog (warning-logged) instead of throwing; a missing API key still throws.
- `GoogleAiModels` typed refs, one per `KnownGeminiModel`, ids derived from the enum. Named plugin-scoped so vertexai can grow a symmetric surface. `googleAI.gemini('...')` remains the escape hatch.
- Optional `httpClient` constructor param (wrapped non-closing) so offline tests inject a mock transport, mirroring #322's seam.
- `test/known_models_wiring_test.dart`: offline coverage of curated resolve, fallback, list enrichment, dedupe, discovery-failure fallback (via a 500), typed-ref names.

## Testing

`dart analyze` clean; all 49 package tests pass (10 new). Live image-bytes round-trip for the image models: see PR comment.

Follow-ups (sibling-plugin seam convergence, TTS options duplication) tracked in #387.

### DIFF
--- a/packages/genkit_google_genai/lib/genkit_google_genai.dart
+++ b/packages/genkit_google_genai/lib/genkit_google_genai.dart
@@ -41,11 +41,11 @@ class GoogleGenAiPluginHandle {
   }
 }
 
-/// Typed [ModelRef]s for the curated Gemini models.
+/// Typed [ModelRef]s for the Gemini models curated by the googleai plugin.
 ///
 /// Each entry is equivalent to `googleAI.gemini('<name>')`, which remains the
 /// escape hatch for models not listed here.
-abstract final class GeminiModels {
+abstract final class GoogleAiModels {
   static final ModelRef<GeminiOptions> gemini35Flash = googleAI.gemini(
     KnownGeminiModel.gemini35Flash.id,
   );
--- a/packages/genkit_google_genai/lib/genkit_google_genai.dart
+++ b/packages/genkit_google_genai/lib/genkit_google_genai.dart
@@ -15,6 +15,7 @@
 import 'package:genkit/plugin.dart';
 
 import 'src/google_api_client.dart';
+import 'src/known_models.dart';
 import 'src/model.dart';
 
 export 'src/model.dart';
@@ -38,4 +39,26 @@ class GoogleGenAiPluginHandle {
       customOptions: TextEmbedderOptions.$schema,
     );
   }
+}
+
+/// Typed [ModelRef]s for the curated Gemini models.
+///
+/// Each entry is equivalent to `googleAI.gemini('<name>')`, which remains the
+/// escape hatch for models not listed here.
+abstract final class GeminiModels {
+  static final ModelRef<GeminiOptions> gemini35Flash = googleAI.gemini(
+    KnownGeminiModel.gemini35Flash.id,
+  );
+
+  static final ModelRef<GeminiOptions> gemini31FlashLite = googleAI.gemini(
+    KnownGeminiModel.gemini31FlashLite.id,
+  );
+
+  static final ModelRef<GeminiOptions> gemini31FlashImage = googleAI.gemini(
+    KnownGeminiModel.gemini31FlashImage.id,
+  );
+
+  static final ModelRef<GeminiOptions> gemini3ProImage = googleAI.gemini(
+    KnownGeminiModel.gemini3ProImage.id,
+  );
 }
--- a/packages/genkit_google_genai/lib/src/google_api_client.dart
+++ b/packages/genkit_google_genai/lib/src/google_api_client.dart
@@ -13,21 +13,32 @@
 // limitations under the License.
 
 import 'package:genkit/plugin.dart';
+import 'package:http/http.dart' as http;
 import 'package:meta/meta.dart';
 
 import 'api_client.dart';
 import 'common_plugin.dart';
 import 'generated/generativelanguage.dart' as gcl;
+import 'known_models.dart';
 import 'model.dart';
 
 @visibleForTesting
 class GoogleGenAiPluginImpl extends CommonGoogleGenPlugin {
   String? apiKey;
 
-  GoogleGenAiPluginImpl({this.apiKey});
+  /// Test-only HTTP transport. When set it replaces the API-key client for
+  /// every request (any per-request `apiKey` option is ignored) and is never
+  /// closed by the plugin; the caller owns its lifecycle.
+  @visibleForTesting
+  final http.Client? httpClient;
+
+  GoogleGenAiPluginImpl({this.apiKey, this.httpClient});
 
   @override
   String get name => 'googleai';
+
+  @override
+  final Map<String, ModelInfo> knownModels = knownGeminiModels;
 
   @override
   Future<GenerativeLanguageBaseClient> getApiClient([
@@ -35,7 +46,7 @@ class GoogleGenAiPluginImpl extends CommonGoogleGenPlugin {
   ]) async {
     return GenerativeLanguageBaseClient(
       baseUrl: 'https://generativelanguage.googleapis.com/',
-      client: httpClientFromApiKey(requestApiKey ?? apiKey),
+      client: httpClient ?? httpClientFromApiKey(requestApiKey ?? apiKey),
     );
   }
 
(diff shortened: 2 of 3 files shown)


packages/genkit_google_genai/lib/src/google_api_client.dart
```diff
@@ -13,29 +13,40 @@
...
   @override
   Future<GenerativeLanguageBaseClient> getApiClient([
     String? requestApiKey,
   ]) async {
     return GenerativeLanguageBaseClient(
       baseUrl: 'https://generativelanguage.googleapis.com/',
-      client: httpClientFromApiKey(requestApiKey ?? apiKey),
+      client: httpClient ?? httpClientFromApiKey(requestApiKey ?? apiKey),
     );
   }
```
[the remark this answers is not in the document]
cabljac: Done in 2aa0d5a - adopted the non-closing wrapper. Agreed it's the right shape: the inherited createModel path also closes unconditionally, so the null-gates didn't cover generate calls. One correction to the severity: nothing fails today since BaseClient.close() is a no-op unless overridden, but it was a real trap for future generate-path tests.

packages/genkit_google_genai/lib/src/google_api_client.dart
```diff
@@ -78,16 +98,30 @@ class GoogleGenAiPluginImpl extends CommonGoogleGenPlugin {
...
     } catch (e, stack) {
       if (e is GenkitException) rethrow;
       logger.warning('Failed to list models: $e', e, stack);
       throw handleException(e, stack);
     } finally {
-      service.client.close();
+      if (httpClient == null) {
+        service.client.close();
+      }
     }
```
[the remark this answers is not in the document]
cabljac: Done in 2aa0d5a.

packages/genkit_google_genai/lib/src/google_api_client.dart
```diff
@@ -145,7 +179,9 @@ class GoogleGenAiPluginImpl extends CommonGoogleGenPlugin {
         } catch (e, stack) {
           throw handleException(e, stack);
         } finally {
-          service.client.close();
+          if (httpClient == null) {
+            service.client.close();
+          }
         }
```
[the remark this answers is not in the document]
cabljac: Done in 2aa0d5a.

packages/genkit_google_genai/lib/src/google_api_client.dart
```diff
@@ -151,3 +188,15 @@ class GoogleGenAiPluginImpl extends CommonGoogleGenPlugin {
...
+/// per-call cleanup never tears down an injected transport.
+class _NonClosingClient extends http.BaseClient {
+  _NonClosingClient(this._inner);
+
+  final http.Client _inner;
+
+  @override
+  Future<http.StreamedResponse> send(http.BaseRequest request) =>
+      _inner.send(request);
+}
```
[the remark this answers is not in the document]
cabljac: Done in 492018c - explicit no-op close() with a comment explaining the caller-owned client.

packages/genkit_google_genai/lib/src/google_api_client.dart
```diff
@@ -13,29 +13,41 @@
...
 
 @visibleForTesting
 class GoogleGenAiPluginImpl extends CommonGoogleGenPlugin {
   String? apiKey;
 
-  GoogleGenAiPluginImpl({this.apiKey});
+  /// Test-only HTTP transport. When set it replaces the API-key client for
+  /// every request (any per-request `apiKey` option is ignored) and is never
+  /// closed by the plugin; the caller owns its lifecycle.
+  final http.Client? httpClient;
```
Lyokone, line 32: If it's really only for test, we should use https://api.flutter.dev/flutter/meta/visibleForTesting-constant.html

packages/genkit_google_genai/lib/src/google_api_client.dart
```diff
@@ -151,3 +187,21 @@ class GoogleGenAiPluginImpl extends CommonGoogleGenPlugin {
...
+
+  @override
+  Future<http.StreamedResponse> send(http.BaseRequest request) =>
+      _inner.send(request);
+
+  @override
+  void close() {
+    // Intentional no-op: the inner client is caller-owned and must survive
+    // the per-call close() in the list/embed/generate paths.
+  }
```
Lyokone, line 206: Is this intentional? Feel a bit wrong, I was assuming it was for testing but not seeing it called in test files?

packages/genkit_google_genai/lib/genkit_google_genai.dart
```diff
@@ -39,3 +40,25 @@ class GoogleGenAiPluginHandle {
     );
   }
 }
+
+/// Typed [ModelRef]s for the Gemini models curated by the googleai plugin.
+///
+/// Each entry is equivalent to `googleAI.gemini('<name>')`, which remains the
+/// escape hatch for models not listed here.
+abstract final class GoogleAiModels {
```
Lyokone, line 48: Probably could be an enum?

packages/genkit_google_genai/lib/src/google_api_client.dart
cabljac: To double check the reasoning: the inherited `createModel` in `common_plugin.dart` calls `service.client.close()` unconditionally in its `finally`, so without the wrapper the first generate call would close the caller-owned client. It's not called directly in test files, but every `list()` test injects a `MockHttpClient`, and each call ends in `service.client.close()` which goes through the no-op.

packages/genkit_google_genai/lib/src/google_api_client.dart
cabljac: Class already has this, so no need to add here, resolving :)

packages/genkit_google_genai/lib/genkit_google_genai.dart
cabljac: I think a plain enum can't hold the refs, since `modelRef()` is called at runtime. An enhanced enum implementing `ModelRef` would work, but `ModelRef.name` returns the namespaced action id ('googleai/gemini-3.5-flash') while `Enum.name` returns the identifier, so the enum would have to shadow its own `name` with a different meaning. Kept them as static finals on an abstract final class instead, with ids derived from the `KnownGeminiModel` enum so there's a single source of truth.

cabljac:
Live image-bytes round-trip done against branch head (9b12d0c), closing out the last scope item of #326:

```
gemini-3-pro-image: OK contentType=image/jpeg dataUri=true bytes=92392 magic=ff d8 ff
gemini-3.1-flash-image: OK contentType=image/jpeg dataUri=true bytes=401195 magic=ff d8 ff
```

Both return inlineData through `fromGeminiPart` as data-URI media parts with valid JPEG bytes (ff d8 ff).

pavelgj (approved):
LGTM

PR #330, opened 2026-07-13
feat(genkit): type-safe agent State with schemantic parsing + agent API polish

pavelgj:
Makes the agent client surface generic over a `State` type and adds first-class support for parsing custom session state into real domain types via `SchemanticType<State>`. Also lands a round of API polish on the agent core to align with the JS reference and improve ergonomics/type-safety.

Previously, custom agent state was only ever a raw JSON view (`Object?` / `Map`). Reading it as a domain type required a manual `MyType.fromJson(chat.state!)` at every call site, and `Agent`/`AgentApi`/`AgentChat` were non-generic. Now:

```dart
// Server: pass a stateSchema -> Agent<MyState>
final agent = ai.defineCustomAgent(
  name: 'planner',
  stateSchema: MyState.schema,
  fn: ...,
);
final chat = agent.chat();
final res = await chat.sendText('hi');
res.state;        // MyState (parsed), not a raw Map
chat.state?.count // typed field access

// Client: same, over HTTP
final remote = remoteAgent(url: '...', stateSchema: MyState.schema);
```

When no `stateSchema` is supplied, `State` defaults to `Object?` and behaves exactly as before (a bare view-cast over the JSON), preserving JS parity and backward compatibility.

### Generic, schema-aware `State`

- `AgentApi<State>` gains an optional `SchemanticType<State>? stateSchema`; forwarded to every `AgentChat` it creates.
- `AgentChat<State>`, `AgentResponse<State>`, `AgentChunk<State>`, and `AgentError<State>` now parse custom state at every read boundary (`chat.state`, `res.state` incl. the server-managed `customPatch` fallback, `chunk.custom`, and `AgentError.state`) via a shared `_castOrParseState` helper (mirrors `Action`'s `inputSchema != null ? parse : as`). Internal storage stays raw JSON so streamed patches still apply.
- `Agent<State>`, `defineCustomAgent<State>`, `definePromptAgent<State>`, and `Genkit.defineAgent<..., State>` are now generic and thread `stateSchema` through; `remoteAgent<State>` accepts `stateSchema`.
- A mismatched `State` (or no schema over a non-matching type) still throws at read time - Dart generics are reified - which is documented and tested.

### Agent API polish

- `agentInputFromText` helper replaced by ergonomic `sendText` / `sendTextStream` / `detachText` convenience methods.
- `abort()` returns a typed `SnapshotStatus?` (was `String?`); `stateManagement` typed as the `AgentStateManagement` enum (was `String`).
- Public `AgentApi` constructor replaces the `createAgentApi` factory; `dynamic` public members tightened to `Object?`; detached-run heartbeat interval typed as `Duration`; asserts added to `chat`/`loadChat`.

### Tests & consumers

- New `agent_core_test.dart` groups: `generic State typing (wire -> client)` (default view-cast path, incl. mismatch-throws) and `schema-aware State typing (SchemanticType<State>)` (parsed wire state, server-managed fallback, `AgentChunk.custom`, `loadChat`, and parsed last-good state on `AgentError`).
- Updated `agent_test.dart`, `remote_agent_test.dart`, `agent_middleware_test.dart`, the `testapps/agents` sample (server, web pages, remote client), and exports.


huangjeff5:
Looks good. Small nit: should DetachedTask be generic too?

pavelgj:
> Looks good. Small nit: should DetachedTask be generic too?

possibly, but the way dart type system works it would not provide much value in type safety... we cannot easily strongly type SessionSnapshot as it's a generated schemantic type...

pavelgj:
> > Looks good. Small nit: should DetachedTask be generic too?
> 
> possibly, but the way dart type system works it would not provide much value in type safety... we cannot easily strongly type SessionSnapshot as it's a generated schemantic type...

Done.

huangjeff5 (approved):
Overall architecture looks solid. `AgentSnapshot<State>` cleanly avoids modifying generated wire schemas and keeps `json_patch.dart` operating on plain JSON (`_clientState!.custom`).

One note on mid-stream `customPatch` handling (`agent_core.dart:L924`):

When `_stateSchema` is provided and the model streams state updates incrementally across chunks, `schema.parse(_clientState!.custom)` throws on intermediate frames where required properties are incomplete, terminating `Stream<AgentChunk<State>>` mid-turn.

IMO this is what should happen:
Gate mid-stream chunk parsing inside `buildStream` (`AgentChat._sendStream`) with a `try/catch`:
- Unconditionally apply `_applyCustomPatch(patch)` to `_clientState!.custom` so raw JSON accumulates continuously across frames.
- If `schema.parse` throws on an intermediate frame, yield `null` on `chunk.custom` (or hold the last-known valid `State`) until `_clientState!.custom` conforms to `_stateSchema`.
- Validate strictly when resolving the terminal `turn.response`.

In other words, don't throw on chunks that provide a partial state; just update the raw JSON and save the parsed object when it actually conforms to the expected schema. That way we support cases where intermediate state is partial.

pavelgj:
> Overall architecture looks solid. `AgentSnapshot<State>` cleanly avoids modifying generated wire schemas and keeps `json_patch.dart` operating on plain JSON (`_clientState!.custom`).
> 
> One note on mid-stream `customPatch` handling (`agent_core.dart:L924`):
> 
> When `_stateSchema` is provided and the model streams state updates incrementally across chunks, `schema.parse(_clientState!.custom)` throws on intermediate frames where required properties are incomplete, terminating `Stream<AgentChunk<State>>` mid-turn.
> 
> IMO this is what should happen: Gate mid-stream chunk parsing inside `buildStream` (`AgentChat._sendStream`) with a `try/catch`:
> 
> * Unconditionally apply `_applyCustomPatch(patch)` to `_clientState!.custom` so raw JSON accumulates continuously across frames.
> * If `schema.parse` throws on an intermediate frame, yield `null` on `chunk.custom` (or hold the last-known valid `State`) until `_clientState!.custom` conforms to `_stateSchema`.
> * Validate strictly when resolving the terminal `turn.response`.
> 
> In other words, don't throw on chunks that provide a partial state; just update the raw JSON and save the parsed object when it actually conforms to the expected schema. That way we support cases where intermediate state is partial.

done

PR #332, opened 2026-07-15
feat(schemantic): add serialize to convert typed values to JSON

pavelgj:
Add `SchemanticType.serialize` as the inverse of `parse`, converting a typed value back into plain JSON. The default implementation handles scalars, lists, maps, and any object exposing a `toJson()` method recursively. Add an optional `serialize` callback to `SchemanticType.from` for custom encodings when bridging your own types.

### DIFF
--- a/packages/schemantic/lib/schemantic.dart
+++ b/packages/schemantic/lib/schemantic.dart
@@ -149,15 +149,41 @@ abstract class SchemanticType<T> {
   const SchemanticType();
 
   /// Creates a schema from a JSON schema and a parse function.
+  ///
+  /// Optionally provide [serialize] to convert a [T] back into plain JSON; when
+  /// omitted, the default [SchemanticType.serialize] is used (which handles
+  /// scalars, lists, maps, and any object exposing a `toJson()` method).
   static SchemanticType<T> from<T>({
     required Map<String, Object?> jsonSchema,
     required T Function(dynamic json) parse,
-  }) => _AdHocSchema(jsonSchema, parse);
+    Object? Function(T value)? serialize,
+  }) => _AdHocSchema(jsonSchema, parse, serialize);
 
   /// Parses the given [json] object into type [T].
   ///
   /// Throws if the JSON data does not match the expected structure.
   T parse(Object? json);
+
+  /// Serializes a [value] of type [T] back into plain JSON (a `Map`, `List`,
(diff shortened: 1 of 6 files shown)


packages/schemantic_builder/test/schema_generator_test.dart
```diff
@@ -288,6 +288,13 @@ Future<void> _testBuilderWithNoFail(
     outputs: outputs,
     onLog: (log) {
       if (log.level >= Level.WARNING) {
+        // Ignore the environmental warning emitted when the running SDK's
```
pavelgj, line 291: this is to fix a CI failure on dart dev version....

packages/schemantic/lib/schemantic.dart
```diff
@@ -149,16 +149,51 @@ abstract class SchemanticType<T> {
...
+  /// `String`, `num`, `bool`, or `null`) - the inverse of [parse].
+  ///
+  /// The default implementation handles scalars, lists, and maps recursively,
+  /// and delegates to `toJson()` for any other object (all generated schemantic
+  /// types expose one). Override this for custom encodings.
+  Object? serialize(T value) => _serializeToJson(value);
+
+  /// Recursively converts a JSON-shaped [value] (scalar, `List`, `Map`, or an
+  /// object with a `toJson()` method) into plain JSON.
+  static Object? _serializeToJson(Object? value) {
```
cabljac, line 177: Looks like the result of `toJson()` is returned as-is, without recursing into it. A hand-rolled `toJson()` returning e.g. `{'child': someDomainObject}` would leak non-JSON values despite the doc promising plain JSON. Intentional trade-off, or should this be `_serializeToJson((value as dynamic).toJson())`?

cabljac (approved):
One comment, no other blockers

packages/schemantic/lib/schemantic.dart
pavelgj: Good catch, wasn't intentional. Fixed it to recurse into the toJson() result so nested domain objects get normalized to plain JSON. Kept the NoSuchMethodError catch scoped to just the toJson() lookup so an error thrown inside a valid nested toJson() doesn't get mislabeled. Added tests for both the nested-normalization and nested-non-serializable cases. (One edge case: a self-referential toJson() would now stack-overflow instead of silently leaking, but that's already a broken toJson().)

PR #342, opened 2026-07-23
feat(a2ui): add A2UI streaming UI protocol package and sample application

pavelgj:
Adds the [A2UI (Agent-to-UI)](https://a2ui.org/) streaming UI protocol and middleware support to Genkit Dart (`packages/genkit_a2ui`) along with a full sample application (`testapps/a2ui`).

An A2UI-enabled agent streams not just prose, but rich, interactive UI surfaces carrying complete design system catalogs (components, layout containers, forms) that a client application (such as Flutter or Web) can render incrementally.

### What's Included

* **`packages/genkit_a2ui` (Protocol & Middleware)**:
  * **`a2ui()` Middleware**: Appends catalog capabilities/guidelines into the system prompt, intercepts model output (both streamed chunks and final messages), extracts `a2ui` codeblocks, validates components, and transforms them into standard `data` parts (`application/a2ui+json`).
  * **`basicCatalog`**: Bundled basic UI catalog (v0.9) containing standard UI components (Text, Columns, Rows, Buttons, TextField, CheckBox, Slider, etc.).
  * **`A2uiStreamParser`**: Incremental regex codeblock extractor with customizable validation modes (`strict`, `warn`, `off`).
  * **Platform Isolation**: Server-only loaders (`dart:io`) are fully isolated from browser/Flutter-safe clients.
  * **Robust Test Suite**: 62 unit tests verifying parser stream handling, validation, auto-synthesis of `createSurface`, and client serialization.

* **`testapps/a2ui` (Runnable Sample App)**:
  * A Shelf-based HTTP server hosting an A2UI-enabled agent.
  * A Flutter web/desktop chat application using `package:genui` to render stream-driven surfaces in real time.

---

#### 1. Server Setup: Define an A2UI-Enabled Agent
```dart
import 'package:genkit_a2ui/a2ui.dart';
import 'package:genkit_google_genai/genkit_google_genai.dart';

final ai = Genkit(plugins: [googleAI(), A2uiPlugin()]);

final uiAgent = ai.defineAgent(
  name: 'uiAgent',
  model: googleAI.gemini('gemini-flash-latest'),
  system: 'You are a helpful assistant. Render UI when a result is clearer shown than told.',
  use: [a2ui()], // Active basic catalog, automated prompt injection, and stream interception
);
```

#### 2. Client Setup: Read Streamed Surfaces (Flutter / Web Safe)
```dart
import 'package:genkit/client.dart';
import 'package:genkit_a2ui/client.dart';

final agent = remoteAgent(url: 'http://localhost:8080/api/uiAgent');
final chat = agent.chat();
final turn = chat.sendStream(text: 'Weather in London');

await for (final chunk in turn.stream) {
  // Extract A2UI envelopes from the chunk and feed them to your renderer (e.g. genui)
  final envelopes = a2uiEnvelopesFromParts(chunk.modelChunk.content);
  if (envelopes.isNotEmpty) {
    surfaceController.handleMessages(envelopes);
  }
}
```

#### 3. Client Interaction: Send Actions back to the Server
```dart
import 'package:genkit_a2ui/client.dart';

// Wrap user interactions (button press, form submits) and send them as the next turn
final message = actionToMessage(
  A2uiClientAction(
    name: 'refreshWeather',
    surfaceId: 'surface-1',
    sourceComponentId: 'refreshBtn',
    timestamp: DateTime.now().toUtc().toIso8601String(),
    context: {'city': 'London'},
  ),
);

final response = chat.sendStream(message: message);
```


packages/genkit_a2ui/lib/src/a2ui_middleware.dart
```diff
@@ -0,0 +1,437 @@
...
+        },
+        context: ctx.context,
+        inputStream: ctx.inputStream,
+        init: null,
+      );
+    }
+
+    // 3) Run downstream model, then transform the final message. The final
+    //    parse replays the same surface ids the stream minted.
+    final response = await next(newRequest, wrappedCtx);
```
apascal07, line 262: The stream parser never gets flushed, so whatever is left in its buffer when the model call ends is dropped. `push()` always holds back up to 8 trailing chars in case they're a partial fence, so the tail of the streamed prose never reaches the client, and a reply shorter than that streams as nothing at all. The final message is fine because it uses its own parser and does flush, which is why the tests miss this: `sampleText` ends on the closing fence. The sample app renders prose straight off the stream, so it shows up there.

packages/genkit_a2ui/lib/src/parser.dart
```diff
@@ -0,0 +1,351 @@
...
+}
+
+/// Opening fence, matched case-insensitively (```a2ui).
+final _openFenceRe = RegExp(r'```[ \t]*a2ui[ \t]*\r?\n', caseSensitive: false);
+
+/// The longest prefix of an opening fence, used to hold back a partial fence.
+const _maxPartialFence = 8; // '```a2ui\n'.length
+
+/// Closing fence: ``` anywhere.
+final _closeFenceRe = RegExp('```');
```
apascal07: This matches any triple backtick, including one inside a JSON string value. The `Text` component's description invites inline Markdown, so a component carrying a fenced snippet truncates the block mid-JSON and the whole surface gets dropped, silently under the default `warn`.

packages/genkit_a2ui/lib/src/parser.dart
```diff
@@ -0,0 +1,351 @@
...
+
+  /// Pass envelopes through unchecked.
+  off,
+}
+
+/// Opening fence, matched case-insensitively (```a2ui).
+final _openFenceRe = RegExp(r'```[ \t]*a2ui[ \t]*\r?\n', caseSensitive: false);
+
+/// The longest prefix of an opening fence, used to hold back a partial fence.
+const _maxPartialFence = 8; // '```a2ui\n'.length
```
apascal07, line 50: The holdback is fixed at 8, but the open fence regex allows whitespace around the tag plus CRLF, so a real fence can be longer. If a chunk boundary leaves 9+ chars of incomplete fence buffered, the slice leaks a backtick into prose and the remainder can never match again, so the whole JSON block streams to the user as raw text.

packages/genkit_a2ui/lib/src/parser.dart
```diff
@@ -0,0 +1,351 @@
...
+      if (normalized != null) out.add(normalized);
+    }
+    if (out.isEmpty) return null;
+
+    // Guarantee the block opens with a `createSurface`, so the client always
+    // has a surface before any update targets it. Models often emit only
+    // `updateComponents`/`updateDataModel` on a follow-up (e.g. a "refresh")
+    // turn; without this the renderer would drop those updates as "surface not
+    // found". Idempotent re-creation is fine - it resets the surface.
+    final hasCreate = out.any((e) => e['createSurface'] != null);
```
apascal07, line 273: Invalid envelopes are dropped before this check, so a block whose `updateComponents` failed validation but whose `updateDataModel` survived comes out as `[createSurface, updateDataModel]`: the client resets the surface, then applies data to nothing. A blank surface replacing a good one is worse than emitting nothing, so it may be better to skip synthesis when something in the block was rejected.

packages/genkit_a2ui/lib/src/parser.dart
```diff
@@ -0,0 +1,351 @@
...
+  String? _validateComponents(Object? components) {
+    final catalog = this.catalog;
+    if (catalog == null) return null;
+    if (components is! List) {
+      return 'updateComponents.components must be an array.';
+    }
+    final known = catalog.components.map((c) => c.name).toSet();
+    final hasRoot = components.any((c) => c is Map && c['id'] == 'root');
+    if (!hasRoot) {
+      return 'component list must contain a component id "root".';
```
apascal07: Requiring `root` in every `updateComponents` matches the injected prompt, which says to re-render the whole surface. But it's silently wrong for anyone using `instructions: 'none'` with their own prompt, and under the default `warn` a legitimate incremental update just disappears into a log line. Worth documenting the coupling on `validate` at least.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
```diff
@@ -0,0 +1,437 @@
...
+      lines.add('[UI surface $surfaceId created]');
+    } else if (env['updateComponents'] != null ||
+        env['updateDataModel'] != null ||
+        env['deleteSurface'] != null) {
+      // Prior assistant surface content - summarize as a rendered surface.
+      lines.add('[rendered UI surface]');
+    }
+  }
+  // Collapse repeated "[rendered UI surface]" lines from one assistant turn.
+  return {...lines}.join(' ');
```
apascal07: Comment says this collapses repeated `[rendered UI surface]` lines, but the set dedupes action lines too, so two identical actions in one message become one.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
```diff
@@ -0,0 +1,437 @@
...
+  ModelRequest _sanitizeInboundA2ui(ModelRequest req) {
+    var changed = false;
+    final messages = req.messages.map((message) {
+      var msgChanged = false;
+      final content = <Part>[];
+      for (final part in message.content) {
+        if (isA2uiPart(part)) {
+          msgChanged = true;
+          final text = _summarizeA2uiPart(a2uiEnvelopesFromParts([part]));
+          if (text.isNotEmpty) content.add(TextPart(text: text));
```
apascal07, line 393: If the summary comes back empty, which happens for an envelope shape this doesn't recognize, and that was the message's only part, `content` ends up empty. Some providers reject that.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
```diff
@@ -0,0 +1,437 @@
...
+    // 3) Run downstream model, then transform the final message. The final
+    //    parse replays the same surface ids the stream minted.
+    final response = await next(newRequest, wrappedCtx);
+    surfaceIds.reset();
+    return _transformResponse(response, catalog, surfaceIds.replayNext);
+  }
+
+  /// Appends A2UI instructions to (or creates) the system message.
+  ModelRequest _injectInstructions(ModelRequest req, A2uiCatalog catalog) {
+    final text = renderCatalogInstructions(catalog);
```
apascal07: Rebuilt on every model call, so more than once per turn in a tool loop, including the 58-name icon list. It's a pure function of the catalog, so it could be cached.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
```diff
@@ -0,0 +1,436 @@
...
+        // (prose before/after a block) is preserved in the aggregated message.
+        final pushed = parser.push(part.text ?? '');
+        final flushed = parser.flush();
+        newContent.addAll(
+          _partsFromSegments([...pushed.segments, ...flushed.segments]),
+        );
+      } else {
+        newContent.add(part);
+      }
+    }
```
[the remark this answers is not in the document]
pavelgj: Done. The final-message transform pushes every text part through the parser first and flushes once at the end, so a block spanning multiple text parts stays intact.

packages/genkit_a2ui/lib/src/catalog.dart
```diff
@@ -0,0 +1,404 @@
...
+    return A2uiCatalog(
+      id: (json['id'] as String?) ?? '',
+      components: rawComponents
+          .map(
+            (c) => A2uiCatalogComponent.fromJson(
+              (c as Map).cast<String, dynamic>(),
+            ),
+          )
+          .toList(),
+    );
```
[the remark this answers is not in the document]
pavelgj: Done. `A2uiCatalog.fromJson` now skips non-map entries defensively instead of an unchecked cast.

packages/genkit_a2ui/lib/src/parser.dart
```diff
@@ -0,0 +1,349 @@
...
+    }
+    for (final c in components) {
+      if (c is! Map || c['component'] is! String) {
+        return 'every component needs a "component" type name.';
+      }
```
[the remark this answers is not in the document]
pavelgj: Done. `_validateComponents` now requires `c['id'] is String`.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
pavelgj: Good catch. The middleware now flushes the stream parser after the model call and emits the drained segments as a final chunk, so the withheld prose tail and any unterminated trailing block reach stream-only consumers. Added two tests: a sub-holdback prose reply streamed one char at a time (the deltas must reconstruct the full text), and an unterminated block that must still surface after flush.

packages/genkit_a2ui/lib/src/parser.dart
pavelgj: Fixed. The closing-fence regex is now anchored to line start, `(^|\n)[ \t]*` followed by the backticks, mirroring the opening fence, so a fenced snippet inside a `Text` markdown value no longer truncates the block. Added a test with a `Text` value containing an inline ```` ```npm test``` ````.

packages/genkit_a2ui/lib/src/parser.dart
pavelgj: You're right that a fence with internal whitespace/CRLF can exceed 8 chars and leak a backtick mid-stream. I've left the constant as-is for now to stay aligned with the JS `MAX_PARTIAL_FENCE`, and the end-of-turn flush at least guarantees nothing is permanently lost. If you'd prefer, I can make the holdback derive from the trailing partial-fence candidate (hold back from the last ``` run when the tail could still complete an open fence) rather than a fixed length. Happy to do that as a follow-up so JS and Dart move together.

packages/genkit_a2ui/lib/src/parser.dart
pavelgj: Restructured `_finalizeBlock`. It now distinguishes a fresh render (the model used the `SURFACE_ID` placeholder or omitted the id, so normalization swapped in the freshly-minted id) from a genuine incremental update to an explicit, pre-existing surface. Incremental updates no longer get a synthesized `createSurface`, so we can't reset a good surface, which was the bad outcome you flagged. For a fresh render there's no prior surface to lose.

The narrower case (an `updateComponents` dropped by validation while a sibling `updateDataModel` on the placeholder survives) still yields `[createSurface, updateDataModel]` against a brand-new empty surface, but nothing existing is clobbered. If you'd rather also skip synthesis whenever any envelope in the block was rejected, I can track a "rejected" flag and bail. Added tests for the incremental and rootless-incremental cases.

packages/genkit_a2ui/lib/src/parser.dart
pavelgj: Moved the root rule out of the per-component catalog check into a batch-level `_validateRoot` that only runs for full-surface renders. Incremental updates to an explicit existing surface may now patch a subtree without re-declaring `root`, even under `strict`, so `instructions: 'none'` users with their own prompt aren't penalized. It also runs regardless of whether a catalog is configured, since it's a protocol rule, not a catalog check. Added a test.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
pavelgj: Fixed. The summary now only collapses consecutive `[rendered UI surface]` sentinels; distinct action lines, and even two identical actions, are preserved. Added a test with two identical actions plus a distinct one.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
pavelgj: Fixed. If sanitizing an a2ui-only message summarizes to nothing (an envelope shape we don't recognize), we keep the message non-empty with a neutral `[UI interaction]` marker so providers that reject empty content stay happy. Added a test with an unrecognized envelope.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
pavelgj: Fixed. `renderCatalogInstructions` is memoized per resolved catalog on the middleware instance, so the tool loop no longer rebuilds the 58-name icon prompt on each raw model call.

apascal07 (approved):

packages/genkit_a2ui/lib/src/parser.dart
cabljac: ah didn't see this when i left the comment on the JS version

packages/genkit_a2ui/lib/src/parser.dart
```diff
@@ -0,0 +1,424 @@
...
+    void swapSurfaceId(Object? payload) {
+      if (payload is! Map) return;
+      final p = payload;
+      final current = p['surfaceId'];
+      if (current == null || current == surfaceIdPlaceholder || current == '') {
+        p['surfaceId'] = surface;
+      }
+    }
+
+    if (e['createSurface'] != null) {
```
cabljac, line 337: nit: this will drop other top level keys, i think you made a fix for this in JS? (the "open-ended" clause of the spec where other fields might be added in later)

I think no problem with current 0.9 spec of A2UI but just for parity across SDKs prob worth a fix

testapps/a2ui/lib/main.dart
```diff
@@ -0,0 +1,391 @@
...
+    });
+    _scrollToBottom();
+    await _runTurn(_chat.sendStream(text: text));
+  }
+
+  /// Sends a rendered surface's action back to the agent as the next turn.
+  Future<void> _onSurfaceSubmit(dynamic message) async {
+    final action = _actionFromSubmit(message);
+    if (action == null) return;
+    setState(() {
```
cabljac, line 173: Should we add a `_busy` guard here as well? If someone presses a button mid-stream, i think we'd get two concurrent sendStream calls, so things might interleave and when one of the turns finishes it could set `_busy` false while the other streams etc

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
```diff
@@ -0,0 +1,485 @@
...
+  /// turn alive; `'strict'` throws on malformed JSON or unknown components
+  /// (best during development); `'off'` passes them through unchecked.
+  String? get validate;
+
+  /// Surface id policy. Provide a fixed id to reuse for every surface. Defaults
+  /// to a fresh UUID per surface.
+  String? get surfaceId;
+
+  /// Protocol version stamped on emitted envelopes. Defaults to `'v0.9'`.
+  String? get version;
```
cabljac, line 69: nit: the JS version of this constrains `version` to a SUPPORTED_VERSIONS here it passes through unchecked

cabljac (approved):
Nothing blocking that i can tell, seems to match or be ahead of JS everywhere.

Left some nits

packages/genkit_a2ui/lib/src/parser.dart
pavelgj: Good catch, fixed for parity with JS. Normalization now spreads the original envelope and only stamps `version`, so open-ended/future top-level keys survive. Added a parser test with extra keys (`traceId`, `extra`).

testapps/a2ui/lib/main.dart
pavelgj: Done. `_onSurfaceSubmit` now bails early with `if (_busy) return;`, so a button press mid-stream can't kick off a second concurrent turn.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
pavelgj: Done. Added `_parseVersion`, which validates against a new `supportedA2uiVersions` set and throws on anything unsupported, matching the JS behavior. Added middleware tests for an accepted and a rejected version.

PR #348, opened 2026-08-07
refactor(genkit)!: introduce pluggable instrumentation provider system

pavelgj:
# Pluggable, OpenTelemetry-decoupled telemetry instrumentation

## Summary

Reworks Genkit Dart's telemetry so instrumentation is a pluggable, externally
configurable abstraction instead of a hardcoded OpenTelemetry (OTel) dependency.
`runInNewSpan` becomes a thin dispatcher over one or more configured
`Instrumentation` providers whose span hook behaves like middleware (it wraps
`next`).

By default Genkit is **not** instrumented at all. In the dev environment a
built-in provider is auto-injected (when a telemetry server is available) so the
Developer UI works out of the box. That built-in runs completely independently
of OpenTelemetry: it posts spans directly to the Genkit telemetry server over
HTTP. It also forwards `package:logging` records, correlated with the active
span, to the same server so logs appear in the Developer UI alongside traces.
Third-party observability platforms can implement the `Instrumentation`
interface without taking an OTel dependency.

```dart
import 'package:genkit/telemetry.dart';

void main() {
  // Optional: stack one or more providers before creating Genkit.
  configureInstrumentation(myInstrumentation());

  final ai = Genkit(/* ... */);
}
```

## Example: a minimal custom instrumentation

A provider just implements `Instrumentation.runInNewSpan`: it wraps `next` and
returns its result. This one logs each action's type, name, and input before the
call, its output after, and any error if it throws. It does not track trace ids,
so it calls `next()` with no `SpanContext`.

```dart
import 'package:genkit/telemetry.dart';

class PrintingInstrumentation implements Instrumentation {
  @override
  Future<O> runInNewSpan<O>(
    SpanMetadata metadata,
    Future<O> Function([SpanContext? span]) next,
  ) async {
    final label = '${metadata.actionType ?? 'op'}/${metadata.name}';
    print('-> $label input: ${metadata.input}');
    try {
      final output = await next();
      print('<- $label output: $output');
      return output;
    } catch (e) {
      print('!! $label error: $e');
      rethrow;
    }
  }
}

void main() {
  configureInstrumentation(PrintingInstrumentation());
  final ai = Genkit(/* ... */);
}
```

## Motivation

- The old `runInNewSpan` was hard-wired to OTel: it read the global tracer
  provider, created spans, and encoded `genkit:*` attributes inline. There was
  no way to plug in a different tracing backend, and every consumer paid for
  OTel whether they wanted it or not.
- The old dev instrumentation relied on a very specifically configured
  collector/exporter pointing at the Genkit telemetry server. That was fragile
  when a user also wanted their own OpenTelemetry configuration (custom config,
  exporters, etc.): Genkit's global-tracer setup and the user's could collide.
  The dev path now runs independently of OTel entirely, so it never interferes
  with a user's OpenTelemetry configuration.
- We want third-party observability backends (and an OTel semantic-convention
  provider) to be opt-in and composable.

### New public API - `package:genkit/telemetry.dart`
Intentionally minimal, OTel-free surface for authoring a custom provider:
- `Instrumentation` - middleware-style span hook: `runInNewSpan(metadata, next)`,
  where `next` optionally receives a `SpanContext`.
- `SpanMetadata` - backend-agnostic span description (`name`, `actionType`,
  `input`, `attributes`).
- `SpanContext` - live-span handle (`traceId`, `spanId`, `setMetadata(...)`),
  no OTel types leaked.
- `configureInstrumentation(...)` - register a provider.
- `DisposableInstrumentation` - optional capability for a provider that holds
  resources (subscriptions, clients); Genkit disposes it on `Genkit.shutdown()`.

Log capture is deliberately kept off this minimal authoring surface: it is an
internal detail of the built-in dev provider (see below), so core stays
unopinionated about where logs come from.

Everything else (the dev builtin, the sink/span-data layer, the dispatcher
entry points, the double-inject guard) is internal implementation detail and can
be promoted later if a real need appears.

### Chaining dispatcher - `o11y/instrumentation.dart` (no OTel import)
- Holds an ordered list of providers; `configureInstrumentation` appends, so
  multiple providers can be active at once.
- `runInNewSpan` composes providers as middleware (index 0 outermost, the user
  `fn` innermost) and hands `fn` a composite `SpanContext` that fans
  `setMetadata` out to every provider's span and resolves `traceId`/`spanId` to
  the first non-empty value in the chain.
- The provider list is snapshotted per call, so reconfiguring instrumentation
  during an async gap cannot corrupt an in-flight chain.
- Zero providers configured -> runs `fn` directly with a no-op span (empty ids);
  nothing telemetry-related is touched.

### Built-in dev provider (internal, no OTel)
- `DirectHttpInstrumentation` mints its own trace/span ids, tracks parentage via
  the current `Zone`, times each operation, and serializes finished spans to a
  `TelemetrySink`. It reproduces the `genkit:*` attribute conventions the
  Developer UI relies on, and exports on both span start (live view) and end.

### Log capture (built-in dev provider, no OTel)
- The dev provider bridges `package:logging` records from `Logger.root` to the
  telemetry server as OTLP logs, correlated with the active span (trace/span ids
  resolved from the record's zone). Dart `Level` is mapped to OTel severity
  numbers/text; it skips the sink's own logger to avoid a
  log -> failed-export -> log feedback loop. Opt out via `captureLogs: false`.
- The bridge lives in the concrete provider, not Genkit core, so core stays
  unopinionated about where logs come from. An OTel-based provider brings its own
  logger integration (mirroring how the Dartastic OTel SDK leaves
  `package:logging` bridging to a separate integration).
- Lifecycle: the subscription is cancelled by `dispose()` (via
  `DisposableInstrumentation`), which `Genkit.shutdown()` calls.

### Action context captured on the span (with redaction)
- Action runs now record the execution context as the `genkit:metadata:context`
  span attribute (matching JS). The context is shallow-copied and top-level
  `auth` and `secrets` keys are redacted to `<redacted>`; serialization is
  wrapped in try/catch so telemetry never crashes the action.

### Backend-agnostic span/sink layer (internal) - `o11y/telemetry/`
- `span_data.dart`: `GenkitSpanData`, `GenkitLogData`, the `TelemetrySink` seam
  (`export(spans)` + `exportLogs(logs)`), and `encodeResourceSpans` /
  `encodeResourceLogs` (OTLP/JSON grouped by resource + scope). No OTel types.
- `collector_http_sink.dart`: `CollectorHttpSink` POSTs OTLP/JSON (`resourceSpans`
  and `resourceLogs`) to `$server/api/otlp`, fire-and-forget so telemetry never
  blocks the operation.
- `telemetry_platform.dart` (+ io/web/stub): resolves `GENKIT_TELEMETRY_SERVER`
  per platform; an empty value is treated as unset on all platforms.

### Dev wiring - `genkit_class.dart`
- Removed the unconditional collector-exporter setup from the constructor.
- Auto-instrumentation is gated on **both** dev mode (`GENKIT_ENV=dev` /
  `isDevEnv`) **and** a configured telemetry server. The dev builtin returns
  `null` when there is no server, so Genkit stays uninstrumented.
- A builtin marker prevents duplicate setup and respects a user-configured
  provider.

### CLI telemetry handshake - reflection v1 and v2
Matches the other SDKs: `GENKIT_TELEMETRY_SERVER` takes precedence; if it is
unset and the CLI sends a telemetry server URL over the reflection handshake,
the dev instrumentation is enabled late.
- V2 (websocket): handles the `configure` JSON-RPC notification.
- V1 (HTTP): reads `telemetryServerUrl` from `POST /api/notify`.
- Both no-op when the env var is set or a builtin is already registered.

### Empty ids are omitted, not sent blank
When Genkit is uninstrumented, trace/span ids are empty. Rather than surfacing
blank ids (which look like a broken exporter), the reflection servers and
`genkit_shelf` now omit `x-genkit-trace-id` / `x-genkit-span-id` headers, the
`runActionState` notification, and the `telemetry` payload when the id is empty.

## Behavior matrix

| Environment | `GENKIT_TELEMETRY_SERVER` | Result |
| --- | --- | --- |
| production | (any) | not instrumented unless the user configures a provider |
| dev | unset, no handshake | not instrumented (no spans) |
| dev | set (e.g. Genkit CLI) | built-in direct-HTTP instrumentation auto-injected; traces posted to the server |
| dev | unset, CLI handshake sends a URL | built-in direct-HTTP instrumentation enabled from the handshake |
| any | user calls `configureInstrumentation(...)` | user providers active (and coexist with the dev builtin) |

## Breaking changes

We are in beta and intentionally avoided API debt:
- Removed the `TelemetryContext` record; `runInNewSpan`'s callback now receives a
  `SpanContext`. The outer `runInNewSpan` signature is unchanged, so call sites
  that ignore the arg are untouched.
- Genkit is no longer instrumented by default in production. Enable telemetry via
  `configureInstrumentation(...)` (or run under the Genkit CLI in dev).
- The built-in dev instrumentation no longer uses OpenTelemetry. Optional OTel
  instrumentation will ship as a separate provider (see follow-ups).

## Testing

- `test/o11y/instrumentation_test.dart`: fake provider covering middleware wrap
  order, `setMetadata` fan-out, first-non-empty id resolution, error
  propagation, and no-instrumentation passthrough; direct-HTTP nested-span
  parent/child + non-zero ids and start/end export; the dev auto-inject gate
  (null without a server, no double-instrument, env precedence); the mid-flight
  reconfigure snapshot; direct-HTTP log capture (bridge to the sink, `Level`
  severity mapping, active-span correlation, empty ids outside a span, ignoring
  the sink's own logger, `captureLogs: false`, and `dispose` stopping capture);
  and provider disposal (`resetInstrumentation` / `disposeInstrumentations`).
- `test/o11y/collector_http_sink_test.dart`: OTLP/JSON encoding for spans and
  logs grouped by resource + scope, and the HTTP sink POST behavior for both.
- `test/core/action_test.dart`: execution context recorded on the span with
  `auth`/`secrets` redacted, and no context attribute when none is supplied.
- `test/core/reflection_v1_test.dart` / `reflection_v2_test.dart`: happy-path
  trace ids when instrumented, telemetry omitted when uninstrumented, and the
  `configure` handshake enabling instrumentation.
- Full `genkit` suite green (635 tests) and `genkit_shelf` green;
  `dart analyze` and `dart format` clean.

## Follow-ups (out of scope)

- Metrics hooks on the `Instrumentation` interface.
- A dedicated `package:genkit/otel.dart` entrypoint with an OTel GenAI-semantic-convention provider.


packages/genkit/lib/src/o11y/instrumentation.dart
```diff
@@ -13,110 +13,144 @@
...
+    if (index == _instrumentations.length) {
+      return _runWithSpan(_CompositeSpanContext(spans), fn);
     }
+    return _instrumentations[index].runInNewSpan(metadata, (span) {
+      spans.add(span);
+      return build(index + 1);
+    });
+  }
+
+  return build(0);
```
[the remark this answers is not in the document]
pavelgj: Good catch. Fixed in 1f4bd333 by snapshotting the provider list at the start of `runInNewSpan` before descending the chain.

packages/genkit/lib/telemetry.dart
```diff
@@ -0,0 +1,49 @@
...
+///
+/// void main() {
+///   configureInstrumentation(myInstrumentation());
+///   final ai = Genkit(/* ... */);
+/// }
+/// ```
+///
+/// By default Genkit is not instrumented. In the dev environment the built-in
+/// OpenTelemetry instrumentation (`genkitDevInstrumentation`) is auto-injected
+/// so the Developer UI works out of the box.
```
huangjeff5, line 34: This is true under `genkit start` (CLI sets the collector). It is not true for `GENKIT_ENV=dev` / `Genkit(isDevEnv: true)` without `GENKIT_TELEMETRY_SERVER`. Reflection still comes up; flows run; traces do not.

Can we say “auto-injected when both dev mode and a collector are present”? Same overclaim is on the `OtelInstrumentation` class comment.

packages/genkit/lib/src/o11y/instrumentation.dart
```diff
@@ -13,110 +13,148 @@
...
+) {
+  return runZoned(() => fn(span), zoneValues: {_spanContextKey: span});
+}
+
+/// A [SpanContext] used when no instrumentation is configured.
+class _NoopSpanContext implements SpanContext {
+  const _NoopSpanContext();
+
+  @override
+  String get traceId => '';
```
huangjeff5, line 121: Default-off is fine. Empty ids serialized into `RunResult`, `onTraceStart`, `x-genkit-trace-id`, and `telemetry: {traceId: ''}` are not.

```dart
final ai = Genkit(); // production
final result = await flow.run('hi');
print(result.traceId); // ''
```

v1 writes `x-genkit-trace-id` / `x-genkit-span-id` from `onTraceStart`; v1 and v2 write `telemetry: {traceId: result.traceId}`. Silent `''` looks like a broken exporter.

If we stay default-off: name the migration (`configureInstrumentation(OtelInstrumentation())`) in README/CHANGELOG, and skip flushing those headers when nothing is instrumented.

packages/genkit/lib/telemetry.dart
```diff
@@ -0,0 +1,49 @@
...
+        Instrumentation,
+        SpanContext,
+        SpanMetadata,
+        configureInstrumentation,
+        isInstrumentedBy,
+        resetInstrumentation,
+        runInNewSpan,
+        setCustomMetadataAttributes;
+export 'src/o11y/otel_instrumentation.dart'
+    show OtelInstrumentation, genkitDevInstrumentation;
```
huangjeff5, line 49:
```dart
import 'package:genkit/telemetry.dart';
import 'package:opentelemetry/sdk.dart' as sdk;

configureInstrumentation(
  OtelInstrumentation(tracerProvider: sdk.TracerProviderBase()),
);
```

This puts `package:opentelemetry` types on the new public surface.

The PR body calls this an OTel-free authoring surface and lists a `package:genkit/otel.dart` split as a follow-up. Can we do that split now, so `telemetry.dart` only exports `Instrumentation` / `SpanMetadata` / `SpanContext` / `configureInstrumentation` / `resetInstrumentation` / `isInstrumentedBy`?

huangjeff5 (commented):
The CLI path (`genkit start` sets `GENKIT_TELEMETRY_SERVER`) looks right. A few things before I'd approve — two product questions:

1. **Late handshake.** JS/Go/Python take `telemetryServerUrl` from the reflection handshake and start exporting after the runtime is already up. Dart v2 `_handleResponse` logs the register result and drops it; v1 `/api/notify` acks and never reads the body. This PR does not touch those files — the ignore is pre-existing — but default-off means "no env at `Genkit()`" now produces empty `traceId`s instead of spans with nowhere to go. `genkit start` still works because it sets the env up front. The gap is `Genkit(isDevEnv: true)` / `GENKIT_ENV=dev` without the collector, then the CLI attaching later. Do we want late apply, or is env-at-construction the Dart rule?

2. **`OtelInstrumentation({TracerProvider?})` on `package:genkit/telemetry.dart`.** That puts `package:opentelemetry` types on the new public surface. Fine to keep the built-in provider in this PR — can the OTel types live in `package:genkit/otel.dart` (or a plugin) so `telemetry.dart` stays the backend-agnostic installer?

Also a land-order note: #350 also edits `action_test.dart`.

Lastly, since production telemetry is now opt-in, please document the `configureInstrumentation(...)` migration in `README.md` / `CHANGELOG.md` so users know how to enable tracing in production.

packages/genkit/lib/src/o11y/telemetry/telemetry_io.dart
```diff
@@ -13,11 +13,15 @@
...
 import 'dart:io' as io;
+import 'package:opentelemetry/api.dart' as api;
 import 'exporter_impl.dart';
 
-void configureCollectorExporter() {
+/// Configures the OTLP collector exporter when `GENKIT_TELEMETRY_SERVER` is set
+/// and returns the backing tracer provider (or `null` if not configured).
+api.TracerProvider? configureCollectorExporter() {
   final telemetryServer = io.Platform.environment['GENKIT_TELEMETRY_SERVER'];
   if (telemetryServer != null) {
```
huangjeff5, line 23: Web uses `isNotEmpty`. io treats `GENKIT_TELEMETRY_SERVER=` as set, so `setupExporter('')` builds `CollectorHttpExporter('/api/otlp')` and we instrument. Same env, different whether every action is wrapped.

Use `isNotEmpty` here too?

packages/genkit/test/core/action_test.dart
```diff
@@ -39,9 +40,18 @@ void main() {
   api.registerGlobalTracerProvider(provider);
 
   group('Action', () {
-    setUp(exporter.reset);
+    setUp(() {
+      exporter.reset();
+      // Actions rely on a configured instrumentation to emit spans. Route
+      // through the provider registered above (the dev factory only
+      // auto-configures when a collector is set).
+      configureInstrumentation(OtelInstrumentation(tracerProvider: provider));
```
huangjeff5, line 48: CI is green because nothing asserts the new gate. This file forces `configureInstrumentation(OtelInstrumentation(...))` in `setUp`, and the downstream assertion is `isA<String>()` — true for `''`. The o11y tests never call `isInstrumentedBy`, `genkitDevInstrumentation()`, or `GENKIT_TELEMETRY_SERVER`.

Please add tests that fail if these disappear:

- `genkitDevInstrumentation()` is null when the collector is unset
- `Genkit(isDevEnv: true)` then has `isInstrumentedBy<OtelInstrumentation>() == false`
- a pre-registered `OtelInstrumentation` is not appended again
- the `List.of(_instrumentations)` snapshot in `runInNewSpan` (Gemini’s comment; still untested after `1f4bd333`)

packages/genkit/lib/src/genkit_class.dart
```diff
@@ -102,6 +103,19 @@ final class Genkit extends GenkitAI {
...
     if (isDevEnv ?? utils.isDevEnv) {
+      // In the dev environment, auto-inject the built-in OpenTelemetry
+      // instrumentation (unless it is already configured) so the Developer UI
+      // receives traces. The factory wires up the collector exporter and routes
+      // Genkit's spans through its tracer provider; it returns null (and we do
+      // not instrument) when no collector is configured
+      // (`GENKIT_TELEMETRY_SERVER` unset). In production, Genkit is not
+      // instrumented unless the user configures a provider.
+      if (!isInstrumentedBy<OtelInstrumentation>()) {
+        final devInstrumentation = genkitDevInstrumentation();
```
huangjeff5, line 114: This decides Dev UI export once, at construction, from `GENKIT_TELEMETRY_SERVER`. Reflection still starts either way.

JS applies a late URL in `setTelemetryServerUrl` (`js/core/src/reflection-v2.ts`); Go in `configureTelemetry`; Python in `_apply_handshake_telemetry`. Dart already ignored the handshake before this PR. What changed is: no env → `_NoopSpanContext` and empty ids, so the old ignore is now user-visible.

```dart
// GENKIT_ENV=dev, GENKIT_TELEMETRY_SERVER unset
final ai = Genkit();
// CLI later sends {telemetryServerUrl: 'http://127.0.0.1:4033'}
await ai.defineFlow('echo', (i) async => i).run('hi');
// traceId == '', collector gets nothing
```

Is the product “env at `Genkit()` or never,” or should the handshake turn export on the way the other SDKs do?

packages/genkit/lib/telemetry.dart
pavelgj: Good catch, reworded. The library doc now says the built-in is auto-injected in dev mode only when a Genkit telemetry server is configured (`GENKIT_TELEMETRY_SERVER` or the CLI reflection handshake), and points to `configureInstrumentation` for production. The `OtelInstrumentation` class comment is moot now: that OTel-backed dev path was removed. The dev instrumentation runs independently of OpenTelemetry and posts directly to the telemetry server.

packages/genkit/lib/src/o11y/instrumentation.dart
pavelgj: Fixed (option a). When uninstrumented, ids are empty and we now omit rather than emit blank:

- `reflection_v1.dart`: `onTraceStart` skips `x-genkit-trace-id`/`x-genkit-span-id` when empty; the `telemetry` payload is dropped when the traceId is empty.
- `reflection_v2.dart`: the `runActionState` notification is skipped for empty ids, and `telemetry` is omitted.
- `genkit_shelf.dart`: trace/span headers are set only when non-empty.

The migration is called out in the PR description (production is off by default; enable with `configureInstrumentation(...)` from `package:genkit/telemetry.dart`) rather than the CHANGELOG, since that file is autogenerated from PR descriptions. Added tests: v1 "omits telemetry when uninstrumented", and the happy path now asserts a real non-empty traceId.

packages/genkit/lib/src/genkit_class.dart
pavelgj: Went with: `GENKIT_TELEMETRY_SERVER` wins if set; otherwise the reflection handshake can enable dev instrumentation late, matching the other SDKs. Wired both transports:

- V2 (websocket): added the `configure` JSON-RPC notification handler, parsing `telemetryServerUrl`.
- V1 (HTTP): `POST /api/notify` now reads `telemetryServerUrl` from the body.

Both funnel through an internal helper that no-ops when the env var is set (env precedence) or a builtin is already registered (no double-instrument). Since `runInNewSpan` snapshots the provider list per call, a late handshake takes effect on the next action run, not one already in flight, matching JS's late-apply. Added a v2 test asserting the `configure` handshake flips instrumentation on when the env var is unset.

packages/genkit/lib/telemetry.dart
pavelgj: Addressed, and then some. Two things happened:

1. The OTel-backed instrumentation was dropped from this branch entirely, so `telemetry.dart` no longer exports any `package:opentelemetry`/dartastic types.
2. We then trimmed the surface further to be conservative about new public API. `telemetry.dart` now exports only `Instrumentation`, `SpanContext`, `SpanMetadata`, and `configureInstrumentation`, which is what you need to author a custom provider. The dev builtin, sink/span-data layer, and dispatcher entry points are all internal now. `resetInstrumentation` stays internal and is marked `@visibleForTesting`.

The optional OTel provider lands in a separate PR and can introduce its own entrypoint (e.g. `package:genkit/otel.dart`) at that point.

packages/genkit/test/core/action_test.dart
pavelgj: Added tests that fail if the gate regresses (`test/o11y/instrumentation_test.dart`, "dev instrumentation gate" group):

- `genkitDevInstrumentation()` is null when `GENKIT_TELEMETRY_SERVER` is unset.
- `Genkit(isDevEnv: true)` with no server leaves `isInstrumentedBy<GenkitBuiltinInstrumentation>() == false`.
- a pre-registered builtin is not appended again (asserted by running a span and checking only the pre-registered provider wraps it).
- the `List.of(_instrumentations)` snapshot: a new "runInNewSpan snapshot" test reconfigures instrumentation mid-flight and asserts the in-flight chain stays consistent and doesn't throw.

Note the symbol names shifted since your comment (`OtelInstrumentation` -> the internal `GenkitBuiltinInstrumentation` marker). Tests that need a real trace id now configure the direct-HTTP instrumentation explicitly instead of relying on `isA<String>()` matching `''`. The gate tests self-skip if `GENKIT_TELEMETRY_SERVER` happens to be set locally so they stay deterministic.

packages/genkit/lib/src/o11y/telemetry/telemetry_io.dart
pavelgj: Fixed. `genkitTelemetryServerUrl()` on io now treats an empty value as unset (returns null), matching web, so the same env produces the same instrumentation decision on both platforms.

packages/genkit/lib/src/o11y/instrumentation_api.dart
```diff
@@ -0,0 +1,94 @@
...
+/// A pluggable telemetry instrumentation provider.
+///
+/// The [Instrumentation.runInNewSpan] hook behaves like middleware: it receives
+/// a `next` function and is expected to invoke it (usually inside its own span),
+/// then return its result. Providers may inspect the awaited result to record
+/// outputs, catch errors to record exceptions, and propagate their own context
+/// around the `next` call.
+///
+/// Configure providers via `configureInstrumentation` (see `instrumentation.dart`).
+
```
MichaelDoyle:
```suggestion
```

MichaelDoyle: So the docs generate correctly. :)

MichaelDoyle (approved):
I can't get fully into the depths of dart specific idioms, but everything appears to preserve the existing behavior and decouples from otel nicely. I particularly like the log bridging. It's going to be very nice for customers/users of Genkit.

packages/genkit/lib/src/o11y/instrumentation_api.dart
```diff
@@ -0,0 +1,94 @@
...
+  /// The Genkit action type, e.g. `'flow'`, `'model'`, `'tool'`.
+  final String? actionType;
+
+  /// The raw input passed to the wrapped operation.
+  ///
+  /// Providers decide whether and how to encode this (e.g. JSON). May be `null`.
+  final Object? input;
+
+  /// Additional static attributes supplied by the caller.
+  final Map<String, String> attributes;
```
MichaelDoyle: Should attributes here be `Map<String, Object?>` rather than `Map<String, String>`?

OpenTelemetry attributes often include ints and booleans (e.g. status codes, retry counts), and `GenkitSpanData` / `_encodeAttributeValue` already support encoding them into intValue and boolValue. Widening this to `Map<String, Object?>` would match `SpanContext.setMetadata` and prevent callers from having to stringify numbers and booleans up front.

huangjeff5 (approved):

packages/genkit_shelf/lib/genkit_shelf.dart
```diff
@@ -194,8 +194,10 @@ Handler shelfHandler(Action action, {ContextProvider? contextProvider}) {
           jsonEncode({'result': result.result}),
           headers: {
             'Content-Type': 'application/json',
-            'x-genkit-trace-id': result.traceId,
-            'x-genkit-span-id': result.spanId,
+            // Omit trace/span headers when uninstrumented (empty ids): a blank
+            // value looks like a broken exporter to clients.
+            if (result.traceId.isNotEmpty) 'x-genkit-trace-id': result.traceId,
+            if (result.spanId.isNotEmpty) 'x-genkit-span-id': result.spanId,
```
huangjeff5, line 224: A streaming flow run that minted a real trace id should still be correlatable. Non-stream already sends x-genkit-trace-id / x-genkit-span-id when those strings are non-empty, and omits them when they are blank.

stream=true / Accept: text/event-stream returns only Content-Type and Cache-Control. The id is on result and never leaves the handler.

Can the streaming response carry those headers when the id is non-empty, the same way the non-stream path does?

packages/genkit/lib/src/o11y/instrumentation_api.dart
pavelgj: Good call. Widened `SpanMetadata.attributes` (and the `runInNewSpan` param) to `Map<String, Object?>`. No caller passes `attributes:` today, and the built-in provider already copies into a `Map<String, Object?>` span whose encoder handles String/bool/int/double, so this is a clean widening with no migration.

packages/genkit_shelf/lib/genkit_shelf.dart
pavelgj: Done. The streaming branch now captures the ids via `onTraceStart` into a `Completer` and awaits it before returning the `Response`, then emits `x-genkit-trace-id`/`x-genkit-span-id` when non-empty, using the same blank-id guard as the non-stream path. `catchError` completes the completer with empty ids so a pre-span failure cannot deadlock header emission, and chunks buffer in the `StreamController` during the brief await. Added shelf tests for both the instrumented (headers present) and uninstrumented (headers omitted) cases.

PR #350, opened 2026-08-17
feat(genkit)!: redesign tool API around ToolResult + multipart, type actionType with ActionType

pavelgj:
Tools now return a `ToolResult<Output>` instead of a bare value, using Dart dot-shorthand for ergonomics, and support multipart tool responses (structured output plus media parts). Also refactors the stringly-typed `actionType` into a typed `ActionType` extension type. Both changes are breaking (pre-1.0).

Primary: new tool API
- `ToolFn` returns `FutureOr<ToolResult<Output>>`:
  - `return .response(output)` for a normal result
  - `return .response(output, parts: [...])` for a multipart result (for example: images or other media returned alongside the structured output)
  - `return .interrupt(data)` to interrupt the generation loop
- `ctx.interrupt(...)` still works (soft-deprecated).
- Every tool implements the multipart contract and registers under `tool.v2`, so the Dev UI treats tools as multipart. Model plugins that support multipart (for example: the Gemini plugins) map the tool response `content` into their native multipart function-response parts.
- MCP behavior is unchanged aside from the mechanical `.response(...)` wrap.

Minor: ActionType refactor
- `ActionType` extension type with named constants (model, flow, embedder, tool, executablePrompt, agent, ...). Open-ended: `ActionType('custom-type')` still works.
- `Action.actionType`, `ActionMetadata.actionType`, and `Registry.lookupAction` are typed as `ActionType`.

Migration (tool API):

  BEFORE:
  ```dart
    ai.defineTool(
      name: 'getWeather',
      description: 'Get the weather for a location',
      inputSchema: WeatherInput.$schema,
      fn: (input, ctx) async {
        if (ctx.resumed == null) {
          ctx.interrupt({'requiresConfirmation': true});
        }
        return 'Sunny in ${input.location}';
      },
    );
```

  AFTER:
```dart
    ai.defineTool(
      name: 'getWeather',
      description: 'Get the weather for a location',
      inputSchema: WeatherInput.$schema,
      fn: (input, ctx) async {
        if (ctx.resumed == null) {
          return .interrupt({'requiresConfirmation': true});
        }
        return .response('Sunny in ${input.location}');
      },
    );
```

  Multipart (new capability):
```dart
    fn: (input, ctx) async => .response(
      {'result': 'captured'},
      parts: [
        MediaPart(media: Media(contentType: 'image/png', url: dataUri)),
      ],
    );
```

  Reading a tool's direct result (e.g. tool.call/runRaw): it is now a `ToolResult`; read `(result as ToolResponseResult).output`.

Migration (ActionType):
- `registry.lookupAction(...)` now takes an `ActionType`: pass a named constant (`ActionType.model`) or dot-shorthand in typed positions (`lookupAction(.tool, name)`); custom types via `ActionType('my-type')`.
- Tools resolve under `tool.v2`; anything running a tool by key must use `/tool.v2/<name>`.
- Plugin `resolve(...)` still receives the actionType as a string.

BREAKING CHANGE: tool functions must return `ToolResult` (`.response(...)` / `.interrupt(...)`), tools are registered/resolved under `tool.v2`, and `actionType` fields/params are now typed as `ActionType`.


packages/genkit/lib/src/ai/tool.dart
```diff
@@ -41,31 +43,139 @@ class ToolFnArgs<Input> {
   dynamic get resumed => toolRequest?.metadata?['resumed'];
 
   /// Interrupts the generation loop with optional [data].
+  ///
+  /// Prefer returning `.interrupt(data)` from your tool function instead. This
+  /// throwing form remains available for cases where you are several helper
+  /// calls deep and cannot easily return a [ToolResult].
```
pavelgj:
```suggestion
```

packages/genkit/lib/src/ai/generate_bidi.dart
```diff
@@ -143,16 +144,23 @@ Future<GenerateBidiSession> runGenerateBidi(
...
-                ToolResponsePart(
-                  toolResponse: ToolResponse(
-                    ref: toolRequest.toolRequest.ref,
-                    name: toolRequest.toolRequest.name,
-                    output: output.result,
-                  ),
+              final result = (await tool.runRaw(
+                toolRequest.toolRequest.input,
+              )).result;
+              final response = switch (result) {
```
huangjeff5:
```dart
return .interrupt({'confirm': true});
```

`generate()` pauses and gives the tool request back to the caller. `generateBidi()` sends `{interrupt: {confirm: true}}` to the live model as the tool's answer and keeps the session going.

A live session has to reply to the model or the function call sits open, so this may be intentional. `defineInterrupt` is documented against a follow-up `generate` call, not bidi.

If HITL is unary-only, this arm should not look like a successful tool payload when the same tool is reused in a live session. Is that the intended bidi behavior?

huangjeff5 (commented):
README samples still use the old tool return type, so they will not compile.

The first sample a new user copies (`packages/genkit/README.md:111`) is:
```dart
fn: (input, _) async {
  return 'Weather in ${input.location}: 72°F and sunny';
},
```

`ToolFn` now returns `FutureOr<ToolResult<Output>>`. The interrupt section (L190) still calls `context.interrupt(input.question)` and returns nothing, and L241 still does `return 'Action confirmed'`. Same stale shape in the `genkit_google_genai` and `genkit_mcp` READMEs. The testapps were migrated; the READMEs were not.

(README.md is not in the PR diff. Keep this in the review body. Do not try to attach it as a line comment on README.md.)

packages/genkit/lib/src/ai/tool.dart
```diff
@@ -41,31 +43,135 @@ class ToolFnArgs<Input> {
...
-  }) : super(
-         fn: (input, ctx) {
+    SchemanticType<Input>? inputSchema,
+    this.toolOutputSchema,
+    Map<String, dynamic>? metadata,
+  }) : toolFn = fn,
+       super(
+         name: name,
+         description: description,
+         inputSchema: inputSchema,
```
apascal07, line 206: `outputSchema` is no longer forwarded to `Action`, so `Action.outputSchema` is null for every tool. Three readers still use it and were not migrated to `toolOutputSchema`. All three fail silently, because the base field still exists and is still typed:

- `genkit_mcp/lib/src/util/convert_tools.dart:35`: `tools/list` now omits `outputSchema` for every tool served by `GenkitMcpServer`, while `_callTool` still emits `structuredContent`. Clients get structured output they cannot validate.
- `core/reflection/reflection_v1.dart:164` and `reflection_v2.dart:218`: the Dev UI action manifest carries no output schema for any tool.

Nothing catches it. `grep outputSchema packages/genkit_mcp/test/` is empty.

packages/genkit/lib/src/ai/tool.dart
```diff
@@ -41,31 +43,135 @@ class ToolFnArgs<Input> {
...
+  Map<String, dynamic> toJson();
+}
+
+/// A normal tool response carrying structured [output] and optional multipart
+/// [parts] and [metadata].
+final class ToolResponseResult<Output> extends ToolResult<Output> {
+  /// The structured output of the tool.
+  final Output output;
+
+  /// Optional multipart content (images, media, etc.) returned alongside the
```
apascal07, line 141: Only the Gemini converter maps `ToolResponse.content`. The other three model plugins read `output` alone, so a tool written with `parts:` silently loses its media depending on which provider runs it:

- `genkit_anthropic/lib/src/plugin_impl.dart:278`: `ToolResultContent.text(jsonEncode(res.output))`
- `genkit_openai/lib/src/converters.dart:53`: `content: jsonEncode(toolResponse.output)`
- `genkit_firebase_ai/lib/genkit_firebase_ai.dart:528`: `{'result': toolResponse.output}`

Grepping those three packages' `lib/` for any tool-response `.content` access returns nothing. There is no log and no error, so the author of a multipart tool has no way to discover the parts were discarded.

packages/genkit/lib/src/ai/tool.dart
```diff
@@ -41,31 +43,135 @@ class ToolFnArgs<Input> {
...
+
+class Tool<Input, Output>
+    extends Action<Input, ToolResult<Output>, void, void> {
+  /// The user-declared output schema (the schema of `Output`, not
+  /// [ToolResult]). Used to build the model-facing tool definition.
+  final SchemanticType<Output>? toolOutputSchema;
 
-class Tool<Input, Output> extends Action<Input, Output, void, void> {
+  /// The underlying tool implementation function.
+  final ToolFn<Input, Output> toolFn;
```
apascal07: `toolFn` has no readers. A repo-wide grep finds only this declaration and the `toolFn = fn,` initializer on line 154.

It is public and exported, so `tool.toolFn(input, ToolFnArgs(ctx))` becomes a supported call that skips everything the wrapper adds: the null-input `ArgumentError` on line 164, `runRaw`'s `inputSchema.parse`, the `runInNewSpan` span, the interrupt `setCustomMetadataAttributes` on line 171, and the `runZoned(zoneValues: {ToolRequestPart: toolRequest})` binding at `generate.dart:991` that makes `ctx.toolRequest` and `ctx.resumed` work.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -958,14 +956,27 @@ _executeTools(
...
+          throw ToolInterruptException(data ?? true);
+        case ToolResponseResult(:final output, :final parts, :final metadata):
+          return ToolResponsePart(
+            toolResponse: ToolResponse(
+              ref: req.toolRequest.ref,
+              name: req.toolRequest.name,
+              output: output,
+              content: parts?.map((p) => p.toJson()).toList(),
+            ),
+            metadata: metadata,
```
apascal07, line 998: `content` and `metadata` are built here but do not survive an interrupt. `_ToolStatus` (line 35) is `({Object? output, ToolInterruptException? interrupt})`, line 1003 records only `toolResponsePart.toolResponse.output`, and when any tool in the turn interrupts, line 384 discards `execution.toolResponses` entirely.

For a turn where `screenshot` returns `.response(out, parts: [MediaPart(image)])` and `confirmTransfer` returns `.interrupt(...)`: `_buildInterruptedResponse` stores only `meta['pendingOutput'] = status.output` (line 886), and on resume `_resolveResume` rebuilds `ToolResponse(ref:, name:, output:)` at line 821 with no `content` and no `metadata`. The image never reaches the model. The restart path at line 551 loses it the same way.

`tool_result_test.dart:148` covers multipart only on the straight-through path.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -958,14 +956,27 @@ _executeTools(
...
+          // Reuse the existing interrupt machinery: bubble the request back to
+          // the caller as a thrown interrupt.
+          throw ToolInterruptException(data ?? true);
+        case ToolResponseResult(:final output, :final parts, :final metadata):
+          return ToolResponsePart(
+            toolResponse: ToolResponse(
+              ref: req.toolRequest.ref,
+              name: req.toolRequest.name,
+              output: output,
+              content: parts?.map((p) => p.toJson()).toList(),
```
apascal07, line 996: Nothing wrote `ToolResponse.content` before this PR, but `genkit_mcp` already reads it, and it expects MCP content blocks rather than Genkit `Part` JSON.

`genkit_mcp/lib/src/util/convert_messages.dart:151` passes them straight through:

```dart
if (response.content is List) {
  final items = response.content as List;
  final blocks = items.whereType<Map>().map((e) => e.cast<String, dynamic>()).toList();
  if (blocks.isNotEmpty) return blocks;
}
```

Serve a prompt whose history contains a multipart tool response (`mcp_server.dart:331` to `toMcpPromptMessages` to `toMcpPromptMessage:106`) and the `tool_result` block's `content` becomes `[{'media': {...}}]`, which has no `type` key and is not a valid MCP content block. The early return also drops the tool's output text. `generate_bidi.dart:160` writes the same encoding.

packages/genkit/lib/src/ai/generate_bidi.dart
```diff
@@ -143,16 +144,23 @@ Future<GenerateBidiSession> runGenerateBidi(
...
+                ToolInterruptResult(:final data) => ToolResponse(
```
apascal07: This turns an interrupt into an ordinary tool response and feeds it to the live model. Line 163 wraps it as a normal `ToolResponsePart` and line 177 sends it, so the model receives the literal payload `{"interrupt": true}` and keeps generating. The caller gets no `finishReason: interrupted`, no `interrupts` list, and no resume path.

That makes `defineInterrupt` a no-op in bidi, which contradicts its dartdoc: "the request bubbles back to the caller instead of executing any logic". `generate.dart:966` does the opposite for the same pattern.

If bidi genuinely cannot surface an interrupt, rejecting interrupt-capable tools at `runGenerateBidi` setup would at least make that discoverable before a live session.

packages/genkit/lib/src/ai/generate_bidi.dart
```diff
@@ -143,16 +144,23 @@ Future<GenerateBidiSession> runGenerateBidi(
...
+                  ref: toolRequest.toolRequest.ref,
+                  name: toolRequest.toolRequest.name,
+                  output: {'interrupt': data ?? true},
                 ),
-              );
+                ToolResponseResult(:final output, :final parts) => ToolResponse(
```
apascal07: The pattern binds `output` and `parts` but not `metadata`, and line 163 builds `ToolResponsePart(toolResponse: response)` with no `metadata:` argument, though the constructor accepts one (`types.g.dart:757`).

`generate.dart:969` destructures all three and passes `metadata: metadata` at line 977. So `.response(out, metadata: {...})` keeps its metadata under `generate` and loses it under `runGenerateBidi`, for the same tool.

packages/genkit/lib/src/ai/generate_bidi.dart
```diff
@@ -143,16 +144,23 @@ Future<GenerateBidiSession> runGenerateBidi(
...
+                  ref: toolRequest.toolRequest.ref,
+                  name: toolRequest.toolRequest.name,
+                  output: output,
+                  content: parts?.map((p) => p.toJson()).toList(),
+                ),
+              };
+              toolResponses.add(ToolResponsePart(toolResponse: response));
```
apascal07: The deprecated throwing form does not reach the switch above. `ctx.interrupt(data)` throws `ToolInterruptException`, which lands in the `catch (e)` on the next line and becomes `output: 'Error: $e'`. `ToolInterruptException` (`interrupt.dart:16`) declares no `toString()`, so that renders as `Error: Instance of 'ToolInterruptException'` and the data payload is lost.

`.interrupt(data)` produces `{'interrupt': data}`. The deprecation notice says the throwing form still works, so the two spellings should not give the model different results. `generate.dart:1002` has the `on ToolInterruptException` clause this path is missing.

packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -482,15 +482,25 @@ gcl.Part toGeminiPart(Part p) {
     );
   }
   if (p.isToolResponse) {
+    final tr = p.toolResponse!;
+    // Multipart tool content (images, media, etc.) becomes function-response
+    // parts, mirroring the Genkit JS converter.
+    final contentParts = tr.content
```
apascal07, line 504: This maps tool content through the generic `toGeminiPart`, but `FunctionResponse.parts` is not a list of `Part`. Gemini's `FunctionResponsePart` has exactly two fields, `inlineData` and `fileData` (see the js-genai type definitions).

`.response(out, parts: [TextPart(text: 'summary')])` therefore emits `{'text': 'summary'}` inside `functionResponse.parts`, and the API rejects the request with `400 Invalid JSON payload received. Unknown name "text"`. A `MediaPart` whose url is not a `data:` URL emits `{'fileData': {...}}`, which js-genai marks as not supported in the Gemini API. Only the data-URI media case is valid, and it is the only case `converter_test.dart:97` covers.

Related: the generated binding has no `parts` on `FunctionResponse` (`generativelanguage.dart:3275`) and neither does the generator schema (`tool/generate_client.dart:117`), so `fromJson` here bypasses the typed constructor and nothing checks the shape of any of the four keys. Genkit JS does the same mapping, so this probably needs fixing on both sides.

packages/genkit_mcp/lib/src/server/mcp_server.dart
```diff
@@ -259,8 +259,16 @@ class GenkitMcpServer {
     final input = params['arguments'];
     try {
       final result = await tool.runRaw(input);
-      final output = result.result;
+      // Tool functions return a ToolResult; unwrap it to the raw output (or
+      // the interrupt payload) before serializing for the MCP client.
+      final toolResult = result.result;
+      final output = switch (toolResult) {
```
apascal07: Two problems in this switch.

`parts` and `metadata` are dropped. The pattern binds only `output`, and the reply is hardcoded to a single text block at line 272. MCP's `content` array carries image and audio blocks natively and `convert_messages.dart` already builds them, so a tool returning `.response(out, parts: [MediaPart(image)])` loses its image at this boundary.

An interrupt exits through the success path. Line 280 returns `structuredContent: {'interrupt': true}` with no `isError`, which is set only in the `catch` at line 285. A remote client cannot tell a human-in-the-loop pause from a real answer.

`grep -rn interrupt packages/genkit_mcp/test/` returns nothing, so neither arm is tested.

packages/genkit/lib/src/core/registry.dart
```diff
@@ -101,7 +101,7 @@ class Registry {
           if (_actions.containsKey(key)) {
             return _actions[key];
           }
-          final action = plugin.resolve(actionType, resolvedActionName);
+          final action = plugin.resolve(actionType.value, resolvedActionName);
```
apascal07: The typed `ActionType` stops here. `GenkitPlugin.resolve(String actionType, String name)` (`core/plugin.dart:40`) still takes a raw string, and the value crossing that boundary silently changed from `'tool'` to `'tool.v2'`.

A plugin that resolves dynamic tools with `if (actionType == 'tool')`, which is the pattern the in-repo plugins use for other types (`genkit_firebase_ai.dart:194`, `openai_plugin.dart:199`, `common_plugin.dart:201`), keeps compiling, returns null, and its tools become unresolvable, with `generate` throwing "Tool X not found". No in-repo plugin matches `'tool'`, so nothing here breaks, but nothing warns an external one either.

Typing `resolve` as `ActionType` would make it a compile error, which seems like the point of the refactor.

packages/genkit/lib/plugin.dart
```diff
@@ -38,7 +38,7 @@ export 'package:genkit/src/ai/model.dart'
     show BidiModel, Model, ModelRef, modelMetadata, modelRef;
 export 'package:genkit/src/ai/tool.dart' show Tool, ToolFn, ToolFnArgs;
 export 'package:genkit/src/core/action.dart'
-    show Action, ActionFnArg, ActionMetadata;
+    show Action, ActionFnArg, ActionMetadata, ActionType;
```
apascal07, line 48: Line 39 was not updated alongside this one. `genkit.dart` now exports `ToolResult`, `ToolResponseResult`, and `ToolInterruptResult`, but `plugin.dart` still exports only `Tool`, `ToolFn`, `ToolFnArgs`.

`genkit_middleware/lib/src/filesystem_middleware.dart` imports only `package:genkit/plugin.dart` and compiles because `.response(...)` resolves from the context type. A plugin author who writes an explicit `Future<ToolResult<String>>` signature, or `case ToolInterruptResult(:final data)` the way `mcp_server.dart:265` does, gets "Undefined class" and has to import `genkit.dart` or reach into `src/`.

packages/genkit/lib/src/genkit_class.dart
```diff
@@ -192,7 +192,65 @@ final class Genkit extends GenkitAI {
...
+  }) {
+    final tool = Tool<Input, Output>(
+      name: name,
+      description: description,
+      inputSchema: inputSchema,
+      toolOutputSchema: outputSchema,
+      metadata: {
+        ...?metadata,
+        'tool': {
+          ...?(metadata?['tool'] as Map<String, dynamic>?),
```
apascal07: Unchecked cast. `ai.defineInterrupt(name: 'x', description: 'y', metadata: {'tool': {}})` throws a `TypeError` at definition time, because the inner `{}` has context type `dynamic` and infers `Map<dynamic, dynamic>`. Same for any `Map.from(...)` value, or a non-Map such as `metadata: {'tool': 'x'}`.

`convert_tools.dart:58` uses `is Map` plus `.cast<String, dynamic>()` for this.

apascal07 (commented):
`ToolFn` changed shape, but no README was updated, so the tool samples on every affected pub.dev page no longer compile: `genkit/README.md:111` and `:190`, `genkit_google_genai/README.md:47`, `genkit_firebase_ai/README.md:52`, `genkit_anthropic/README.md:69`, `genkit_mcp/README.md:208`, `genkit_openai/README.md:101`. Line 190 also still teaches `context.interrupt()`, which this PR deprecates. The `example/example.dart` files were all migrated here because CI analyzes them; the READMEs have no gate.

pavelgj:
Triaged and addressed the review feedback. Summary of fixes (all in this branch):

**Correctness**
- `outputSchema` now surfaced for tools: added `Action.manifestOutputSchema` (overridden by `Tool` to return `toolOutputSchema`); reflection v1/v2 and MCP `tools/list` use it, so tool output schemas appear again in the Dev UI manifest and MCP.
- Multipart `parts` no longer silently dropped by non-Gemini providers: Anthropic maps image media into `tool_result` image content; OpenAI and Firebase AI log a warning when content is dropped (their tool-response shapes are text/structured-only).
- Interrupt turns preserve completed tools' `content` + `metadata`: `_ToolStatus` carries them, `_buildInterruptedResponse` stashes `pendingContent`/`pendingMetadata`, and `_resolveResume` restores them so resume matches the straight-through path.
- MCP `ToolResponse.content` encoding: `_toolResponseContent` now converts Genkit `Part` JSON into real MCP blocks (text/image/audio) and keeps the output text, instead of passing raw Part JSON through.
- Gemini `FunctionResponse.parts`: restricted to media parts mapped to `inlineData`/`fileData`; text/other parts are skipped rather than emitting invalid keys.
- Bidi interrupts: both the returned `.interrupt` and the deprecated throwing form now raise a clear `UNIMPLEMENTED` error instead of feeding the live model a bogus `{interrupt: true}`; metadata is forwarded for normal responses.
- MCP server `_callTool`: maps multipart content into MCP blocks, and surfaces interrupts with `isError: true` so a HITL pause is distinguishable from a real answer.
- `defineInterrupt` metadata merge: uses `is Map` + `.cast` instead of an unchecked cast, so `metadata: {'tool': {...}}` no longer throws.

**API / robustness**
- `GenkitPlugin.resolve` is now typed `resolve(ActionType, String)` (updated registry + all in-repo plugins), so matching against the wrong string is a compile error.
- `plugin.dart` now exports `ToolResult`, `ToolResponseResult`, `ToolInterruptResult`.
- Removed the unused public `Tool.toolFn` field (it bypassed validation/span/zone binding).

**Tests**: added coverage for multipart+metadata survival across interrupt/resume, `defineInterrupt` metadata cast, and MCP multipart/interrupt handling. `melos`-scope analyze is clean and the genkit + mcp + plugin suites pass.

apascal07:
Changes not pushed?

packages/genkit/lib/src/ai/tool.dart
pavelgj: Good catch! Done. Added `Action.manifestOutputSchema` (Tool returns `toolOutputSchema`); reflection v1/v2 and MCP `tools/list` read it now.

packages/genkit/lib/src/ai/tool.dart
pavelgj: Good catch! Done. Anthropic maps image media into `tool_result` content; OpenAI + Firebase AI log a warning when they cannot represent parts.

packages/genkit/lib/src/ai/tool.dart
pavelgj: Good catch! Done, removed the field.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Good catch! Done. Threaded through `_ToolStatus` / `_buildInterruptedResponse` (`pendingContent`/`pendingMetadata`) / `_resolveResume`, with a test covering resume.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Good catch! Done. MCP now converts the Part JSON into real content blocks (text/image/audio) instead of passing it through.

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Right, HITL is unary-only. Bidi now raises a clear UNIMPLEMENTED error for interrupt-capable tools instead of feeding the model `{interrupt: true}`.

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Good catch! Done, metadata is forwarded now.

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Good catch! Done. The throwing form now takes the same interrupt path as `.interrupt(...)`.

packages/genkit_google_genai/lib/src/common_plugin.dart
pavelgj: Good catch! Done. Restricted to media parts mapped to inlineData/fileData; other parts are skipped instead of emitting invalid keys.

packages/genkit_mcp/lib/src/server/mcp_server.dart
pavelgj: Good catch! Done. `_callTool` maps multipart content to MCP blocks and returns interrupts with `isError: true`. Added tests.

packages/genkit/lib/src/core/registry.dart
pavelgj: Good catch! Done. `GenkitPlugin.resolve` is now `resolve(ActionType, String)`; updated all in-repo plugins.

packages/genkit/lib/plugin.dart
pavelgj: Good catch! Done, exported `ToolResult`/`ToolResponseResult`/`ToolInterruptResult`.

packages/genkit/lib/src/genkit_class.dart
pavelgj: Good catch! Done, uses `is Map` + `.cast` now. Added a test.

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Confirmed unary-only. Bidi now rejects interrupt-capable tools with an UNIMPLEMENTED error rather than sending a fake tool answer to the live model.

pavelgj:
> Changes not pushed?

pushed now... got stuck.

pavelgj:
Good catch! Done. Migrated the tool samples in all six affected package READMEs (`genkit`, `genkit_google_genai`, `genkit_firebase_ai`, `genkit_anthropic`, `genkit_openai`, `genkit_mcp`) to return `.response(...)` / `.interrupt(...)`. In `genkit/README.md` I also updated the interrupt section to teach `return .interrupt(...)`. (The remaining `return ...` samples in the genkit/shelf READMEs are flows, not tools, so they are unaffected.)

packages/genkit/lib/src/ai/generate_bidi.dart
apascal07: The `UNIMPLEMENTED` throw does not escape. It is raised inside the `try`, and `GenkitException` is not a `ToolInterruptException`, so the new rethrow clause does not match it and the trailing `catch (e)` does:

```dart
case ToolInterruptResult():
  throw GenkitException(..., status: StatusCodes.UNIMPLEMENTED);   // inside try
...
} on ToolInterruptException {
  rethrow;
} catch (e) {
  toolResponses.add(... output: 'Error: $e');                      // catches it
}
```

So a returned `.interrupt()` still reaches the model, now as `output: 'Error: GenkitException: Tool "x" attempted to interrupt during a live (bidi) session...'`, and the session keeps streaming. `handleStream`'s outer catch never sees it.

Adding `on GenkitException { rethrow; }` next to the existing clause, or moving the throw out of the `try`, would make it propagate.

`generate_bidi_test.dart` still has no interrupt test, so neither path is covered.

packages/genkit/lib/src/ai/generate_bidi.dart
apascal07: Still inconsistent, in the opposite direction now. `ctx.interrupt(...)` throws `ToolInterruptException` and does propagate via the new rethrow clause. `.interrupt(...)` throws `GenkitException`, which that clause does not match, so it lands in the trailing `catch (e)` and becomes an `Error: ...` tool response. Detail on the thread above.

The form the migration guide tells users to move to is the one that is swallowed.

apascal07 (commented):
Comments inline.

packages/genkit/lib/src/ai/generate_bidi.dart
huangjeff5: The new `return .interrupt(...)` path still answers the live model. `ctx.interrupt()` (the deprecated throw) already does the right thing.

`ctx.interrupt()` throws `ToolInterruptException` inside `runRaw`. `on ToolInterruptException { rethrow; }` fires, `session.send` does not run, the session dies.

`return .interrupt(...)` does not throw from `runRaw`. It returns a `ToolInterruptResult`. The switch then throws `GenkitException`, which is not a `ToolInterruptException`, so `rethrow` never runs. `catch (e)` turns it into `output: 'Error: Tool "confirm" attempted to interrupt…'` and `session.send` still replies to the model.

If HITL is unary-only, I think we should take that throw out of the catch that invents a tool answer, so `.interrupt()` hits the same exit `ctx.interrupt()` already does: outer `addError`, no `session.send`.

```dart
late final ToolResult result;
try {
  result = (await tool.runRaw(toolRequest.toolRequest.input)).result;
} on ToolInterruptException {
  throw GenkitException(
    'Tool "$name" attempted to interrupt during a live (bidi) session. '
    'Interrupts are not supported by generateBidi; use unary generate().',
    status: StatusCodes.UNIMPLEMENTED,
  );
} catch (e) {
  toolResponses.add(
    ToolResponsePart(
      toolResponse: ToolResponse(
        ref: toolRequest.toolRequest.ref,
        name: [toolRequest.toolRequest.name](http://toolrequest.toolrequest.name/),
        output: 'Error: $e',
      ),
    ),
  );
  continue;
}

switch (result) {
  case ToolInterruptResult():
    throw GenkitException(
      'Tool "$name" attempted to interrupt during a live (bidi) session. '
      'Interrupts are not supported by generateBidi; use unary generate().',
      status: StatusCodes.UNIMPLEMENTED,
    );
  case ToolResponseResult(:final output, :final parts, :final metadata):
    toolResponses.add(/* existing response mapping */);
}
```

Unary `generate` is unchanged. Can we add a test that `return .interrupt(...)` and `ctx.interrupt(...)` both fail the session and the mock model never sees a tool message?

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Good catch! Fixed, the interrupt throw is now outside the error-catch, so both spellings propagate to the outer `addError` and never `session.send`. Added a bidi interrupt test.

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Good catch! Fixed with the thread above, `.interrupt(...)` and `ctx.interrupt(...)` now take the same exit. Covered by a test.

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Good catch! Done as suggested, moved the throw out of the error-catch so both forms fail the session with no tool answer. Added a test covering both.

apascal07 (approved):

Issue #366, opened 2026-08-24
test: shared mocked-HTTP fake-server harness for provider plugins


cabljac:
Coverage today: `genkit_openai`, `genkit_anthropic`, and `genkit_firebase_ai` have good converter unit tests but zero mocked-HTTP coverage of request pipelines. `genkit_vertexai` already has such a harness (`test/test_http_client.dart` shared mock, 26 tests across `vertex_test.dart`/`embedders_test.dart`/`known_models_test.dart` driving `model.run`, embed, and model-listing requests) which, via `CommonGoogleGenPlugin`, incidentally covers `genkit_google_genai`'s pipeline too.

Live tests: anthropic's `test_live.dart` has no key gate at all (fails with 6 errors when run directly rather than skipping). It escapes CI only because `test_live.dart` doesn't match Dart's default `test/**_test.dart` glob. Net effect: no live test in the repo ever exercises an API in CI - the three `test_live.dart` files miss the glob, and openai's `integration_test.dart` (which does match and run) self-skips without `OPENAI_API_KEY`.

Injection seams differ per plugin, which is the real design constraint for a shared harness:
- `genkit_openai`: `httpClient` param (`genkit_openai.dart:109`) - mock client works
- `genkit_vertexai`: `authClient` param - already used by its mocked tests
- `genkit_anthropic`: `baseUrl` only, no client injection - needs a real localhost fake server
- `genkit_google_genai`: neither `baseUrl` nor client override on `googleAI()` - needs a plugin API change first

(JS models this in `compat-oai/tests/fake_openai_server.ts` - a real localhost HTTP server, though local to compat-oai rather than shared.)

One shared harness unblocks pipeline tests everywhere; the wire-test copies added by #370/#374/#375 are the consolidation targets.

PR #373, opened 2026-08-24
fix(genkit_openai): populate usage metadata for streaming and non-streaming responses

### LINKED ISSUE #358: fix(genkit_openai): usage metadata never populated
Zero `usage` references in `lib/` - `ModelResponse.usage` is never set despite the SDK response carrying token counts. Streaming should also pass `stream_options.include_usage` (JS `model.ts:668-706`).

cabljac:
Fixes #358

`ModelResponse.usage` was never populated - token accounting blank for every user, both modes. Verified live: `response.usage: null` while the raw response carried `{prompt_tokens: 31, completion_tokens: 2, total_tokens: 33}`.

Fix: map the SDK usage block (prompt/completion/total + cached and reasoning details) into `GenerationUsage` on both response paths, and opt streaming requests in via `stream_options.include_usage` (JS parity). The SDK accumulator already carries final-chunk usage, so streaming needed only the opt-in and the mapping.

Red/green commits: four failing tests first (wire non-streaming incl. detail fields, streaming request body, streaming usage from the terminal SSE frame incl. a no-spurious-chunk pin, key-gated live), then the fix + converter unit tests.

Notes: `include_usage` is sent unconditionally, matching JS - strict compat endpoints that reject unknown stream options have no opt-out, same as JS. Wire harness duplicated from #370 pending #366; the two PRs merge in any order.

### DIFF
--- a/packages/genkit_openai/lib/src/converters.dart
+++ b/packages/genkit_openai/lib/src/converters.dart
@@ -263,4 +263,17 @@ abstract final class GenkitConverter {
       _ => FinishReason.unknown,
     };
   }
+
+  /// Map OpenAI token usage to Genkit [GenerationUsage].
+  static GenerationUsage? mapUsage(sdk.Usage? usage) {
+    if (usage == null) return null;
+    return GenerationUsage(
+      inputTokens: usage.promptTokens.toDouble(),
+      outputTokens: usage.completionTokens?.toDouble(),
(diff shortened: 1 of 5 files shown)


pavelgj (approved):

pavelgj (approved):
LGTM

PR #375, opened 2026-08-24
fix(genkit_google_genai): gate responseJsonSchema on constrained JSON mode

### LINKED ISSUE #364: fix(genkit_google_genai): responseJsonSchema sent unconditionally
Any request with an output schema attaches `responseJsonSchema` even in text mode, and non-JSON requests serialize `responseMimeType: ''` (`common_plugin.dart:274-277`). JS gates on `constrained && jsonMode` (`gemini.ts:854-860`).

cabljac:
## Problem

Any request with an output schema attaches `responseJsonSchema`, even in text mode, and unset `responseMimeType` serializes as `""` on the wire (`common_plugin.dart:274-277`). The TTS converter shares the unconditional schema attachment. JS gates on `output.constrained && jsonMode` (`js/plugins/google-genai/src/googleai/gemini.ts:854-860`).

## Fix

Attach `responseJsonSchema` only when `constrained && isJsonMode` (both converters); omit `responseMimeType` when unset. Not porting JS's `legacyResponseSchema`: migration-only, Dart pre-1.0 has no legacy users.

One deliberate divergence: a user-configured `responseMimeType` still passes through in non-JSON mode. JS clobbers it via spread order (`gemini.ts:839-841`), which reads as a bug, not intent. Pinned by a wire test so a parity sweep can't silently revert it.

`genkit_vertexai` inherits this code path (`VertexAiPluginImpl extends CommonGoogleGenPlugin`) - same narrowing, safe direction, worth a line in its release notes.

## Tests

Wire-level: injected MockClient captures the serialized `generationConfig` (pattern from #370/#374, consolidation in #366). Red commit first; pins cover both converters in both JSON-mode directions, mimeType passthrough, `contentType: application/json` selecting JSON mode, and a full `Genkit.generate(outputSchema:)` path exercising `applyFormat`. Each pin was verified by reverting the code it guards and confirming the test fails.

Fixes #364

## Release ordering

Safe for `lite.generate` only with #377 (merged, unreleased - published core is 0.15.1): before it, lite never configured formatters and this gate would drop schemas. Bump the min `genkit` constraint to the release shipping #377 before publishing this plugin.

### DIFF
--- a/packages/genkit_google_genai/lib/src/common_plugin.dart
+++ b/packages/genkit_google_genai/lib/src/common_plugin.dart
@@ -78,6 +78,7 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
             options,
             req.output?.schema,
             isJsonMode,
+            constrained: req.output?.constrained ?? false,
           );
           safetySettings = toGeminiSafetySettings(options.safetySettings);
           tools = toGeminiTools(
@@ -95,6 +96,7 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
             options,
             req.output?.schema,
             isJsonMode,
+            constrained: req.output?.constrained ?? false,
           );
           safetySettings = toGeminiSafetySettings(options.safetySettings);
           tools = toGeminiTools(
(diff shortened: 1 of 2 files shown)


pavelgj (approved):
LGTM

PR #396, opened 2026-08-26
fix(anthropic): handle model-specific thinking defaults

### LINKED ISSUE #354: fix(genkit_anthropic): thinking defaults to manual 'enabled' mode, rejected by Claude 4.7+
`_mapThinkingConfig` treats omitted `type` as `'enabled'` (`plugin_impl.dart:381`). Anthropic docs: manual extended thinking is deprecated on Claude 4.6 and returns 400 on 4.7+ (opus-4-7/4-8, sonnet-5, fable-5 - 4 of the 7 models in PR #322's catalog). A blanket flip to adaptive also breaks: adaptive 400s on sonnet-4-5/haiku-4-5, where enabled is the only mode.

Fix: resolve default thinking mode per model; expose `output_config.effort` for the adaptive path.

adesinah:
This PR handles model-specific thinking defaults. It resolve default thinking mode per model; expose output_config.effort for the adaptive path.

"Tracked" = present in the curated KnownClaudeModel enum (10 models). Lookup is by exact bare name, then by dated-snapshot alias — claude-haiku-4-5-20251001 strips to claude-haiku-4-5, so dated ids from the models endpoint hit the same entry.

Capability metadata — graceful fallback. modelInfoFor() returns curated ModelInfo (typed label, stable stage, tiered supports) for tracked names, and falls back to commonModelInfo for anything else. Untracked names always resolve; you can call any model string the API accepts, tracked or not. Tracking only enriches — it adds the label/stage and guarantees the model shows up in listings even when discovery fails or omits it.

Thinking mode — explicit failure. When thinking is supplied without a type, tracked models fill in their defaultThinkingMode (adaptive for 4.6-and-newer, enabled for 4.5). Untracked models throw INVALID_ARGUMENT asking for an explicit type rather than guessing. An explicit type always works for any model, tracked or not — the catalog is consulted only to fill the blank.

The asymmetry is intentional: a wrong capability guess is cheap and self-correcting, whereas a wrong thinking mode is a hard API rejection at request time, so that one refuses to guess.

Closes: #354

### DIFF
--- a/packages/genkit_anthropic/README.md
+++ b/packages/genkit_anthropic/README.md
@@ -78,18 +78,29 @@ final response = await ai.generate(
 print(response.text);
 ```
 
-### Thinking (Claude 3.7+)
+### Thinking
 
 ```dart
 final response = await ai.generate(
-  model: anthropic.model('claude-sonnet-4-5'),
+  model: anthropic.model('claude-sonnet-5'),
   prompt: 'Solve this 24 game: 2, 3, 10, 10',
-  config: AnthropicOptions(thinking: ThinkingConfig(budgetTokens: 2048)),
+  config: AnthropicOptions(
+    // Uses the model's compatible default thinking mode.
+    thinking: ThinkingConfig(),
+    outputConfig: AnthropicOutputConfig(effort: 'high'),
+  ),
 );
 
 // The thinking content is available in the message parts
 print(response.message?.content);
 ```
+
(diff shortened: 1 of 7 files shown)


cabljac (approved):
LGTM - maybe worth mentioning the breaking change about older models in PR body, no other comments from me.

pavelgj (approved):

PR #397, opened 2026-08-26
feat!: add cancellation support for action calls and generate API

pavelgj:
BREAKING CHANGE:
when using `maxTurns` option with the `generate` API, it will no longer throw a `GenkitException` with `ABORTED` status, but instead return an response with `aborted` finish reason. The response will contain the last turn's history, allowing you to continue execution if needed.

## Cancellation

```dart
import 'dart:async';
import 'package:genkit_google_genai/genkit_google_genai.dart';

final ai = Genkit(plugins: [googleAI()]);

// Caller owns the controller; the token is handed to generate.
final controller = CancellationController();

final stream = ai.generateStream(
  model: googleAI.gemini('gemini-flash-latest'),
  prompt: 'Write a long, detailed essay about the history of the internet.',
  cancel: controller.token,
);

controller.cancel('user pressed stop');

// Chunks emitted before the cancel still arrive; the stream then closes.
await for (final chunk in stream) {
  stdout.write(chunk.text);
}

// Rather than throwing, generate resolves with an aborted response.
final res = await stream.onResult;
if (res.finishReason == FinishReason.aborted) {
  print('\nCancelled: ${res.finishMessage}');
}
```

## Resuming a cancelled request

An aborted response carries no model message, but `res.messages` holds the last-good history (the request messages, plus any completed tool turns). Feed that straight back into a fresh `generate` with a new (uncancelled) token to pick up where you left off.

```dart
// First attempt, cancelled mid-flight.
final controller = CancellationController();
final aborted = await ai.generate(
  model: googleAI.gemini('gemini-flash-latest'),
  prompt: 'Explain quantum computing in detail.',
  cancel: controller.token,
);

controller.cancel();

// Resume: reuse the preserved history.
if (aborted.finishReason == FinishReason.aborted) {
  final resumed = await ai.generate(
    model: googleAI.gemini('gemini-flash-latest'),
    // The last-good conversation state is right here.
    messages: aborted.messages,
  );
  print(resumed.text);
}
```

Same idea after a tool-loop cancel: because the history preserves completed user/model/tool turns (and drops the unanswered model turn so it stays a clean resume point), passing `aborted.messages` back lets the model continue the tool loop from the last settled state.

### DIFF
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -47,9 +47,16 @@ import 'session.dart';
 /// ended (e.g. `interrupted`, `length`). When omitted, no per-turn reason is
 /// reported.
 class TurnResult {
-  TurnResult({this.finishReason});
+  TurnResult({this.finishReason, this.finishMessage});
 
   final AgentFinishReason? finishReason;
+
+  /// An optional human-readable message describing why the turn ended.
+  ///
+  /// Carries, for example, the "Reached max turns of N" text from a generation
+  /// that overran its turn limit, so the [SessionRunner] can surface it rather
+  /// than dropping it.
+  final String? finishMessage;
 }
 
 /// Per-turn context handed to the handler passed to [SessionRunner.run].
@@ -404,14 +411,50 @@ class SessionRunner<State> {
           if (cancel?.isCancelled ?? false) {
             lastTurnFinishReason = AgentFinishReason.aborted;
             lastTurnError = null;
+            // Persist the turn as `aborted`. The detached route already wrote
+            // `aborted` via `_abortSnapshotInStore`, in which case the
+            // abort-aware mutator makes this a no-op; but the attached route
+            // (`AgentTurn.abort()` -> token cancel) has written nothing, so
+            // without this the trailing `invocationEnd` snapshot would persist a
+            // half-finished turn as `completed` and later be picked as a resume
+            // point.
+            final snapshotId = await maybeSnapshot(
+              status: 'aborted',
+              snapshotId: turnSnapshotId,
+              finishReason: AgentFinishReason.aborted,
+            );
             _notifyEndTurn(
-              _lastSnapshot?.snapshotId,
+              snapshotId ?? _lastSnapshot?.snapshotId,
               AgentFinishReason.aborted,
             );
             return true;
           }
 
           final finishReason = turnResult?.finishReason;
+
+          // A turn that resolved `aborted` *without* the token being cancelled
+          // is not a cooperative cancel: it is an overrun (e.g. the generate
+          // loop hit `maxTurns`). Route it to the failure path so the reason
+          // (e.g. "Reached max turns of N") surfaces as an error instead of
+          // being silently dropped as a success with a null message.
+          if (finishReason == AgentFinishReason.aborted) {
+            lastTurnFinishReason = AgentFinishReason.failed;
+            lastTurnError = toErrorDetails(
+              GenkitException(
+                turnResult?.finishMessage ?? 'Turn aborted.',
+                status: StatusCodes.ABORTED,
+              ),
+            );
+            final snapshotId = await maybeSnapshot(
+              status: 'failed',
+              error: lastTurnError,
+              snapshotId: turnSnapshotId,
+              finishReason: AgentFinishReason.failed,
+            );
+            _notifyEndTurn(snapshotId, AgentFinishReason.failed);
+            return true;
+          }
+
           lastTurnFinishReason = finishReason;
           lastTurnError = null;
 
@@ -898,7 +941,7 @@ final class _InProcessTransport extends AgentTransport {
   TurnStream runTurn(
     AgentInput input,
     AgentInit init, {
-    required CancellationToken cancel,
+    CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) {
     final bidi = _startBidi(input, init, context: context, cancel: cancel);
@@ -909,7 +952,7 @@ final class _InProcessTransport extends AgentTransport {
   Future<AgentOutput>? run(
     AgentInput input,
     AgentInit init, {
-    required CancellationToken cancel,
+    CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) {
     return _startBidi(input, init, context: context, cancel: cancel).onResult;
@@ -1137,7 +1180,9 @@ Agent<State> defineCustomAgent<State>(
           // Capture the disposer so a reused, long-lived `ctx.cancel` token
           // doesn't accumulate one stranded listener (pinning this turn's
           // controller and closure) per turn. Disposed when the flow settles.
-          final unlinkCancel = ctx.cancel.onCancel(cancelController.cancel);
+          // `link` forwards the cancellation reason so a caller-supplied
+          // `controller.cancel('...')` survives the hop into this turn's token.
+          final unlinkCancel = ctx.cancel?.link(cancelController);
           final cancelToken = cancelController.token;
           void Function()? unsubscribe;
           // Background heartbeat timer for the detached snapshot. Started in
@@ -1302,7 +1347,7 @@ Agent<State> defineCustomAgent<State>(
               // The turn has settled (the snapshot reached a terminal status),
               // so stop refreshing its heartbeat.
               stopHeartbeat();
-              unlinkCancel();
+              unlinkCancel?.call();
               unsubscribe?.call();
               offArtifactAdded();
               offArtifactUpdated();
@@ -1597,10 +1642,16 @@ Agent<State> definePromptAgent<State>(
         context: options.context,
         inputStream: null,
         init: null,
-        cancel: options.cancel ?? CancellationToken.none,
+        cancel: options.cancel,
       ));
 
-      // Keep everything that is NOT a prompt-template message.
+      final aborted = res.finishReason == FinishReason.aborted;
+
+      // Keep everything that is NOT a prompt-template message. On an aborted
+      // turn `res.message` is null; the abort path already carries the last-good
+      // history in `res.modelRequest`, so leaving `keep` without a trailing
+      // model message preserves the user turn as the resume point rather than
+      // treating the user's own message as the model reply.
       final reqMessages = res.modelRequest?.messages;
       if (reqMessages != null) {
         final keep = reqMessages
@@ -1628,13 +1679,18 @@ Agent<State> definePromptAgent<State>(
       final reason = res.finishReason;
       return TurnResult(
         finishReason: reason != null ? AgentFinishReason(reason.value) : null,
+        finishMessage: aborted ? res.finishMessage : null,
       );
     });
 
     final msgs = sess.getMessages();
+    // Only surface the last message as the agent's reply when it is actually a
+    // model turn. An aborted turn leaves the user message last (see above), so
+    // returning it would echo the user's prompt back as the agent's reply.
+    final lastMessage = msgs.isNotEmpty ? msgs.last : null;
     return AgentResult(
       artifacts: sess.getArtifacts(),
-      message: msgs.isNotEmpty ? msgs.last : null,
+      message: lastMessage?.role == Role.model ? lastMessage : null,
       finishReason: sess.lastTurnFinishReason,
     );
   }
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -1134,7 +1134,10 @@ Agent<State> defineCustomAgent<State>(
           // transport token to it, so an attached `runTurn(cancel:)` also
           // cooperatively stops this turn's `generate`.
           final cancelController = CancellationController();
-          ctx.cancel.onCancel(cancelController.cancel);
+          // Capture the disposer so a reused, long-lived `ctx.cancel` token
+          // doesn't accumulate one stranded listener (pinning this turn's
+          // controller and closure) per turn. Disposed when the flow settles.
+          final unlinkCancel = ctx.cancel.onCancel(cancelController.cancel);
           final cancelToken = cancelController.token;
           void Function()? unsubscribe;
           // Background heartbeat timer for the detached snapshot. Started in
@@ -1299,6 +1302,7 @@ Agent<State> defineCustomAgent<State>(
               // The turn has settled (the snapshot reached a terminal status),
               // so stop refreshing its heartbeat.
               stopHeartbeat();
+              unlinkCancel();
               unsubscribe?.call();
               offArtifactAdded();
               offArtifactUpdated();
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -390,8 +390,26 @@ class SessionRunner<State> {
       );
 
       try {
-        await runInNewSpan('runTurn-${turnIndex + 1}', (_) async {
+        final aborted = await runInNewSpan('runTurn-${turnIndex + 1}', (
+          _,
+        ) async {
           final turnResult = await fn(input, turnContext);
+
+          // The generate loop now resolves (rather than throws) on a
+          // cooperative cancel, so a returned turn can still be an abort. Mirror
+          // the catch-path handling: record `aborted`, skip the `completed`
+          // snapshot (the abort path already persisted `aborted`, and the
+          // abort-aware mutator would skip a late `completed` write anyway), and
+          // stop processing further inputs.
+          if (cancel?.isCancelled ?? false) {
+            lastTurnFinishReason = AgentFinishReason.aborted;
+            lastTurnError = null;
+            _notifyEndTurn(
+              _lastSnapshot?.snapshotId,
+              AgentFinishReason.aborted,
+            );
+            return true;
+          }
 
           final finishReason = turnResult?.finishReason;
           lastTurnFinishReason = finishReason;
@@ -408,8 +426,9 @@ class SessionRunner<State> {
           _lastGoodFinishReason = finishReason;
 
           _notifyEndTurn(snapshotId, finishReason);
-          return 0;
+          return false;
         }, input: input);
+        if (aborted) break;
         turnIndex++;
       } catch (e) {
         // An aborted turn rejects out of `generate` and lands here. Treat it as
@@ -857,23 +876,24 @@ final class _InProcessTransport extends AgentTransport {
 
   BidiActionStream<AgentStreamChunk, AgentOutput, AgentInput> _startBidi(
     AgentInput input,
-    AgentInit init, [
+    AgentInit init, {
     Map<String, dynamic>? context,
-  ]) {
-    final bidi = primaryAction.streamBidi(init: init, context: context);
+    CancellationToken? cancel,
+  }) {
+    final bidi = primaryAction.streamBidi(
+      init: init,
+      context: context,
+      cancel: cancel,
+    );
     bidi.send(input);
     bidi.close();
     return bidi;
   }
 
-  // NOTE: [cancel] is accepted to satisfy the [AgentTransport] contract but is
-  // not threaded into `generate` on the in-process transport today, so it does
-  // not stop an in-flight model call for an *attached* turn. Cooperative
-  // cancellation currently only takes effect on the detached path: `abort`
-  // flips the persisted snapshot to `aborted`, and a detached worker observing
-  // that (via `SnapshotChangeNotifier`) cancels its own turn. Aborting an
-  // attached in-process turn is therefore effectively persist-only. Threading
-  // cancellation through `generate` to cancel attached turns is future work.
+  // [cancel] is threaded into the agent action's `generate` call (via
+  // `streamBidi`), so aborting an attached in-process turn cooperatively stops
+  // the in-flight model call. The detached path additionally observes the
+  // persisted `aborted` status via `SnapshotChangeNotifier`.
   @override
   TurnStream runTurn(
     AgentInput input,
@@ -881,12 +901,10 @@ final class _InProcessTransport extends AgentTransport {
     required CancellationToken cancel,
     Map<String, dynamic>? context,
   }) {
-    final bidi = _startBidi(input, init, context);
+    final bidi = _startBidi(input, init, context: context, cancel: cancel);
     return (stream: bidi, output: bidi.onResult);
   }
 
-  // See [runTurn]: [cancel] is not wired into `generate` here, so aborting an
-  // attached in-process turn is persist-only (does not stop the model).
   @override
   Future<AgentOutput>? run(
     AgentInput input,
@@ -894,7 +912,7 @@ final class _InProcessTransport extends AgentTransport {
     required CancellationToken cancel,
     Map<String, dynamic>? context,
   }) {
-    return _startBidi(input, init, context).onResult;
+    return _startBidi(input, init, context: context, cancel: cancel).onResult;
   }
 
   @override
@@ -1111,7 +1129,13 @@ Agent<State> defineCustomAgent<State>(
 
           String? detachedSnapshotId;
           final detachCompleter = Completer<void>();
-          final cancelToken = CancellationToken();
+          // Own a controller for this turn (cancelled on detach-abort or when
+          // the persisted snapshot flips to `aborted`) and link the ambient
+          // transport token to it, so an attached `runTurn(cancel:)` also
+          // cooperatively stops this turn's `generate`.
+          final cancelController = CancellationController();
+          ctx.cancel.onCancel(cancelController.cancel);
+          final cancelToken = cancelController.token;
           void Function()? unsubscribe;
           // Background heartbeat timer for the detached snapshot. Started in
           // `onDetach`, cleared when the flow settles (or on abort).
@@ -1190,7 +1214,7 @@ Agent<State> defineCustomAgent<State>(
                     .onSnapshotStateChange(snapshotId, (snap) {
                       if (snap.status?.value == 'aborted') {
                         stopHeartbeat();
-                        cancelToken.cancel();
+                        cancelController.cancel();
                         localUnsubscribe?.call();
                       }
                     }, context: ctx.context);
@@ -1569,6 +1593,7 @@ Agent<State> definePromptAgent<State>(
         context: options.context,
         inputStream: null,
         init: null,
+        cancel: options.cancel ?? CancellationToken.none,
       ));
 
       // Keep everything that is NOT a prompt-template message.
--- a/packages/genkit/lib/src/ai/agents/agent_core.dart
+++ b/packages/genkit/lib/src/ai/agents/agent_core.dart
@@ -78,7 +78,7 @@ abstract base class AgentTransport {
   TurnStream runTurn(
     AgentInput input,
     AgentInit init, {
-    required CancellationToken cancel,
+    CancellationToken? cancel,
     Map<String, dynamic>? context,
   });
 
@@ -90,7 +90,7 @@ abstract base class AgentTransport {
   Future<AgentOutput>? run(
     AgentInput input,
     AgentInit init, {
-    required CancellationToken cancel,
+    CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) => null;
 
@@ -678,14 +678,14 @@ final class AgentChat<State> {
   /// turns resolve to a synthetic `aborted` response.
   Future<AgentResponse<State>> _buildResponse(
     Future<AgentOutput> output,
-    CancellationToken cancel,
+    CancellationToken? cancel,
     int messageCountBeforeTurn,
   ) async {
     AgentOutput raw;
     try {
       raw = await output;
     } catch (e) {
-      if (cancel.isCancelled) {
+      if (cancel?.isCancelled ?? false) {
         raw = AgentOutput(finishReason: AgentFinishReason.aborted);
       } else {
         // A thrown transport error / non-200 rejects before reaching the
@@ -772,13 +772,13 @@ final class AgentChat<State> {
     Map<String, dynamic>? context,
   }) async {
     // In `_send` the token is observed and forwarded only (never cancelled
-    // here), so a caller token flows straight through; absent one, the
-    // never-cancelled sentinel is fine.
-    final token = cancel ?? CancellationToken.none;
+    // here), so a caller token flows straight through; absent one (`null`),
+    // there is no cancellation to observe.
+    final token = cancel;
     // Bail before pushing the message or dispatching the turn if the caller's
     // token is already cancelled: there's no point starting work, and we must
     // not leave an orphaned user message in `messages`.
-    if (token.isCancelled) {
+    if (token?.isCancelled ?? false) {
       return _abortedResponse();
     }
     final runFuture = _transport.run(
@@ -859,8 +859,10 @@ final class AgentChat<State> {
   }) {
     // Own a controller for this turn so `AgentTurn.abort()` can cancel it, and
     // link it to the caller's token (if any) so an external cancel also aborts.
+    // `link` forwards the cancellation reason so a caller-supplied
+    // `controller.cancel('...')` survives the hop into this turn's token.
     final controller = CancellationController();
-    final unlink = cancel?.onCancel(controller.cancel);
+    final unlink = cancel?.link(controller);
     final token = controller.token;
     // Bail before pushing the message or dispatching the turn if the caller's
     // token is already cancelled: return an empty stream and a synthetic
@@ -901,8 +903,11 @@ final class AgentChat<State> {
       (_) => AgentResponse<State>._(AgentOutput(), messages),
     );
     // Detach the caller-token listener once the turn settles so a reused
-    // caller token does not accumulate one handler per turn.
-    if (unlink != null) responsePromise.whenComplete(unlink);
+    // caller token does not accumulate one handler per turn. `whenComplete`
+    // returns a *new* future that re-completes with the same error; `ignore()`
+    // it so a rejecting turn does not reach `Zone.handleUncaughtError` (the
+    // `catchError` above handles the original future, not this derived one).
+    if (unlink != null) responsePromise.whenComplete(unlink).ignore();
 
     Stream<AgentChunk<State>> buildStream() async* {
       var previousText = '';
--- a/packages/genkit/lib/src/ai/agents/agent_core.dart
+++ b/packages/genkit/lib/src/ai/agents/agent_core.dart
@@ -32,55 +32,14 @@ import 'dart:async';
 import 'package:meta/meta.dart';
 import 'package:schemantic/schemantic.dart';
 
+import '../../core/cancellation.dart';
 import '../../schema_extensions.dart';
 import '../../types.dart';
 import 'json_patch.dart';
 import 'state_codec.dart';
 
-// ---------------------------------------------------------------------------
-// Cancellation (Dart has no AbortSignal/AbortController).
-// ---------------------------------------------------------------------------
-
-/// A minimal cooperative cancellation primitive, the Dart stand-in for the
-/// Web's `AbortSignal`/`AbortController`.
-final class CancellationToken {
-  final Completer<void> _completer = Completer<void>();
-  final List<void Function()> _listeners = [];
-
-  /// Whether cancellation has been requested.
-  bool get isCancelled => _completer.isCompleted;
-
-  /// Completes when [cancel] is called.
-  Future<void> get whenCancelled => _completer.future;
-
-  /// Registers [callback] to run when this token is cancelled, and returns a
-  /// disposer that unregisters it. Unlike [whenCancelled] (a one-shot future
-  /// that can never be detached), this lets a caller-supplied token be reused
-  /// across turns without leaking per-turn handlers. If the token is already
-  /// cancelled, [callback] runs synchronously and the returned disposer is a
-  /// no-op.
-  void Function() onCancel(void Function() callback) {
-    if (_completer.isCompleted) {
-      callback();
-      return () {};
-    }
-    _listeners.add(callback);
-    return () => _listeners.remove(callback);
-  }
-
-  /// Requests cancellation (idempotent).
-  void cancel() {
-    if (_completer.isCompleted) return;
-    _completer.complete();
-    // Snapshot then clear so listeners that (re)register during fan-out don't
-    // fire twice and can't be stranded in the list.
-    final listeners = [..._listeners];
-    _listeners.clear();
-    for (final listener in listeners) {
-      listener();
-    }
-  }
-}
+export '../../core/cancellation.dart'
+    show CancellationController, CancellationToken;
 
 // ---------------------------------------------------------------------------
 // Transport
@@ -812,7 +771,10 @@ final class AgentChat<State> {
     CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) async {
-    final token = cancel ?? CancellationToken();
+    // In `_send` the token is observed and forwarded only (never cancelled
+    // here), so a caller token flows straight through; absent one, the
+    // never-cancelled sentinel is fine.
+    final token = cancel ?? CancellationToken.none;
     // Bail before pushing the message or dispatching the turn if the caller's
     // token is already cancelled: there's no point starting work, and we must
     // not leave an orphaned user message in `messages`.
@@ -895,12 +857,17 @@ final class AgentChat<State> {
     CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) {
-    final token = cancel ?? CancellationToken();
+    // Own a controller for this turn so `AgentTurn.abort()` can cancel it, and
+    // link it to the caller's token (if any) so an external cancel also aborts.
+    final controller = CancellationController();
+    final unlink = cancel?.onCancel(controller.cancel);
+    final token = controller.token;
     // Bail before pushing the message or dispatching the turn if the caller's
     // token is already cancelled: return an empty stream and a synthetic
     // `aborted` response, and leave `messages` untouched. Mirrors the JS
     // core's pre-aborted short-circuit.
     if (token.isCancelled) {
+      unlink?.call();
       return AgentTurn<State>._(
         stream: const Stream.empty(),
         response: Future.value(_abortedResponse()),
@@ -933,6 +900,9 @@ final class AgentChat<State> {
     responsePromise.catchError(
       (_) => AgentResponse<State>._(AgentOutput(), messages),
     );
+    // Detach the caller-token listener once the turn settles so a reused
+    // caller token does not accumulate one handler per turn.
+    if (unlink != null) responsePromise.whenComplete(unlink);
 
     Stream<AgentChunk<State>> buildStream() async* {
       var previousText = '';
@@ -977,7 +947,7 @@ final class AgentChat<State> {
     return AgentTurn<State>._(
       stream: buildStream(),
       response: responsePromise,
-      onAbort: token.cancel,
+      onAbort: controller.cancel,
     );
   }
 
@@ -1066,11 +1036,11 @@ final class AgentChat<State> {
     }
     final init = _buildInit();
 
-    final token = CancellationToken();
+    final controller = CancellationController();
     final turn = _transport.runTurn(
       agentInput,
       init,
-      cancel: token,
+      cancel: controller.token,
       context: context,
     );
     final raw = await turn.output;
--- a/packages/genkit/lib/src/ai/generate.dart
+++ b/packages/genkit/lib/src/ai/generate.dart
@@ -1148,6 +1148,13 @@ _executeTools(
       // response (which would let the loop continue as if the tool "failed").
       rethrow;
     } catch (e) {
+      if (cancelToken.isCancelled) {
+        // A generic failure (e.g. a closed HTTP client throwing
+        // SocketException) raised during cancellation must still propagate as a
+        // cancellation rather than be recorded as a tool error that lets the
+        // loop continue.
+        throw CancelledException(reason: cancelToken.reason);
+      }
       toolResponses.add(
         ToolResponsePart(
           toolResponse: ToolResponse(
--- a/packages/genkit/lib/src/ai/generate.dart
+++ b/packages/genkit/lib/src/ai/generate.dart
@@ -236,16 +236,68 @@ GenerateResponseHelper _abortedResponse({
   Map<String, dynamic>? config,
   Object? reason,
 }) {
+  final request = ModelRequest(messages: history, config: config);
   return GenerateResponseHelper(
     ModelResponse(
       finishReason: FinishReason.aborted,
       finishMessage: reason is String
           ? reason
           : (reason?.toString() ?? 'Generation was cancelled'),
+      // Stamp the request onto the ModelResponse too, not just the helper, so
+      // the resumable history survives the reflection boundary (the registered
+      // `generate` util action returns `response.modelResponse`, dropping the
+      // helper's own `_request`).
+      request: GenerateRequest(messages: history, config: config),
     ),
-    request: ModelRequest(messages: history, config: config),
+    request: request,
     output: null,
   );
+}
+
+/// Decides whether an exception [e] raised during a generation turn should be
+/// converted into an aborted response, or rethrown.
+///
+/// Returns a [GenerateResponseHelper] (the abort) when:
+/// - [e] is a [CancelledException] produced by *this* turn's [cancel] token
+///   (matched by identity, or by the token being cancelled), or
+/// - [cancel] is cancelled and [e] is a generic failure surfaced because the
+///   plugin honored cancellation by tearing down its transport (e.g. a
+///   `SocketException` from a closed HTTP client). In that case the original
+///   error is preserved as the finish reason so a genuine failure that merely
+///   raced the cancel is not silently masked.
+///
+/// Returns `null` when the caller should rethrow: a [CancelledException] from an
+/// unrelated token (e.g. a tool's own internal timeout) is a real failure, not
+/// an abort of this generation.
+///
+/// All abort sites pass the same [history] shape (the turn's accumulated,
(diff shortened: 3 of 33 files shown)


packages/genkit/lib/src/core/cancellation.dart
```diff
@@ -0,0 +1,135 @@
...
+  /// leaking per-operation handlers. If the token is already cancelled,
+  /// [callback] runs synchronously and the returned disposer is a no-op.
+  void Function() onCancel(void Function() callback) {
+    if (_completer.isCompleted) {
+      callback();
+      return () {};
+    }
+    _listeners.add(callback);
+    return () => _listeners.remove(callback);
+  }
```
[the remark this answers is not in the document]
pavelgj: Done. `onCancel` now returns a no-op disposer immediately when called on `CancellationToken.none`, so its listener list can no longer accumulate callbacks. Since `none` is the pervasive default, this also covers most of the reuse cases the other two comments flag.

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -1074,7 +1092,13 @@ Agent<State> defineCustomAgent<State>(
...
           String? detachedSnapshotId;
           final detachCompleter = Completer<void>();
-          final cancelToken = CancellationToken();
+          // Own a controller for this turn (cancelled on detach-abort or when
+          // the persisted snapshot flips to `aborted`) and link the ambient
+          // transport token to it, so an attached `runTurn(cancel:)` also
+          // cooperatively stops this turn's `generate`.
+          final cancelController = CancellationController();
+          ctx.cancel.onCancel(cancelController.cancel);
+          final cancelToken = cancelController.token;
```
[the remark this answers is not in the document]
pavelgj: Done. Captured the disposer as `unlinkCancel` and call it in the `flow()` `finally` block alongside the other teardown (`unsubscribe?.call()`, `offArtifactAdded()`, etc.), so a reused long-lived token no longer strands a listener per turn.

packages/genkit/lib/src/ai/generate_bidi.dart
```diff
@@ -107,7 +109,11 @@ Future<GenerateBidiSession> runGenerateBidi(
     tools: toolDefs,
   );
 
-  final session = model.streamBidi(init: initRequest);
+  final session = model.streamBidi(init: initRequest, cancel: cancel);
+  // Close the input side of the session when cancellation is requested so no
+  // further turns can be sent; the model's own `cancel` handling stops the
+  // in-flight turn.
+  cancel?.onCancel(() => unawaited(session.close()));
```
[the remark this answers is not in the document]
pavelgj: Done, applied as suggested. The disposer is captured and disposed via `session.onResult.whenComplete(unsubscribe)` so the closure (and the session it pins) is released once the session settles.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -1031,6 +1128,10 @@ _executeTools(
       interrupted = true;
       toolStatus[toolRequest.toolRequest.ref ?? toolRequest.toolRequest.name] =
           (output: null, content: null, metadata: null, interrupt: e);
+    } on CancelledException {
+      // Cancellation must propagate, not be swallowed into an error tool
+      // response (which would let the loop continue as if the tool "failed").
+      rethrow;
     } catch (e) {
```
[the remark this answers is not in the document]
pavelgj: Done. The generic `catch (e)` block in `_executeTools` now checks `cancelToken.isCancelled` first and rethrows as a `CancelledException(reason: cancelToken.reason)`, so a transport-level failure (e.g. a `SocketException` from a closed HTTP client during cancellation) propagates as an abort instead of being recorded as a spurious error tool response. Added a regression test ("a generic tool error during cancellation aborts instead of being swallowed as a tool failure") that asserts the loop stops and the response is aborted.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -327,29 +371,50 @@ Future<GenerateResponseHelper> _runGenerateLoop(
...
+        reason: ctx.cancel.reason,
       );
-    },
-    context: ctx.context,
-    inputStream: null,
-    init: null,
-  ));
+    }
+    rethrow;
+  }
```
apascal07, line 417: There is no cancel checkpoint on the path where the model returns successfully, so for any plugin that does not itself observe `ctx.cancel`, a cancelled call resolves as `finishReason: stop`.

`ai.generate(model: openAI.model('gpt-4o'), prompt: '...', cancel: t)` with `t.cancel()` mid-flight: genkit_openai never reads `ctx.cancel`, so the request runs to completion and bills in full. `composedModel` returns normally, so this catch is skipped, `toolRequests` is empty, and line 437 returns the full response. The turn-entry checkpoint at line 264 only fires on a *subsequent* turn, which never happens on a single-turn call.

README.md:275 says "Rather than throwing, a cancelled call resolves with a response whose `finishReason` is `FinishReason.aborted`". That holds only for genkit_google_genai. A grep for `ctx.cancel` outside `packages/genkit` matches nothing else, so genkit_anthropic, genkit_openai, genkit_chrome, genkit_firebase_ai and `remote_model` all silently return `stop`.

A post-model `if (ctx.cancel.isCancelled) return _abortedResponse(...)` before line 436 would make the documented contract hold for every plugin instead of per-plugin opt-in.

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -385,9 +385,27 @@ class SessionRunner<State> {
...
+        ) async {
           final turnResult = await fn(input, turnContext);
 
+          // The generate loop now resolves (rather than throws) on a
+          // cooperative cancel, so a returned turn can still be an abort. Mirror
+          // the catch-path handling: record `aborted`, skip the `completed`
+          // snapshot (the abort path already persisted `aborted`, and the
+          // abort-aware mutator would skip a late `completed` write anyway), and
+          // stop processing further inputs.
+          if (cancel?.isCancelled ?? false) {
```
apascal07, line 399: The comment above says the abort path already persisted `aborted`, so skipping the snapshot is safe. That holds for the detached route only. This PR newly makes it reachable from the attached route, where nothing persists anything.

`AgentTurn.abort()` is `onAbort: controller.cancel` (agent_core.dart:930). It flips a token and writes nothing. The chain now connects end to end: `controller.cancel` -> `_startBidi(cancel:)` -> `ctx.cancel` -> line 1103 -> `SessionRunner.cancel`. So this branch fires and breaks with no write.

`_abortSnapshotInStore` is the only writer of `aborted`, reached solely from `Agent.abort(snapshotId)`, and the store-to-token listener at line 1178 lives inside `onDetach`. So `_abortAwareMutator` (line 147, which vetoes only when the *stored* status is already `aborted`) sees null and lets the next write through. `flow()` then calls `await runner.maybeSnapshot()` at line 1261 with no status, and line 548 writes `SnapshotStatus(status ?? 'completed')`.

The version guard at line 515 does not save it, because `session.addMessages([inputMessage])` at line 361 already bumped the version.

Net: a half-finished aborted turn is persisted as `completed`, and line 708 says only `completed` snapshots are resumable, so a later `loadChat` picks it as the last-good resume point.

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -385,9 +385,27 @@ class SessionRunner<State> {
...
+            lastTurnFinishReason = AgentFinishReason.aborted;
+            lastTurnError = null;
+            _notifyEndTurn(
+              _lastSnapshot?.snapshotId,
+              AgentFinishReason.aborted,
+            );
+            return true;
+          }
+
           final finishReason = turnResult?.finishReason;
```
apascal07, line 409: `maxTurns` now returns `FinishReason.aborted` instead of throwing, but the guard at line 399 tests `cancel?.isCancelled`, which is false for an overrun. So a runaway tool loop reaches this line and takes the success path.

Result: `maybeSnapshot(status: 'completed', finishReason: aborted)` at line 413, `lastTurnError = null`, and `lastGoodState`/`turnIndex` both advance. The finalizer's failure branch at line 1318 requires `lastTurnFinishReason == failed && lastTurnError != null`, so it is skipped and `AgentOutput` carries no error. `AgentChat._buildResponse` only throws on `failed` (agent_core.dart:695), so it returns normally, and `AgentResponse.finishMessage` reads `_raw.error?.message`, which is null. The "Reached max turns of N" string is dropped entirely.

Before this PR the throw landed in the catch below and produced `AgentFinishReason.failed`, a `failed` snapshot with error details, and an `AgentError` to the caller. testapps/agents/lib/coding_agent.dart (`maxTurns: 30`) and orchestrator_agent.dart (`maxTurns: 10`) both hit this.

`AgentFinishReason` already carries `aborted` and `failed` separately (types.dart:445-447), so the agent layer can distinguish what `FinishReason` now conflates. The guard could test `turnResult?.finishReason` as well as the token.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -525,6 +611,7 @@ Future<GenerateResponseHelper> _runGenerateAction(
         generateRegistry,
```
apascal07, line 611: This `_executeTools` call got `cancel:` threaded in, but unlike the two sites inside `_runGenerateLoop` (lines 407 and 451) it is not wrapped in the abort-to-response guard. So a cancelled restart escapes `generate()` as a throw.

`ai.generate(model: m, prompt: 'go', interruptRestart: [part], cancel: (CancellationController()..cancel()).token)`: `coreGenerate` enters `if (resumeRestart.isNotEmpty)` at line 609, strictly before `_runGenerateLoop`'s entry checkpoint at line 264 ever runs. `coreTool`'s `throwIfCancelled` at line 1074 fires, line 1131 deliberately rethrows, and nothing upstream converts it (`_runGenerateAction`, `coreGenerate`, `runInNewSpan`, `generateHelper` and `GenkitAI.generate` all lack a catch).

That throws `CancelledException` where the README, testapps/ai_features/lib/cancellation_sample.dart, and this PR's own `generate_cancel_test.dart` ("already-cancelled token returns an aborted response") all say it resolves with `finishReason: aborted`. Also reachable from `lite.dart`, which exposes both `cancel:` and `interruptRestart:`.

packages/genkit/lib/src/ai/generate_bidi.dart
```diff
@@ -164,6 +175,7 @@ Future<GenerateBidiSession> runGenerateBidi(
             try {
               result = (await tool.runRaw(
                 toolRequest.toolRequest.input,
+                cancel: cancel,
```
apascal07, line 178: The token is threaded into `tool.runRaw` here, but the bidi loop did not get the `on CancelledException { rethrow; }` guard that generate.dart:1131 added for the same reason. The generic `catch (e)` at line 183 turns a cancel into a fabricated tool answer, and the loop then hits a closed sink.

Cancel during a pending bidi tool call: the listener at line 120 runs synchronously (`_cancel` fans out inline) and `session.close()` evaluates `_inputSink?.close()` before its first suspension, so the controller is already closed. `tool.runRaw` then throws `CancelledException` from `Action.run`'s `throwIfCancelled` (action.dart:263). It is not a `ToolInterruptException`, so line 183 writes `output: 'Error: GenkitException: Operation was cancelled...'` back to the model and continues. The loop reaches `session.send(...)` at line 218, and `BidiActionStream.send` (action.dart:468-473) has no `isClosed` guard, so `_inputSink.add` throws `StateError: Cannot add event after closing`, which the outer catch pushes onto `outputController`.

The consumer sees a `StateError`, never a clean cancellation, after the model was fed a bogus tool result. generate_bidi_test.dart contains zero occurrences of `cancel`, so none of this path is covered.

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -1524,6 +1552,7 @@ Agent<State> definePromptAgent<State>(
         context: options.context,
         inputStream: null,
         init: null,
+        cancel: options.cancel ?? CancellationToken.none,
```
apascal07, line 1555: When this generate resolves aborted, `res.message` is null, so `if (res.message != null) keep.add(res.message!)` at line 1564 does not run and `sess.setMessages(keep)` leaves the *user's* message last. Line 1592 then takes `msgs.last` as the agent's message, so `AgentResponse.text` returns the user's own prompt back as the agent's reply.

The non-null message also defeats the orphan rollback. agent_core.dart:686-692 removes the eagerly-pushed user message only when `raw.message == null`, so on a store-backed agent the user message is now duplicated in `messages` on the next turn.

Both follow from the same line: `message: msgs.isNotEmpty ? msgs.last : null` needs to exclude a non-model last message, or be guarded on `res.finishReason == FinishReason.aborted`.

packages/genkit/lib/src/ai/agents/agent_core.dart
```diff
@@ -913,6 +880,9 @@ final class AgentChat<State> {
     responsePromise.catchError(
       (_) => AgentResponse<State>._(AgentOutput(), messages),
     );
+    // Detach the caller-token listener once the turn settles so a reused
+    // caller token does not accumulate one handler per turn.
+    if (unlink != null) responsePromise.whenComplete(unlink);
```
apascal07, line 885: `whenComplete` returns a *new* future that re-completes with the same error. It is discarded here, so nothing listens on it and a rejecting turn reaches `Zone.handleUncaughtError`.

The `catchError` on line 880 does not cover this. Handled-ness is per derived future, not per error object.

`unlink` is effectively always non-null: `_send` sets `token = cancel ?? CancellationToken.none` (line 757) and passes it on (line 778), and `none.onCancel` returns a non-null no-op disposer (cancellation.dart:75).

Trigger: a client does `chat.send(text: 'hi')` against a remote agent (`RemoteAgentTransport.run` returns null at remote_agent.dart:233, so `_send` falls through to `_sendStream`) and the server returns `finishReason: 'failed'`. `_buildResponse` throws `AgentError` at line 695. The caller's try/catch handles `send()`'s rejection correctly, but the orphan future holds the same error with zero listeners. Console app: unhandled exception and non-zero exit. package:test: the enclosing test fails after passing. Flutter: `PlatformDispatcher.onError`.

Not caught by CI because both existing failed-turn tests (agent_core_test.dart:421 and :440) use `supportsRun: true` and take the `run()` fast path.

`.ignore()` on the derived future, or folding `unlink` into the existing `then`/`catchError` pair, avoids it.

packages/genkit/lib/src/ai/generate_bidi.dart
```diff
@@ -109,7 +111,16 @@ Future<GenerateBidiSession> runGenerateBidi(
...
-  final session = model.streamBidi(init: initRequest);
+  final session = model.streamBidi(init: initRequest, cancel: cancel);
+  // Close the input side of the session when cancellation is requested so no
+  // further turns can be sent; the model's own `cancel` handling stops the
+  // in-flight turn. Capture the disposer and drop it once the session settles
+  // so a reused, long-lived `cancel` token doesn't leak this closure (and the
+  // session it pins) across sessions.
+  final unsubscribe = cancel?.onCancel(() => unawaited(session.close()));
+  if (unsubscribe != null) {
+    unawaited(session.onResult.whenComplete(unsubscribe));
```
apascal07, line 122: `unawaited` is an empty-bodied no-op. It silences the analyzer lint and attaches no error handler, so the future returned by `whenComplete` carries the error with no listener and reaches `Zone.handleUncaughtError`.

Reading `session.onResult` here also makes this newly reachable. `ActionStream.onResult` lazily creates `_completer` (action.dart:411-414), and `setError`'s guard is `if (_completer?.isCompleted == false)` (action.dart:447-454). With a null completer that is `null == false`, so before this line existed the branch never fired for bidi, because nothing in `runGenerateBidi` touched `onResult`.

So on any bidi model failure (transport, auth) with a `cancel` token supplied, `completeError` now fires, its only listener is the non-error-handling `whenComplete`, and the isolate takes an unhandled async error. The caller still gets the same error through `outputController`, so this is a pure duplicate that crashes. Passing `CancellationToken.none` triggers it too, since `onCancel` returns a non-null disposer.

Separately, `onResult` completes only when the session closes, so `unsubscribe` never runs for a session that is abandoned without `close()`, leaving the closure (and the session it pins) on a long-lived caller token.

packages/genkit/lib/src/ai/generate_types.dart
```diff
@@ -90,25 +90,33 @@ final class GenerateResponseHelper<Output> extends GenerateResponse {
...
   /// the final model response.
   ///
   /// This is useful for continuing the conversation in multi-turn scenarios.
+  /// When the response has no message (e.g. an aborted turn), only the request
+  /// history is returned, so callers can attempt to resume from the last good
+  /// state.
   List<Message> get messages => [
     ...(_request?.messages ?? _response.request?.messages ?? []),
-    _response.message!,
+    if (_response.message != null) _response.message!,
```
apascal07, line 119: `messages` is guarded for a null message here, but `jsonOutput` (line 155) is not, and it throws on an aborted response.

`jsonOutput` falls back to `extractJson(text)` when `output` is null. An aborted response has no message, so `text` is `''`. In extract.dart, no `{` or `[` is found (line 39), `jsonDecode('')` throws and is swallowed (lines 40-42), `allowPartial` is false so the repair branch is skipped, and `primitivePattern` at line 49 matches neither a quote nor a digit in `''`. Line 60 then throws `FormatException('No JSON object or array found')`.

So `final res = await ai.generate(prompt: '...', outputSchema: FooSchema, cancel: t)` followed by a read of `res.jsonOutput` crashes on the abort path this PR documents as a normal outcome. Every other accessor degrades safely (`text` gives `''`, `media`/`toolRequests`/`interrupts` give null or `[]`); `jsonOutput` is the one that does not.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -327,29 +371,50 @@ Future<GenerateResponseHelper> _runGenerateLoop(
...
+          ),
+        );
+      },
+      context: ctx.context,
+      inputStream: null,
+      init: null,
+      cancel: ctx.cancel,
+    ));
+  } catch (e) {
+    if (ctx.cancel.isCancelled) {
```
apascal07, line 409: This discriminates on "is the token cancelled right now" rather than on the exception's identity, and `e` is bound then dropped entirely. Any genuine failure that lands after a cancel is reported as a clean abort with no trace of the cause.

A provider 503 (or a `RetryMiddleware` rethrow after exhausting attempts, or a `TypeError` in format parsing) surfaces at t=5.000s and the user presses stop at t=5.001s. `isCancelled` is true, so this returns `finishReason: aborted` with `finishMessage: 'Generation was cancelled'` (`ctx.cancel.reason` is null for a no-arg `cancel()`). The 503 vanishes, and a retry wrapper sees a successful return and does not retry.

Deterministic variant: a hallucinated tool name makes `_executeTools` throw `GenkitException('Tool X not found', NOT_FOUND)` from line 1063, which sits outside the per-request try. If the token is cancelled by then, that is reported as `aborted` too.

The window is the whole model call plus all tool execution. Matching on `e is CancelledException`, or carrying `e` through as the aborted response's `underlyingException`, keeps the diagnostic. Same shape at line 462.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -1031,7 +1128,18 @@ _executeTools(
       interrupted = true;
       toolStatus[toolRequest.toolRequest.ref ?? toolRequest.toolRequest.name] =
           (output: null, content: null, metadata: null, interrupt: e);
+    } on CancelledException {
```
apascal07, line 1146: This rethrows any `CancelledException` without checking which token produced it, while the caller only converts when `ctx.cancel` is cancelled. So a tool's own internal token escapes `generate()` as a throw.

A tool implementing its own timeout:

```dart
genkit.defineTool(name: 'search', description: '...', fn: (input, ctx) async {
  final timeout = CancellationController();
  Timer(const Duration(seconds: 5), timeout.cancel);
  await doWork();
  timeout.token.throwIfCancelled();
  return .response(result);
});
```

Caller invokes `ai.generate(model: m, prompt: 'go', toolNames: ['search'])` with no `cancel:` at all, so `ctx.cancel` is `CancellationToken.none`. The timeout fires, this rethrows, line 462 evaluates `ctx.cancel.isCancelled` as false, and line 469 rethrows. `generate()` throws even though the caller never asked for cancellation, where any other tool failure would have become an error `ToolResponsePart` and let the loop continue.

Gating the rethrow on `cancelToken.isCancelled`, or giving `CancelledException` a token reference and matching on identity, keeps it scoped to the caller's own cancel.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -977,9 +1071,11 @@ _executeTools(
       ActionFnArg<void, dynamic, void> c,
     ) async {
       _recordResumedMetadata(c.context);
+      c.cancel.throwIfCancelled();
```
apascal07, line 1089: This checkpoint and `Action.run`'s (action.dart:262) are the only two a tool gets, and both fire *before* the tool body runs. Nothing fires during it, because `ToolFnArgs` exposes no way to reach the token.

`ToolFnArgs` holds it in a library-private `_base` (tool.dart:26) and exposes only `context`, `toolRequest`, `resumed` and `interrupt`. A grep for `cancel` in tool.dart returns zero hits, and tool.dart:215 drops it at `fn(input as Input, ToolFnArgs(ctx))`. There is no ambient fallback either: the only zone keys in packages/genkit/lib are `_genkitContextKey`, `#api.context`, `ToolRequestPart` and `_sessionZoneKey`, and the token is not in the context map.

So a tool doing a 5-minute fetch cannot see a cancel issued at t=1s, and the caller's `await` does not resolve to aborted until the tool finishes on its own. genkit_ai.dart:128 documents the parameter as "observed by the model call, tools, and middleware", which is not true for tools today.

This also blocks the fix for genkit_middleware/agents_middleware.dart:307, where sub-agent delegation cannot forward a token it has no way to read.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -1031,7 +1128,18 @@ _executeTools(
       interrupted = true;
       toolStatus[toolRequest.toolRequest.ref ?? toolRequest.toolRequest.name] =
           (output: null, content: null, metadata: null, interrupt: e);
+    } on CancelledException {
+      // Cancellation must propagate, not be swallowed into an error tool
+      // response (which would let the loop continue as if the tool "failed").
+      rethrow;
```
apascal07, line 1149: This guard is defeated by genkit_middleware/lib/src/filesystem_middleware.dart:390, where a catch-all in the `tool` hook wraps `next(request, ctx)` and converts every exception into a tool failure.

An agent using `filesystemMiddleware` is cancelled while a filesystem tool is pending. `tool.runRaw` throws `CancelledException` from `Action.run`, it propagates up through `next`, and the middleware's `catch (e)` turns it into an error `ToolResponsePart`. `_executeTools` never sees the exception, so this rethrow never fires, and the loop continues exactly as the comment says it must not.

The turn-entry checkpoint at line 264 aborts on the next turn, but the fabricated "cancelled" tool result is by then part of `currentRequest.messages` and lands in the resumable history the README tells users to feed back into a fresh `generate`.

Mirroring this guard (`on CancelledException { rethrow; }`) above that catch-all fixes it. Worth checking the other genkit_middleware hooks for the same shape.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -635,6 +722,10 @@ Future<GenerateResponseHelper> generateHelper<CustomOptions>(
   StreamingCallback<GenerateResponseChunk>? onChunk,
   List<GenerateMiddlewareOneof>? middleware,
 
+  /// Cooperative cancellation token, observed by the model call, tools, and
+  /// middleware to abort generation.
+  CancellationToken? cancel,
```
apascal07, line 728: prompt.dart:327 is the one `generateHelper` call site that does not pass `cancel:`, so the whole prompt surface silently ignores the feature.

`PromptGenerateOptions` (prompt.dart:126-152) has no `cancel` field, and `ExecutablePrompt.call`/`.stream` (prompt.dart:252-261) take only `(Input? input, [PromptGenerateOptions? opts])`. prompt.dart contains the string `cancel` zero times. So `final p = await ai.prompt('summarize'); await p(input);` cannot be cancelled at all: there is no parameter to pass a token to, and `_generate` would drop it anyway. It defaults to `CancellationToken.none`, which makes every checkpoint in `_runGenerateLoop`, `coreModel` and `_executeTools` dead for that path.

The inconsistency is visible inside this PR: `definePromptAgent` threads it correctly at agent.dart:1555, so an agent built on a prompt is cancellable while the prompt itself is not.

testapps/ai_features/lib/cancellation_sample.dart
```diff
@@ -0,0 +1,76 @@
...
+  final preCancelled = CancellationController()..cancel();
+  final res2 = await ai.generate(
+    model: googleAI.gemini('gemini-flash-latest'),
+    prompt: 'This should never reach the model.',
+    cancel: preCancelled.token,
+  );
+  print(
+    'Pre-cancelled generate finished as ${res2.finishReason?.value} '
+    'with ${res2.messages.length} message(s) of history: ${jsonEncode(res2.messages).toString()}'
+    ,
```
apascal07, line 72: This bare comma on its own line is not `dart format` output, so the Format CI job fails on this PR.

`.github/workflows/format.yml:38-39` runs `dart format --output=none --set-exit-if-changed .` from the repo root with no `working-directory`, there is no `.pubignore` anywhere, and `testapps/ai_features` is a workspace member, so it is in scope. `dart format` attaches a trailing comma to the end of the preceding argument and never emits it alone.

Line 58 is a second hit at 109 columns against an 80-column page width.

While fixing: `jsonEncode(...).toString()` on lines 58 and 71 is redundant twice over. `jsonEncode` already returns a `String`, and it is already inside a `${}` interpolation.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -292,10 +331,15 @@ Future<GenerateResponseHelper> _runGenerateLoop(
     ModelRequest req,
     ActionFnArg<ModelResponseChunk, ModelRequest, void> c,
   ) {
+    // Cooperative checkpoint right before the (potentially expensive) model
+    // call. Middleware wrapping `model` runs before this and can observe
+    // `c.cancel` itself.
+    c.cancel.throwIfCancelled();
```
apascal07, line 337: This checkpoint is what eventually stops a cancelled call, but `RetryMiddleware` sits above it and sleeps through the cancel before reaching it.

`_retry` (packages/genkit/lib/src/ai/middleware/retry.dart:184-205) takes a bare `Future<T> Function()` thunk. It never receives or inspects `ctx.cancel`, and `await Future.delayed(delay)` at line 202 has no cancellation race.

With genkit_google_genai the interaction is worse than a plain sleep. Cancel closes the HTTP client (common_plugin.dart:115), so the pending request or stream fails with `http.ClientException`, which is not a `GenkitException`. `handleException` (common_plugin.dart:249-254) maps it to `StatusCodes.INTERNAL`, which is in `defaultRetryStatuses` (retry.dart:131-137), so `_shouldRetry` returns true and the loop backs off.

No extra billable call results, because the retried attempt hits this line first and `CANCELLED` is not retryable. The cost is latency: `_calculateDelay` caps before jitter, so a cancel landing on attempt 3 sleeps up to 6s with defaults, and up to `maxDelayMs * 1.5` (90s) with a raised `maxRetries`. `await ai.generate(...)` stays pending that whole time after `cancel()` returns.

Racing the delay against `ctx.cancel.whenCancelled` fixes it.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -234,19 +258,34 @@ Future<GenerateResponseHelper> _runGenerateLoop(
   int currentTurn = 0,
   int messageIndex = 0,
 }) async {
+  // Cooperative checkpoint at the start of every turn so a cancel between turns
+  // resolves with an aborted response carrying the last-good history (rather
+  // than throwing), letting the caller resume from `response.messages`.
+  if (ctx.cancel.isCancelled) {
+    return _abortedResponse(
+      history: options.messages,
```
apascal07, line 266: The four `_abortedResponse` call sites pass two different histories, so the resumable state the README tells users to feed back depends on *when* the cancel fired.

This site and the `maxTurns` branch at line 284 pass `options.messages`. The two catches at lines 411 and 464 pass `currentRequest.messages`, which is post-`applyFormat` (line 303) and post-`_resolveResume` (line 368).

Concretely, `ai.generate(prompt: 'x', outputSchema: FooSchema, cancel: t)`: cancelled before the call, `res.messages` is `[user('x')]`; cancelled during the model call, it is `[user('x' + TextPart(<JSON format instructions>, metadata: {'purpose': 'output'}))]`. Feed the second into a fresh `generate` with no `outputSchema` and the model still sees the stale JSON instruction.

Duplicate injection is not a risk, since `injectInstructions` short-circuits on `messages.any(hasOutputInstruction)` (formatters.dart:169). The corollary is the real one: that guard also *suppresses* injection for a different schema on resume.

There is a second divergence on the same lines. A cancel at this checkpoint yields interrupt-unresolved history, while a cancel one step later yields the `_resolveResume`-rewritten version with the extra `Role.tool` message. Picking one definition of "resumable history" and using it at all four sites would remove both.

packages/genkit/lib/src/core/cancellation.dart
```diff
@@ -0,0 +1,140 @@
...
+
+  /// Whether cancellation has been requested.
+  bool get isCancelled => _completer.isCompleted;
+
+  /// The optional reason passed to [CancellationController.cancel], or `null`.
+  Object? get reason => _reason;
+
+  /// Completes when the owning controller is cancelled. For
+  /// [CancellationToken.none] this future never completes.
+  Future<void> get whenCancelled => _completer.future;
```
apascal07, line 63: `onCancel` guards the shared singleton with `identical(this, none)` at line 75. This getter does not, so it hands back a future that never completes, and every awaiter is retained for the life of the process.

`none` is a `static final` whose `_completer` is never completed, and it is the default `ctx.cancel` for every action (action.dart:261). `await ctx.cancel.whenCancelled` or `Future.any([work, ctx.cancel.whenCancelled])` appends one `_FutureListener` to that future per invocation, holding the continuation and everything the closure captures. Nothing completes it, so nothing is released. At any real request rate that grows until OOM.

The doc at lines 37-40 points callers straight at this ("await [whenCancelled]"), and the doc two lines above says outright that on `none` the future never completes, so the hazard is already known here. No `lib/` caller passes `none` into `whenCancelled` today (remote_agent.dart:184 always gets a fresh controller's token), so it is latent rather than live.

Guarding this the way `onCancel` is guarded, or asserting on `none` so the pattern fails loudly, closes it.

packages/genkit/lib/src/core/cancellation.dart
```diff
@@ -0,0 +1,140 @@
...
+
+  void _cancel(Object? reason) {
+    if (_completer.isCompleted) return;
+    _reason = reason;
+    _completer.complete();
+    // Snapshot then clear so listeners that (re)register during fan-out don't
+    // fire twice and can't be stranded in the list.
+    final listeners = [..._listeners];
+    _listeners.clear();
+    for (final listener in listeners) {
```
apascal07, line 104: One throwing listener stops the rest of the fan-out, and because `_listeners.clear()` already ran on line 103 the survivors are dropped permanently rather than left registered for a retry.

The exception also escapes `CancellationController.cancel()` synchronously, so a caller's stop path takes a throw from a call documented at line 124 as "Requests cancellation (idempotent)".

Cancellation is chained (agent_core.dart:843, agent.dart:1103), so each hop runs the next token's whole fan-out inline inside this unprotected loop. The chain today is app token -> `cancelController.cancel` -> that token's `_cancel` -> the plugin's `service.client.close`. A partial teardown is invisible, because line 99 completes the future before the loop runs: `isCancelled` and `whenCancelled` both report a clean cancel while the HTTP client was never closed.

No listener in the tree throws synchronously today, so this is latent. A try/catch per listener routing to the logger or `Zone.current.handleUncaughtError` contains it.

packages/genkit/lib/src/ai/agents/agent_core.dart
```diff
@@ -875,12 +837,17 @@ final class AgentChat<State> {
     CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) {
-    final token = cancel ?? CancellationToken();
+    // Own a controller for this turn so `AgentTurn.abort()` can cancel it, and
+    // link it to the caller's token (if any) so an external cancel also aborts.
+    final controller = CancellationController();
+    final unlink = cancel?.onCancel(controller.cancel);
```
apascal07, line 843: `onCancel` takes `void Function()` while `CancellationController.cancel` is `void Function([Object? reason])`. Dart accepts the tear-off and invokes it with zero arguments, so the child token's `_reason` is always null. Same at agent.dart:1103.

A `defineCustomAgent` body calling `ai.generate(..., cancel: options.cancel)` therefore reads `finishMessage: 'Generation was cancelled'` where the app passed `controller.cancel('user pressed stop')`. A direct `ai.generate(cancel: token)` preserves it, because generateHelper passes the caller's token straight through (generate.dart:837) with no relink. So the same reason string survives or vanishes depending on call path.

Low severity, but the fix is small: `(reason) => controller.cancel(reason)` with an `onCancel` that passes the reason through, or an explicit `link`/`linkedController` helper. That helper would also cover the three places this PR hand-rolls the same capture-disposer-dispose-on-settle pattern (here, agent.dart:1103, generate_bidi.dart:120).

packages/genkit/test/ai/generate_cancel_test.dart
```diff
@@ -0,0 +1,335 @@
...
+        prompt: 'hello',
+        cancel: controller.token,
+      );
+
+      final res = await stream.onResult;
+      expect(res.finishReason, FinishReason.aborted);
+      expect(res.message, isNull);
+    });
+
+    test('cancel aborts between tool-loop turns, preserving history', () async {
```
apascal07, line 185: This test does not cover the path its name describes. The cancel fires *inside* the second model call (lines 194-195), so it lands in the mid-model catch at generate.dart:409, not the turn-entry checkpoint at generate.dart:264. Those two return different histories (`currentRequest.messages` vs `options.messages`), so the between-turns path is still untested.

Other gaps in the new surface:

- generate_test.dart kept both `maxTurns` cases but dropped the `StatusCodes.ABORTED` assertion with no replacement, and neither case asserts `message`/`output` are null or that the history is the full accumulated loop.
- generate_bidi_test.dart contains zero occurrences of `cancel`, so the whole bidi cancel path (the tool loop and the session-close listener) is uncovered.
- Nothing covers cancel on the resume/restart path, which is the one that throws today.
- Nothing covers an attached `AgentTurn.abort()` against a store, which is where the snapshot lands as `completed`.
- agent_core_test.dart's two failed-turn tests both use `supportsRun: true`, so they take the `run()` fast path and never reach `_sendStream`, where the unhandled-rejection issue lives.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -234,19 +258,34 @@ Future<GenerateResponseHelper> _runGenerateLoop(
...
+  // Check turn limits. Treat exceeding the limit like a cooperative abort:
+  // resolve with an aborted response carrying the history so far, so the caller
+  // can inspect/resume rather than catching an exception.
   final maxTurns = options.maxTurns ?? _defaultMaxTurns;
   if (currentTurn >= maxTurns) {
-    throw GenkitException(
-      'Reached max turns of $maxTurns. Adjust maxTurns option to increase the max number of turns.',
-      status: StatusCodes.ABORTED,
+    return _abortedResponse(
+      history: options.messages,
```
apascal07, line 284: Four in-repo callers branch on `finishReason` and handle only `interrupted` and `failed`, so they degrade silently now that an overrun returns `aborted` here.

- genkit_middleware/lib/src/agents_middleware.dart:340 returns `'(no response)'` where a sub-agent overrun used to surface as `"Error calling agent 'x': Reached max turns"`.
- testapps/agentic_patterns/lib/autonomous_operation.dart:94 loops on `finishReason == interrupted`. `aborted` exits the loop and line 131 returns `response.text`, which is `''` for a message-less response, so the flow reports success with an empty answer.
- testapps/middleware/bin/coding_agent.dart:107 breaks on anything that is not `interrupted`, then prints an empty `AI Response:`. Its `catch` no longer fires.
- The registered `generate` util action returns `response.modelResponse` (generate.dart:67), and `_abortedResponse` puts the history on the helper's `_request`, never on the `ModelResponse`. So over reflection a Dev UI run that overruns `maxTurns` gets HTTP 200 with `{"finishReason":"aborted","finishMessage":"Reached max turns..."}`, no message, no candidates and no history. Stamping `ModelResponse.request` on the abort path would carry the resume state across that boundary.

A sweep for `finishReason ==` and `switch` on `FinishReason` would catch the rest.

packages/genkit/lib/src/core/action.dart
```diff
@@ -203,12 +209,16 @@ class Action<Input, Output, Chunk, Init>
     Stream<Input>? inputStream,
     Init? init,
     TraceStartCallback? onTraceStart,
+    CancellationToken? cancel,
   }) async {
     return (await run(
       input,
       onChunk: onChunk,
       context: context,
+      inputStream: inputStream,
```
apascal07, line 218: Forwarding `init` here skips the schema validation `runRaw` applies. `runRaw` takes `dynamic init` and runs `initSchema!.parse(init)` when both are non-null (lines 249-256); `call` takes a statically typed `Init?` and passes it straight through.

The static type covers the shape, so what is skipped is every value-level constraint the schema carries: enum membership, numeric ranges, required nested fields. `action.call(input, init: SomeInit(mode: 'nope'))` reaches the action body unvalidated, while the same value through `runRaw` is rejected.

Nothing in-repo passes `init` to `call`, and before this change `call` discarded it entirely, so there is no regression. It is a new gap on a newly live path. Routing both through the same validation, or documenting `call` as the pre-validated entry point, closes it.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Intentional — this is the documented "best-effort" contract. If the model call completes before the cancel takes effect, we return the completed result rather than discarding finished (already-billed) work and relabeling it `aborted`. A post-model `if (isCancelled) return _abortedResponse()` would throw away a valid response the user paid for.

This is now spelled out in the README ("Cancellation is cooperative and best-effort… If the model call happens to complete before the cancellation takes effect, its result is returned normally — the completed work is not discarded… whether a mid-flight cancel yields `aborted` or a completed result depends on the plugin") and covered by the test "a completed model call is returned even if the token was cancelled while it ran". Plugins that want prompt aborts tear down their transport on cancel (e.g. `genkit_google_genai`); others run to completion. Leaving as-is by design.

packages/genkit/lib/src/ai/agents/agent.dart
pavelgj: Done. The `SessionRunner.run` loop now detects `cancel?.isCancelled` and persists the turn as `aborted` via `maybeSnapshot(status: 'aborted', …)` before breaking, so the attached route (`AgentTurn.abort()` → token cancel, which writes nothing itself) no longer lets the trailing `invocationEnd` write persist a half-finished turn as `completed`. On the detached route that already wrote `aborted`, the abort-aware mutator makes this a no-op.

Added a regression test "an attached AgentTurn.abort() persists the turn as aborted, not completed" (agent_test.dart) that holds a turn in flight, calls `AgentTurn.abort()`, and asserts the persisted snapshot is `aborted`.

packages/genkit/lib/src/ai/agents/agent.dart
pavelgj: Done. The run loop now distinguishes an overrun from a cooperative cancel: a turn that resolves `aborted` *without* the token being cancelled is routed to the failure path — `lastTurnError` is set from `turnResult?.finishMessage` and the snapshot is written `failed` — so the "Reached max turns of N" text surfaces as an error instead of being dropped as a null-message success. A genuine cooperative cancel still records `aborted`.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. `coreGenerate`'s `resumeRestart.isNotEmpty` branch now wraps `_executeTools` in the same abort-to-response guard (`_abortResponseIfCancelled`) as the two sites inside `_runGenerateLoop`, so an already-cancelled restart resolves with `finishReason: aborted` instead of throwing. Added the test "an already-cancelled token on the resume/restart path returns an aborted response instead of throwing".

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Done. The bidi tool loop now has an `on CancelledException { rethrow; }` guard above the generic `catch (e)`, so a cancel propagates instead of becoming an `Error: …cancelled` tool answer. `BidiActionStream` also now tracks `isClosed` and `send()` no-ops once the input sink is closed, so a send that races the cancel-triggered `close()` no longer throws `StateError`. Added the test "cancel during a pending tool call does not fabricate a tool answer or crash the session" (generate_bidi_test.dart).

packages/genkit/lib/src/ai/agents/agent.dart
pavelgj: Done. `definePromptAgent` now guards the returned reply with `lastMessage?.role == Role.model ? lastMessage : null`, so an aborted turn (where `res.message` is null and the user message is left last) no longer echoes the user's prompt back as the agent's reply, and the orphan-rollback in `agent_core` still fires (no duplicate user message on the next turn).

packages/genkit/lib/src/ai/agents/agent_core.dart
pavelgj: Done. The derived future is now suppressed with `.ignore()`, so a rejecting turn no longer reaches `Zone.handleUncaughtError` while the caller still sees the error through the awaited `response`. Added a streaming-path regression test "a failed turn on the streaming path throws AgentError without an unhandled rejection" (agent_core_test.dart), which uses `supportsRun: false` so it actually exercises `_sendStream` (the two existing failed-turn tests take the `run()` fast path).

packages/genkit/lib/src/ai/generate_bidi.dart
pavelgj: Done. Switched to `session.onResult.whenComplete(unsubscribe).ignore()` so a bidi session that settles with an error doesn't surface a duplicate unhandled async error via this cleanup hook (the caller already sees it through `outputController`).

packages/genkit/lib/src/ai/generate_types.dart
pavelgj: Done. `jsonOutput` now degrades to `null` on a message-less / empty-text response, matching the other accessors (`text` → `''`, `media`/`toolRequests`/`interrupts` → null/`[]`). Covered by the test "jsonOutput returns null on an aborted response rather than throwing".

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. `_abortResponseIfCancelled` now preserves the original cause: when the token is cancelled but the exception is a generic failure, the aborted response carries it (`reason: cancel.reason ?? e`) rather than discarding it as a clean "Generation was cancelled". Applied at both catch sites. Added the test "a genuine failure that races a cancel is not masked as a clean abort", which asserts the underlying cause (e.g. a 503) survives in the finish message.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. `CancelledException` now carries a `token` reference, and `_executeTools` rethrows only when the exception belongs to the caller's token (`identical(e.token, cancelToken) || cancelToken.isCancelled`); an unrelated tool-internal timeout token is recorded as an error tool response and the loop continues. Added the test "a tool's own internal cancellation does not escape generate when the caller never asked to cancel".

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. `ToolFnArgs` now exposes a public `cancel` getter (`CancellationToken? get cancel => _base.cancel`) with docs showing how to race `cancel.whenCancelled` mid-body, so a long-running tool can observe cancellation during execution. Added the test "a tool can observe the cancellation token via ctx.cancel".

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. The `filesystem_middleware` `tool` hook now has `on CancelledException { rethrow; }` above the catch-all that converts tool errors into failure responses, so a cancellation propagates rather than being recorded as a fabricated "cancelled" tool result in the resumable history. (Swept the other middleware hooks for the same shape.)

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. `PromptGenerateOptions` now has a `cancel` field, and `ExecutablePrompt.call` / `.stream` thread it through `_generate` down to `generateHelper`, so the prompt surface is cancellable and consistent with `definePromptAgent`.

testapps/ai_features/lib/cancellation_sample.dart
pavelgj: Done. `cancellation_sample.dart` is now `dart format --output=none --set-exit-if-changed` clean (no bare-comma-on-its-own-line, no over-width lines), and the redundant `jsonEncode(...).toString()` calls were reduced to plain `jsonEncode(...)` inside the interpolations.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. `_retry` now receives the cancel token, skips a retry when it's already cancelled, and races the backoff delay against `cancel.whenCancelled`, so a cancel landing during backoff resolves promptly instead of waiting out the full (possibly capped-to-`maxDelayMs`) delay.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. All four abort sites now pass the same pre-format history (`options.messages` / `opts.messages`), so the resumable state a caller feeds back no longer depends on when the cancel fired — no stale JSON format-injection instructions on resume, and no divergence around the `_resolveResume`-rewritten tool message.

packages/genkit/lib/src/core/cancellation.dart
pavelgj: Resolved by removing the hazard entirely: `CancellationToken.none` is gone and the ambient token is now nullable (`CancellationToken?`) throughout (action, tool, generate, agents, plugins). There's no never-completing singleton to await, so `whenCancelled` is a plain getter again and this leak class is closed. (This is a breaking-change follow-up applied across the surface; observers use null-aware calls like `ctx.cancel?.throwIfCancelled()`.)

packages/genkit/lib/src/core/cancellation.dart
pavelgj: Done. The fan-out now wraps each listener in a try/catch that routes a throwing listener to the zone error handler, so one bad listener can't drop the surviving listeners or make the documented-idempotent `cancel()` throw synchronously. Covered by the test "a throwing listener does not abort the fan-out or escape cancel".

packages/genkit/lib/src/ai/agents/agent_core.dart
pavelgj: Done. Added a reason-forwarding `link(CancellationController child)` helper (plus `onCancelWithReason`), and the three hand-rolled capture-dispose-on-settle sites (agent.dart, agent_core.dart, generate_bidi.dart) now use it, so a caller's `cancel('user pressed stop')` reason survives the hop into the per-turn token. Covered by "link forwards cancellation and reason to the child" and "link disposer unlinks the child".

packages/genkit/test/ai/generate_cancel_test.dart
pavelgj: Done. Filled the coverage gaps:

- **cancel between tool-loop turns** (turn-entry checkpoint, distinct from mid-model): "cancel aborts between tool-loop turns, preserving history".
- **resume/restart path with an already-cancelled token**: "an already-cancelled token on the resume/restart path returns an aborted response instead of throwing".
- **maxTurns overrun**: now asserts `finishReason: aborted` + the "max turns" `finishMessage` and non-empty history.
- **generic tool error during cancellation**: "a generic tool error during cancellation aborts instead of being swallowed as a tool failure".
- **bidi cancel**: "cancel during a pending tool call does not fabricate a tool answer or crash the session" (tool-loop cancel + closed-sink).
- **attached `AgentTurn.abort()` against a store**: "an attached AgentTurn.abort() persists the turn as aborted, not completed" (agent_test.dart).
- **streaming failed-turn (unhandled rejection)**: "a failed turn on the streaming path throws AgentError without an unhandled rejection" — uses `supportsRun: false` so it actually reaches `_sendStream`.

The specific test this comment was left on now cancels mid-model deliberately for that catch site; the between-turns checkpoint is exercised by the separate between-turns test above.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Done. `agents_middleware` and `autonomous_operation` now handle `aborted` explicitly, and `coding_agent` surfaces the abort `finishMessage` instead of printing an empty response. `_abortedResponse` also stamps `ModelResponse.request` so the resumable history survives the reflection boundary (the registered `generate` util returns `response.modelResponse`), so a Dev UI run that overruns `maxTurns` carries the history + finishMessage rather than an empty 200.

packages/genkit/lib/src/core/action.dart
pavelgj: Done. `Action.call` now routes `init` through the same `initSchema!.parse(init)` validation `runRaw` applies (skipping only when either the schema or the value is absent), so value-level constraints — enum membership, numeric ranges, required nested fields — are enforced on the `call` path too.

apascal07 (approved):

PR #399, opened 2026-08-27
feat: add Interrupt tool and export tools in Lite API

pavelgj:
Introduce the `Interrupt` tool class for human-in-the-loop and out-of-band processing patterns that interrupt the generation loop.

Export `Interrupt`, `Tool`, and related tool types across `genkit.dart`, `lite.dart`, and `plugin.dart` so tools can be passed directly to `generate`/`generateStream`, including via the lightweight Lite API.

Update the README to demonstrate defining and using a tool with the Lite API.

### DIFF
--- a/packages/genkit/test/lite_test.dart
+++ b/packages/genkit/test/lite_test.dart
@@ -214,6 +214,232 @@ void main() {
     expect(response.text, '{"result": "success"}');
   });
 
+  group('lite tools and interrupts', () {
+    test('runs a tool and feeds its output back to the model', () async {
+      final weather = Tool(
+        name: 'weather',
+        description: 'Gets the weather.',
+        inputSchema: .map(.string(), .dynamicSchema()),
+        fn: (input, ctx) async => .response('sunny'),
+      );
+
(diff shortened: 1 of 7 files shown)


pavelgj:
@cabljac PTAL

packages/genkit/test/lite_test.dart
```diff
@@ -214,6 +214,232 @@ void main() {
...
+        model: model,
+        prompt: 'what is the weather?',
+        tools: [weather],
+      );
+
+      expect(response.text, 'weather is sunny');
+    });
+
+    test('a tool returning .interrupt halts the generation loop', () async {
+      final needsApproval = Tool(
```
cabljac, line 265: nit: are there any tests using `lite.Tool` etc, covering these exports directly? If someone deleted those exports today, seems like things would still pass?

cabljac (approved):
lgtm, one nit on test coverage

packages/genkit/test/lite_test.dart
pavelgj: Done.

PR #400, opened 2026-08-27
fix(genkit_anthropic)!: re-send thinking blocks on multi-turn requests

### LINKED ISSUE #353: fix(genkit_anthropic): thinking blocks are dropped from multi-turn requests
`toAnthropicMessage` has no ReasoningPart branch - thinking blocks are filtered from input (`plugin_impl.dart:256-291`; asserted intentional by `unit_test.dart:130`), and redacted thinking is mapped to an empty text part (`:359`) then removed by the empty-text filter (`:361`) - dropped entirely. Anthropic docs: "you must pass the thinking blocks from the assistant message back to the API, complete and unmodified." Multi-turn thinking + tool loops are broken.

Fix: re-send thinking blocks with signatures; preserve redacted blocks; align signature metadata key with JS (`thoughtSignature`, currently `signature`, `:357`). SDK constraint: anthropic_sdk_dart 2.2.0 has no thinking/redacted_thinking input-block variant (`InputContentBlock` factories end at `toolReference`), so re-sending needs the unknown-block escape hatch or an SDK change.

Related: #301 - streaming aggregation drops `thoughtSignature` in `genkit_firebase_ai`; same metadata family.

adesinah:
`toAnthropicMessage` had no ReasoningPart branch, so thinking blocks were dropped from every follow-up request, and redacted thinking was mapped to an empty text part and filtered out. Anthropic requires thinking blocks be
echoed back complete and unmodified, so multi-turn thinking and tool loops failed on the second turn.

Adds a reasoning branch to `toAnthropicMessage` that rebuilds the block with
`ThinkingInputBlock` and `RedactedThinkingInputBlock`, the typed variants
`anthropic_sdk_dart` 8 provides. (An earlier revision went through
`InputContentBlock.fromJson`, the raw-JSON escape hatch, because the SDK had
no typed input block; `9daafcc` replaced that once the dependency moved to 8.)

Replay is restricted to assistant turns, unsigned reasoning is skipped rather
than sent for the API to reject, and the signature metadata key is renamed to
`thoughtSignature` to match the Gemini plugins and Genkit JS.

## Breaking changes

Three, all in how reasoning is represented. Each accepts the old shape on the
way out, so history persisted by 0.3.1 still replays; only what the plugin
*writes* changes.

- **Signature metadata key** is `thoughtSignature`, was `signature`. Matches the
  Gemini plugins and Genkit JS. A consumer reading `metadata['signature']` off a
  new response gets null.
- **Redacted thinking** comes back as `CustomPart(custom: {redactedThinking})`,
  was `ReasoningPart(reasoning: '', metadata: {redactedThinking})`. Matches what
  JS persists and reads, so a conversation can cross between the SDKs; the old
  shape round-tripped in neither direction. Also removes the empty
  `ReasoningPart` consumers saw in `message.content`.
- **`anthropic_sdk_dart` `^2.0.0` -> `^8.0.0`**, crossing six breaking majors
  (3.0 sealed unions, 4.0 `MessageRole.system`, 6.0 `user_profile_id`, 7.0
  `fallbackCreditToken`, 8.0 removed `mid_conv_system`). An app pinning
  `anthropic_sdk_dart` 2.x-7.x alongside `genkit_anthropic` will fail version
  solving. Compat is otherwise fine: 8.0.0 needs `>=3.9.0`, the repo floor is
  `^3.10.0`.

The title carries `!` so `tools/version.dart` bumps the major rather than
shipping this as a patch.

## Known limitation

Reasoning from another provider is forwarded, not dropped. The Gemini plugins
write the same `thoughtSignature` key, so a Gemini reasoning part on a model
turn is rebuilt as an Anthropic thinking block with a signature Anthropic
cannot verify, and is rejected server-side. Genkit JS collides identically.

Closes: #353


cabljac (approved):
lgtm!

davidmigloz:
fyi @adesinah  : the latest release of the `anthropic_sdk_dart`  client adds support for [preserved-thinking](https://platform.claude.com/docs/en/build-with-claude/thinking) controls, including `ThinkingInputBlock`/`RedactedThinkingInputBlock` blocks:
https://pub.dev/packages/anthropic_sdk_dart/changelog#800

adesinah:
Thanks @davidmigloz — adopted in 3d10671.

The workaround existed only because the typed variants didn't, so this removes it entirely:

```dart
// before
sdk.InputContentBlock.fromJson({
  'type': 'thinking',
  'thinking': p.reasoning ?? '',
  'signature': signature,
})

// after
sdk.ThinkingInputBlock(thinking: p.reasoning ?? '', signature: signature)
```

Same for `redacted_thinking` → `RedactedThinkingInputBlock`. Net -13 lines, and the doc comment explaining the workaround is gone with it.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -303,12 +303,46 @@ sdk.SystemPrompt? convertSystemMessage(Message m) {
...
+/// dropped rather than turned into a request the API will reject.
+List<sdk.InputContentBlock> _toAnthropicThinkingBlocks(Part p) {
+  final metadata = p.metadata;
+
+  final redacted = metadata?[_redactedThinkingKey];
+  if (redacted is String && redacted.isNotEmpty) {
+    return [sdk.RedactedThinkingInputBlock(data: redacted)];
+  }
+
+  final signature = metadata?[_thoughtSignatureKey];
```
Ehesp: Should this not also check `signature` as a fallback? `final signature = metadata?[_thoughtSignatureKey] ?? metadata?['signature'];`?

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -480,7 +514,14 @@ Message fromAnthropicMessage(sdk.Message m) {
...
-            metadata: {'signature': signature},
+            metadata: {_thoughtSignatureKey: signature},
+          ),
+          // The payload is opaque and unreadable, but Anthropic still requires
+          // it echoed back on later turns, so it is preserved rather than
+          // dropped. The reasoning text is necessarily empty.
+          sdk.RedactedThinkingBlock(:final data) => ReasoningPart(
+            reasoning: '',
+            metadata: {_redactedThinkingKey: data},
           ),
```
cabljac, line 584: JS persists a redacted block as `{ custom: { redactedThinking } }` with no `reasoning` key (`js/plugins/anthropic/src/runner/converters/shared.ts:97-99`), and reads it back from `part.custom` (`runner/base.ts:303-307`). Here it becomes `ReasoningPart(reasoning: '', metadata: {redactedThinking})`.

I checked both directions: a JS-shaped part through `toAnthropicMessage` has `isReasoning == false` (`schema_extensions.dart:147` keys on `containsKey('reasoning')`), so it falls through to the empty list at `:382` and the block is dropped. Going the other way, JS sees `reasoning: ''` (falsy), no `custom`, no `text`, and throws at `stable.ts:209`.

Suggest emitting `CustomPart(custom: {'redactedThinking': data})` here, and checking `p.custom?['redactedThinking']` before the reasoning branch in `_toAnthropicThinkingBlocks`. Accepting both shapes on the way out would also be fine. This also removes the empty `ReasoningPart` consumers currently see in `response.message.content`.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -303,12 +303,46 @@ sdk.SystemPrompt? convertSystemMessage(Message m) {
...
+const _thoughtSignatureKey = 'thoughtSignature';
+
+/// Metadata key carrying the opaque payload of a redacted thinking block.
+const _redactedThinkingKey = 'redactedThinking';
+
+/// Rebuilds the Anthropic thinking block a [ReasoningPart] came from.
+///
+/// Returns an empty list for reasoning this plugin cannot round-trip: Anthropic
+/// rejects a thinking block whose signature is missing, so a part that never
+/// carried one - hand-built history, or reasoning from another provider - is
```
cabljac: "reasoning from another provider is dropped" is not what happens. `genkit_google_genai` writes the same `thoughtSignature` key (`common_plugin.dart:576`), so a Gemini reasoning part on a model turn gets sent as an Anthropic `signature`. I ran it: `ReasoningPart(reasoning: 'gemini thought', metadata: {'thoughtSignature': 'geminiSig=='})` produces `{"type":"thinking","thinking":"gemini thought","signature":"geminiSig=="}`, which Anthropic will reject server-side.

JS has the identical collision, so this is parity, but the comment should say it is a known cross-provider limitation rather than claim the part is dropped.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -303,12 +303,46 @@ sdk.SystemPrompt? convertSystemMessage(Message m) {
...
+List<sdk.InputContentBlock> _toAnthropicThinkingBlocks(Part p) {
+  final metadata = p.metadata;
+
+  final redacted = metadata?[_redactedThinkingKey];
+  if (redacted is String && redacted.isNotEmpty) {
+    return [sdk.RedactedThinkingInputBlock(data: redacted)];
+  }
+
+  final signature = metadata?[_thoughtSignatureKey];
+  if (signature is! String || signature.isEmpty) return const [];
```
cabljac: JS throws here instead (`stable.ts:100-106`, "Anthropic thinking parts require a signature ... Preserve the metadata.thoughtSignature"). With the silent drop, a hand-built or JSON-mangled history that lost its signature produces a request with no thinking block, and the user gets an opaque server 400 rather than a client-side `INVALID_ARGUMENT` naming the missing key.

The unit test pins the drop so I assume it is deliberate. Would you be open to at least a `_logger.warning` here, or a `GenkitException(status: INVALID_ARGUMENT)` to match JS?

packages/genkit_anthropic/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ environment:
   sdk: ^3.10.0
 
 dependencies:
-  anthropic_sdk_dart: ^2.0.0
+  anthropic_sdk_dart: ^8.0.0
```
cabljac, line 12: `^2.0.0 -> ^8.0.0` crosses six breaking majors (3.0 sealed unions, 4.0 `MessageRole.system`, 6.0 `user_profile_id`, 7.0 `fallbackCreditToken`, 8.0 removed `mid_conv_system`). `tools/version.dart:80` maps `fix`/`refactor` to a patch bump, so this ships as a patch of `genkit_anthropic`, and with squash-merge only the PR title survives into the changelog. Any app pinning `anthropic_sdk_dart` 2.x-7.x alongside `genkit_anthropic` will fail version solving.

Compat is fine (8.0.0 needs `>=3.9.0`, repo floor is `^3.10.0`, CI green). I just think the bump wants calling out somewhere release-visible, either in the PR title/body as a breaking footer or as its own changelog line.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -480,7 +514,14 @@ Message fromAnthropicMessage(sdk.Message m) {
                       as Part,
           sdk.ThinkingBlock(:final thinking, :final signature) => ReasoningPart(
             reasoning: thinking,
-            metadata: {'signature': signature},
+            metadata: {_thoughtSignatureKey: signature},
```
cabljac, line 573: nit: `signature` -> `thoughtSignature` is the right key (matches JS `shared.ts:88` and the Gemini plugins, Go is the odd one out), but any consumer reading `metadata['signature']` from Anthropic reasoning parts now silently gets null. Worth a changelog line.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -303,12 +303,46 @@ sdk.SystemPrompt? convertSystemMessage(Message m) {
...
+    return [sdk.RedactedThinkingInputBlock(data: redacted)];
+  }
+
+  final signature = metadata?[_thoughtSignatureKey];
+  if (signature is! String || signature.isEmpty) return const [];
+
+  return [
+    sdk.ThinkingInputBlock(thinking: p.reasoning ?? '', signature: signature),
+  ];
+}
```
cabljac, line 388: nit: every path returns 0 or 1 elements. `sdk.InputContentBlock?` with `[?_toAnthropicThinkingBlock(p)]` at the call site reads more directly than the `List` plus `const []` (which is `List<Never>`).

packages/genkit_anthropic/test/test_live.dart
```diff
@@ -148,6 +148,45 @@ void main() {
       );
     }, timeout: Timeout(Duration(minutes: 2)));
 
+    test('should replay thinking blocks through a tool loop', () async {
```
cabljac, line 151: nit: this is the only test that proves the server accepts the replayed block, and it never runs under `dart test` (file is `test_live.dart`, not `*_test.dart`), nor skips without a key (`:39` throws `Bad state`). Pre-existing, not this PR's problem, but do you have a run log from this passing against the API you could paste into the PR?

Related, nothing pins the streaming path: `_handleStreaming` ignores `SignatureDelta` and relies on the SDK's `MessageStreamAccumulator` to merge it before `fromAnthropicMessage`. I checked the 8.0.0 accumulator and it does (`message_stream_accumulator.dart:107,308-311`), so behaviour is correct, but a mock-SSE test asserting `thoughtSignature` and `redactedThinking` on the final message would stop a future SDK bump breaking it silently.

cabljac (changes requested):
Round-trip for a pure-Dart conversation looks right and the wire/round-trip tests pin it. Two things to sort before merge: the redacted-thinking part shape diverges from JS in a way that breaks histories crossing runtimes, and the doc comment on `_toAnthropicThinkingBlocks` is wrong for Gemini-produced history. Rest is smaller, inline.

Also the PR body still describes the `InputContentBlock.fromJson` escape hatch that the second commit removed, worth a refresh.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: you're right, and for a reason I missed at first. This PR renames the key: `main` writes `metadata['signature']` (plugin_impl.dart:483 on main), this branch writes `thoughtSignature`.

So any conversation persisted by 0.3.1 carries `signature`, and after upgrading the new read path wouldn't find it and would drop the block; the exact failure this PR exists to fix, just moved to existing sessions.

Added the fallback as you wrote it, with the current key taking precedence. Read-only, we still write `thoughtSignature`. Two wire tests cover it: replay under the old key, and precedence when a part carries both.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Confirmed, and fixed the comment. It now says a hand-built part without a signature is dropped, and calls the cross-provider case a known limitation — the Gemini plugins write the same key, so the part is forwarded and Anthropic rejects it server-side. Noted that JS collides identically.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Handled the compat half of this: `_toAnthropicThinkingBlocks` now falls back to `metadata['signature']` so history persisted by 0.3.1 still replays (see Elliot's thread above). We still only write `thoughtSignature`.

A consumer reading `metadata['signature']` off new responses does still get null, so the changelog line stands — I'll put it in the PR body since melos generates the changelog from the commits.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Fixed, taking both halves. `fromAnthropicMessage` emits `CustomPart(custom: {redactedThinking})` now, and the outbound path checks `p.custom` first, falling back to the old `metadata` shape so 0.3.1 history still replays.

Also had to widen the branch condition in `toAnthropicMessage` — a CustomPart isn't `isReasoning`, so it would have fallen straight through as you described. Wire tests cover both shapes, and the empty ReasoningPart is gone from `message.content`.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Took the warning, not the throw. The drop is deliberate: Anthropic only requires the blocks back in specific tool-use flows, so a history that lost its signature usually still completes, and throwing would break cases that work today. A 400 is the worse outcome but not the only one.

`_logger.warning` now names the missing key and says to preserve the model turn's metadata, so the diagnostic is there without the failure mode.

packages/genkit_anthropic/pubspec.yaml
adesinah: Good catch on the patch bump. Retitled to `fix(genkit_anthropic)!:` :  `version.dart:59` reads the `!` and `bumpType` maps breaking to major, and with squash-merge the title is what survives, so that's the release-visible spot.

Added a Breaking changes section to the body covering all three: the SDK bump with the six majors, the `thoughtSignature` rename, and the redacted-block shape change.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Done: `_toAnthropicThinkingBlock` returns `sdk.InputContentBlock?` and the call site is `[?_toAnthropicThinkingBlock(p)]`.

cabljac:
Code side looks right to me. I re-checked the thinking-block shapes against the JS plugin (`shared.ts:81-99`, `base.ts:296-307`) and the round-trip is byte-identical, so no concerns there. 102/102 passing locally.

Four things left, all docs/hygiene:

1. PR body para 2 still describes the `InputContentBlock.fromJson` hatch that 9daafcc removed. Worth a refresh before merge.
2. `plugin_impl.dart:324` doc says "returns an empty list", but `:333` returns `InputContentBlock?` and the `return null` at `:348` is reachable.
3. `plugin_impl.dart:342`: a malformed redacted block gets dropped, and the warning names `thoughtSignature`, which isn't the key being checked. Repro:

```dart
custom: {'redactedThinking': ''}   // -> content: [], warning names the wrong key
```

4. T7a still has no live run log. Given this changes multi-turn behaviour, can we get one before merge?

Also wants a CHANGELOG entry for the break.

One adjacent thing, separate from this PR: an assistant turn whose only part gets dropped serialises to `{"role":"assistant","content":[]}`, which Anthropic 400s. I verified it reproduces on main so it isn't a regression from you, but I accepted the T3 rebuttal earlier and I don't think it holds. Do you agree? If so I'll open a separate issue rather than widen this one.

adesinah:
1, 2, 3 done in `319763c`.

**Body.** Refreshed — it now describes the typed `ThinkingInputBlock` path and notes the `fromJson` hatch as the revision `9daafcc` replaced, since the history is otherwise confusing.

**Doc.** "Returns an empty list" was mine, left over from when it returned a `List`. Now says it returns null and what null means.

**Malformed redacted block.** Reproduced exactly as you wrote it — `content: []` and a warning naming `thoughtSignature`. Fixed by judging the part on the key it carries rather than letting it fall through: anything with `redactedThinking` present is handled as a redacted block, and an unusable payload gets its own message. Test asserts the warning names `redactedThinking` and not `thoughtSignature`.

**CHANGELOG.** I don't think there's an entry to hand-write — `packages/*/CHANGELOG.md` is generated at release from commit subjects, and the existing "### Breaking Changes" sections are all melos output. The title already carries `!`, so `version.dart` bumps the major and the subject becomes the line. Happy to add one manually if you'd rather not rely on that.

**T7a.** Still no log — I don't have a key. It's the one claim in this PR nothing offline can prove, so I agree it should block merge.

**Empty content.** You're right, and I'd rather not defend the earlier rebuttal — I can't find it in the threads and the behaviour is plain enough. Verified on a clean `origin/main` worktree:

```
ReasoningPart(reasoning: 'unsigned thought')  ->  {"role":"assistant","content":[]}
```

Pre-existing, but this PR makes it easier to reach: it adds two more ways for a part to be dropped. A separate issue is the right call — the fix is dropping the message when its content empties, not dropping the part, and that touches every branch of `toAnthropicMessage`.

103 pass, `--fatal-infos` clean.

adesinah:
T7a has a run log now — full live suite against the API, 8/8:

```
00:00 +0: loading test/test_live.dart
00:00 +0: Anthropic Integration should generate simple text
00:02 +1: Anthropic Integration should stream text
00:03 +2: Anthropic Integration should generate structured output
00:05 +3: Anthropic Integration should stream structured output
00:06 +4: Anthropic Integration should use tools
00:10 +5: Anthropic Integration should support thinking
01:02 +6: Anthropic Integration should replay thinking blocks through a tool loop
01:07 +7: Anthropic Integration should use adaptive thinking for newer models
01:09 +8: All tests passed!
```

`should replay thinking blocks through a tool loop` is the one that mattered: it proves the server accepts the replayed block, which no mock can. It also took ~52s of the run, which is the tool loop actually going out and back rather than anything hanging.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -303,12 +303,102 @@ sdk.SystemPrompt? convertSystemMessage(Message m) {
...
+///
+/// Reasoning from another provider is a known limitation, not something this
+/// catches. The Gemini plugins write the same [_thoughtSignatureKey], so a
+/// Gemini part arrives with a signature Anthropic cannot verify and is
+/// forwarded, then rejected server-side. Genkit JS collides the same way.
+sdk.InputContentBlock? _toAnthropicThinkingBlock(Part p) {
+  // A part carrying the key at all is a redacted block, so it is judged as
+  // one. Falling through to the signature check would blame a missing
+  // thoughtSignature for a payload problem, naming a key nothing read.
+  if (_claimsRedactedThinking(p)) {
```
cabljac, line 338: nit: this short-circuits on key presence, so a part carrying both a valid `thoughtSignature` and a `redactedThinking` key drops the real thought and its signature. Only reachable from a hand-built part, since the two keys land on disjoint part types coming back from the API. A test pinning the precedence would say it's deliberate.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -303,12 +303,102 @@ sdk.SystemPrompt? convertSystemMessage(Message m) {
...
+      'exactly as it arrived.',
+    );
+    return null;
+  }
+
+  final metadata = p.metadata;
+  final signature =
+      metadata?[_thoughtSignatureKey] ?? metadata?[_legacyThoughtSignatureKey];
+  if (signature is! String || signature.isEmpty) {
+    _logger.warning(
```
cabljac, line 355: nit: we reconvert the full history on each request, so one unsigned part in turn 1 warns on every turn after it. Drop to `fine` after the first, or log once per request.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -303,12 +303,102 @@ sdk.SystemPrompt? convertSystemMessage(Message m) {
...
+  return null;
+}
+
 /// Converts a Genkit [Message] to an Anthropic [sdk.InputMessage].
 sdk.InputMessage toAnthropicMessage(Message m) {
   final isUser = m.role == Role.user || m.role == Role.tool;
 
   final blocks = m.content.expand<sdk.InputContentBlock>((p) {
-    if (p.isText) {
+    if (p.isReasoning || p.custom?.containsKey(_redactedThinkingKey) == true) {
```
cabljac, line 395: Anthropic's docs cover this one: toggling thinking mid-turn doesn't error, the API silently disables it for that request and may strip blocks that would make the turn structure invalid. So replaying here is fine. Worth a sentence on the `_toAnthropicThinkingBlock` doc comment saying so, since the wire tests assert the send and it otherwise reads like an oversight.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -480,7 +570,17 @@ Message fromAnthropicMessage(sdk.Message m) {
                       as Part,
           sdk.ThinkingBlock(:final thinking, :final signature) => ReasoningPart(
             reasoning: thinking,
-            metadata: {'signature': signature},
+            metadata: {_thoughtSignatureKey: signature},
+          ),
+          // The payload is opaque and unreadable, but Anthropic still requires
```
cabljac, line 575: nit: two stacked comment blocks here. The first one ("The reasoning text is necessarily empty") is left over from the `ReasoningPart` version and contradicts the `CustomPart` described just below.

cabljac (approved):
Changelog's generated from the squash subject, so let's leave it as is, no need for a hand-written entry.

Few nits inline, none blocking.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Pinned with a test — redacted wins, and the comment now says why: its payload is the half that can't be reconstructed. A thought whose signature was lost can be produced again; an opaque payload can't.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Once per request now, `fine` after that. Reset at the top of generate — conversion is synchronous, so the flag is request-scoped in practice.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Added that to the doc comment, near enough verbatim — the API disables thinking for the request and strips blocks that would leave the turn structure invalid, so sending them unconditionally is safe. Useful pointer, thanks.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Removed the leftover sentence — it described the ReasoningPart shape this PR replaced.

packages/genkit_anthropic/pubspec.yaml
adesinah: Rebased onto main, which conflicted here: #434 relaxed this to `>=2.1.0 <9.0.0` as part of a monorepo-wide constraint relaxation.

I kept `^8.0.0`. That relaxation is right for main — it builds thinking blocks through `InputContentBlock.fromJson` — but this branch calls `ThinkingInputBlock` and `RedactedThinkingInputBlock` directly, which 8.0.0 introduced, so a 2.x resolution would not compile. Comment in the pubspec says as much.

Which means your original point stands rather than being resolved by #434: the floor still moves for anyone on 2.x-7.x. That is covered by the `!` in the title and the Breaking changes section in the body.

packages/genkit_anthropic/pubspec.yaml
```diff
@@ -9,7 +9,10 @@ environment:
   sdk: ^3.12.0
 
 dependencies:
-  anthropic_sdk_dart: '>=2.1.0 <9.0.0'
+  # Floor is 8.0.0, not the workspace-wide >=2.1.0: this plugin now builds
+  # thinking blocks with ThinkingInputBlock / RedactedThinkingInputBlock, which
+  # 8.0.0 introduced. Resolving anything older would not compile.
+  anthropic_sdk_dart: ^8.0.0
```
cabljac, line 15: Picking this back up from your CHANGELOG note above. You're right that the sections are generated, but the `0.3.2-wip` entry isn't: kevmoo hand-wrote it in #434 and the rebase carried it. It says the constraint widened, while this line narrows it to ^8.0.0.

Generation won't clean it up either, the new entry gets prepended above it. So worth correcting it to say the floor moved to 8.0.0, which is also the one fact the generated line can't carry since the subject doesn't mention the SDK.

(Commenting here because CHANGELOG.md isn't in this PR's diff.)

ifielker (approved):

packages/genkit_anthropic/pubspec.yaml
adesinah: Good catch, fixed in df46b39 — the entry now says the floor moved to 8.0.0.

PR #402, opened 2026-08-27
feat(genkit_anthropic): add beta/stable API selection and native structured outputs

### LINKED ISSUE #356: feat(genkit_anthropic): beta/stable API selection and native structured outputs
No `apiVersion`/betas concept - always stable `messages.create`. JS selects stable/beta per request and uses native `output_format: json_schema` (beta) for structured output. Dart's forced-`return_output`-tool approach conflicts with manual thinking (forced tool_choice errors there, per API docs). Prerequisite for structured outputs, effort, files API.

adesinah:
The plugin always called stable `messages.create`, so no beta-gated feature was reachable, and structured output was faked with a synthetic return_output tool forced via `tool_choice` - which Anthropic rejects alongside extended thinking.

The PR adds apiVersion ('stable' | 'beta') on both the plugin and per-request config, with the request winning and stable as the default, plus a betas override for features the curated list does not know yet. Beta requests send the anthropic-beta header on the unary and streaming paths alike. On beta, a model on Anthropic's Structured Outputs list sends the schema natively via `output_config.format`, so structured output composes with manual thinking. Stable keeps the forced-tool fallback, which now reports `INVALID_ARGUMENT` when combined with manual thinking rather than letting the API reject the request.

**Rebased onto #453, and the capability claims are reconciled with it.** `supports.constrained` drives core's simulation now, so it has to name the mechanism that will actually serve the request, and the mechanism depends on the surface:

| | curated, on Structured Outputs list | uncurated |
| --- | --- | --- |
| **beta** | `output_config.format`, claims `constrained: true` | claims nothing; core simulates |
| **stable** | forced `return_output` tool, claims `'no-tools'` | claims nothing; core simulates |

That `true` is the narrowing #453 deferred here: `'no-tools'` existed only because the forced tool pins `tool_choice`, and the native path pins nothing, so the caller's own tools stay reachable. The claim follows the plugin's `apiVersion` rather than the request's, because core reads it before the request's config is in hand — a request that asks for beta may be simulated where native would have served it (a longer prompt), and one that overrides to stable while carrying tools is refused outright, since neither mechanism can serve that.

The uncurated default stays as #453 set it: no claim, so core simulates. An earlier commit here assumed uncurated names support native output; that was written to dodge the forced-tool 400 on `claude-fable-5-1`, which simulation now handles without guessing at a per-model list.

Also fixes a caller-supplied toolChoice silently unforcing `return_output`.

Closes #356

### DIFF
--- a/packages/genkit_anthropic/lib/src/plugin_impl.dart
+++ b/packages/genkit_anthropic/lib/src/plugin_impl.dart
@@ -31,6 +31,26 @@ final _logger = Logger('genkit_anthropic');
 /// entry.
 final commonModelInfo = ModelInfo(supports: baseClaudeSupports);
 
+/// Beta features requested when a request resolves to the beta API surface.
+///
+/// Sent as the `anthropic-beta` header. Mirrors the list the Genkit JS plugin
+/// enables by default; a request can replace it via `AnthropicOptions.betas`.
+const defaultAnthropicBetas = <String>[
+  'files-api-2025-04-14',
+  'effort-2025-11-24',
+  'structured-outputs-2025-11-13',
+  'task-budgets-2026-03-13',
+];
+
+/// Resolves whether a request runs against the beta API surface.
+///
+/// The request's own `apiVersion` wins, then the plugin default, else stable.
+@visibleForTesting
+bool resolveBetaEnabled(String? requestApiVersion, String? pluginApiVersion) {
+  final selected = requestApiVersion ?? pluginApiVersion;
+  return selected == 'beta';
+}
+
 /// Core Genkit plugin implementation for Anthropic Claude models.
 ///
 /// Automatically discovers available models from the Anthropic API and
@@ -50,6 +70,11 @@ class AnthropicPluginImpl extends GenkitPlugin {
   /// instrumentation, or injecting a mock transport in tests.
   final http.Client? httpClient;
 
+  /// Default Anthropic API surface (`'stable'` or `'beta'`) for every request.
+  ///
+  /// A request's own `apiVersion` overrides this. Defaults to stable.
+  final String? apiVersion;
+
   sdk.AnthropicClient? _client;
 
   /// Creates an [AnthropicPluginImpl].
@@ -58,6 +83,7 @@ class AnthropicPluginImpl extends GenkitPlugin {
     this.headers,
     this.baseUrl,
     this.httpClient,
+    this.apiVersion,
   });
 
   @override
@@ -169,9 +195,16 @@ class AnthropicPluginImpl extends GenkitPlugin {
 
         try {
           final createRequest = _buildCreateRequest(req, modelName, options);
+          // Empty on the stable surface, which suppresses the header entirely.
+          final betas = resolveBetaEnabled(options.apiVersion, apiVersion)
+              ? (options.betas ?? defaultAnthropicBetas)
+              : const <String>[];
 
           if (ctx.streamingRequested) {
-            final stream = requestClient.messages.createStream(createRequest);
+            final stream = requestClient.messages.createStream(
+              createRequest,
+              betas: betas,
+            );
             final accumulator = sdk.MessageStreamAccumulator();
             await for (final event in stream) {
               accumulator.add(event);
@@ -184,7 +217,10 @@ class AnthropicPluginImpl extends GenkitPlugin {
               usage: mapUsage(message.usage),
             );
           } else {
-            final response = await requestClient.messages.create(createRequest);
+            final response = await requestClient.messages.create(
+              createRequest,
+              betas: betas,
+            );
             return ModelResponse(
               finishReason: mapFinishReason(response.stopReason),
               message: fromAnthropicMessage(response),
@@ -238,26 +274,41 @@ class AnthropicPluginImpl extends GenkitPlugin {
         req.tools?.map(toAnthropicTool).toList() ?? <sdk.ToolDefinition>[];
 
     sdk.ToolChoice? toolChoice;
+    sdk.JsonOutputFormat? outputFormat;
 
     if (req.output?.schema != null) {
       final schema = Map<String, dynamic>.from(req.output!.schema!);
-      if (!schema.containsKey('type')) {
-        schema['type'] = 'object';
-      }
-      const toolName = 'return_output';
-      tools.add(
-        sdk.ToolDefinition.custom(
-          sdk.Tool(
-            name: toolName,
-            description: 'Return the structured output.',
-            inputSchema: sdk.InputSchema.fromJson(schema),
+
+      if (_supportsNativeStructuredOutput(modelName)) {
+        // Native structured output needs no forced tool, so it composes with
+        // manual thinking - unlike the fallback below. The schema goes over
+        // as-authored; adding a `type` here would collide with a `$ref` root.
+        outputFormat = sdk.JsonOutputFormat(schema: _toAnthropicSchema(schema));
+      } else {
+        // Older and uncurated models are not on Anthropic's Structured Outputs
+        // list, so the schema is served by a tool the model is forced to call.
+        _assertToolOutputAllowed(req, modelName, options);
+        // Anthropic's tool input schemas must declare an object type.
+        if (!schema.containsKey('type')) {
+          schema['type'] = 'object';
+        }
+        const toolName = 'return_output';
+        tools.add(
+          sdk.ToolDefinition.custom(
+            sdk.Tool(
+              name: toolName,
+              description: 'Return the structured output.',
+              inputSchema: sdk.InputSchema.fromJson(schema),
+            ),
           ),
-        ),
-      );
-      toolChoice = sdk.ToolChoice.tool(toolName);
+        );
+        toolChoice = sdk.ToolChoice.tool(toolName);
+      }
     }
 
-    if (req.toolChoice != null) {
+    // A caller-supplied choice must not silently unforce `return_output`; that
+    // would leave the model free to answer without producing the schema.
+    if (req.toolChoice != null && toolChoice == null) {
       toolChoice = switch (req.toolChoice) {
         'auto' => sdk.ToolChoice.auto(),
         'any' => sdk.ToolChoice.any(),
@@ -267,7 +318,7 @@ class AnthropicPluginImpl extends GenkitPlugin {
     }
 
     final thinking = _mapThinkingConfig(options.thinking, modelName);
-    final outputConfig = _mapOutputConfig(options.outputConfig);
+    final outputConfig = _mapOutputConfig(options.outputConfig, outputFormat);
 
     return sdk.MessageCreateRequest(
       model: modelName,
@@ -502,12 +553,9 @@ Map<String, dynamic> _extractOutput(Map<String, dynamic> input) {
   return input;
 }
 
-sdk.ThinkingConfig? _mapThinkingConfig(
-  ThinkingConfig? config,
-  String modelName,
-) {
-  if (config == null) return null;
-
+/// Resolves the effective thinking type, applying the curated per-model default
+/// when the request does not name one.
+String _resolveThinkingType(ThinkingConfig config, String modelName) {
   final type =
       config.type ?? knownClaudeModelFor(modelName)?.defaultThinkingMode.name;
   if (type == null) {
@@ -516,6 +564,16 @@ sdk.ThinkingConfig? _mapThinkingConfig(
       status: StatusCodes.INVALID_ARGUMENT,
     );
   }
+  return type;
+}
+
+sdk.ThinkingConfig? _mapThinkingConfig(
+  ThinkingConfig? config,
+  String modelName,
+) {
+  if (config == null) return null;
+
+  final type = _resolveThinkingType(config, modelName);
 
   return switch (type) {
     'disabled' => sdk.ThinkingConfig.disabled(),
@@ -531,11 +589,83 @@ sdk.ThinkingConfig? _mapThinkingConfig(
   };
 }
 
-sdk.OutputConfig? _mapOutputConfig(AnthropicOutputConfig? config) {
+/// Whether [modelName] is on Anthropic's Structured Outputs list.
+///
+/// Curated models carry the flag; names resolved dynamically are assumed not to
+/// support it, so they keep the tool-based fallback.
+bool _supportsNativeStructuredOutput(String modelName) =>
+    knownClaudeModelFor(modelName)?.structuredOutputs ?? false;
+
+/// Rewrites a Genkit JSON schema into the shape Anthropic accepts.
+///
+/// Anthropic rejects `$schema` and requires `additionalProperties: false` on
+/// every object, including ones nested under `$defs`.
+Map<String, dynamic> _toAnthropicSchema(Map<String, dynamic> schema) {
+  final out = <String, dynamic>{};
+  for (final entry in schema.entries) {
+    if (entry.key == r'$schema') continue;
+    out[entry.key] = _normalizeSchemaValue(entry.value);
+  }
+  // A `$ref` node may not carry sibling constraints; Anthropic rejects the
+  // combination outright. Named Genkit schemas arrive as a bare `$ref` plus
+  // `$defs`, and the recursion above has already closed the definitions.
+  if (out.containsKey(r'$ref')) return out;
+
+  // A hand-written schema may describe an object through its keywords alone.
+  // Anthropic needs the type spelled out before it will accept the closed
+  // marker below, so infer it the way the tool fallback does.
+  if (!out.containsKey('type') &&
+      (out.containsKey('properties') || out.containsKey('required'))) {
+    out['type'] = 'object';
+  }
+  if (out['type'] == 'object') {
+    out['additionalProperties'] = false;
+  }
+  return out;
+}
+
+Object? _normalizeSchemaValue(Object? value) => switch (value) {
+  final Map<String, dynamic> map => _toAnthropicSchema(map),
+  final Map map => _toAnthropicSchema(map.cast<String, dynamic>()),
+  final List list => list.map(_normalizeSchemaValue).toList(),
+  _ => value,
+};
+
+/// Guards the tool-based structured-output fallback against manual thinking.
+///
+/// The fallback forces `tool_choice`, which Anthropic rejects when extended
+/// thinking is on. Native structured output has no such conflict, so the fix is
+/// to move to a model that supports it rather than to drop thinking.
+void _assertToolOutputAllowed(
+  ModelRequest req,
+  String modelName,
+  AnthropicOptions options,
+) {
+  final thinking = options.thinking;
+  if (thinking == null) return;
+  // Resolve through the same path the wire uses, so a bare ThinkingConfig()
+  // that defaults to manual on a 4.5-era model is caught too.
+  if (_resolveThinkingType(thinking, modelName) != 'enabled') return;
+
+  throw GenkitException(
+    'Structured output with manual thinking is not supported for '
+    '"$modelName": it needs a forced tool call, which Anthropic rejects '
+    'alongside extended thinking. Use a model with native structured output '
+    'support, or set thinking.type to "adaptive" or "disabled".',
+    status: StatusCodes.INVALID_ARGUMENT,
+  );
+}
+
+sdk.OutputConfig? _mapOutputConfig(
+  AnthropicOutputConfig? config,
+  sdk.JsonOutputFormat? format,
+) {
   final effort = config?.effort;
-  if (effort == null) return null;
+  if (effort == null && format == null) return null;
+  if (effort == null) return sdk.OutputConfig(format: format);
 
   return sdk.OutputConfig(
+    format: format,
     effort: switch (effort) {
       'low' => sdk.EffortLevel.low,
       'medium' => sdk.EffortLevel.medium,
--- a/packages/genkit_anthropic/lib/src/plugin_impl.dart
+++ b/packages/genkit_anthropic/lib/src/plugin_impl.dart
@@ -33,21 +33,39 @@ final commonModelInfo = ModelInfo(supports: baseClaudeSupports);
 
 /// Beta features requested when a request resolves to the beta API surface.
 ///
-/// Sent as the `anthropic-beta` header. Mirrors the list the Genkit JS plugin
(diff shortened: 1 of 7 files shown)


cabljac:
Beta plumbing looks right and the stable default matches pre-PR behaviour. The structured-output path has a few request-shaping problems, most of them in `_toAnthropicSchema`. I ran each of these against the live API; server responses quoted below.

**Nullable objects lose `additionalProperties`.** `plugin_impl.dart:621` closes an object with `out['type'] == 'object'`, but a nullable object arrives as `"type": ["object","null"]` and fails the equality check:

```json
"properties":{"nested":{"type":["object","null"],"properties":{"a":{"type":"string"}}}}
```

The nested object goes out with no `additionalProperties`:

```
400 output_config.format.schema: For 'object' type, 'additionalProperties' must be explicitly set to false
```

The type inference just above at `:617-619` has the same blind spot. Treating `type` as either a string or a list containing `object` fixes both.

**`_normalizeSchemaValue` recurses into containers, not just schema nodes.** `:627-632` descends into every `Map` regardless of position, so `_toAnthropicSchema` then applies type inference and closing to the `properties` and `$defs` maps themselves. It bites when a field is named `type`, `properties` or `required`:

```json
in:  {"type":"object","properties":{"required":{"type":"boolean"}}}
out: {"type":"object","properties":{"required":{"type":"boolean"},"type":"object","additionalProperties":false},"additionalProperties":false}
```

`$defs` corrupts the same way, and it still happens when the root takes the `$ref` early-return at `:612`, since the damage is done on the way down.

**Validation keywords now reach a path that rejects them.** `minimum`/`maximum`/`minLength` ride through to `output_config.format`:

```
400 output_config.format.schema: For 'integer' type, properties maximum, minimum are not supported
```

They rode through the forced-tool path too, but `input_schema` never validated them and the tool carries no `strict: true`, so nothing complained. `model.g.dart:240-256` shows the Genkit generator emitting `minimum: 1` and `minimum: 0.0, maximum: 1.0`, so `@IntegerField(minimum:)` output types hit this.

**Uncurated models still take the forced-`tool_choice` fallback.** `_supportsNativeStructuredOutput` at `:596` is an exact-map lookup, so `claude-fable-5-1` and `claude-mythos-5-1` fall through to `tool_choice: {type: tool}`:

```
400 tool_choice: type "tool" and "any" are not supported for this model.
```

Adding them to `KnownClaudeModel` works, but since every model Anthropic currently lists supports structured outputs, inverting the default may age better: unknown names go native, and the tool fallback is reserved for curated entries that explicitly opt out.

**`tool_choice` can now go out with `tools: null`.** `:331` nulls empty tools, and the new guard at `:309` lets a caller-supplied `toolChoice` survive. Curated model + output schema + caller `toolChoice` + no user tools:

```
400 tool_choice.any may only be specified while providing tools
```

Newly reachable, since an output schema used to always append `return_output`.

On `defaultAnthropicBetas` at `:38-43`: three of the four are graduated, and all four are inert here (the plugin never touches `/v1/files`, exposes no `taskBudget`, and already sends `output_config.format` on the stable surface). I checked `effort-2025-11-24` and it is still accepted today, so nothing is broken right now. But Anthropic 400s on an unrecognised beta name rather than ignoring it, so when a transition alias is eventually retired every `apiVersion: 'beta'` request fails until a release. Worth either trimming to what the plugin consumes, or saying in the PR body that the JS-parity list is deliberate.

nit: `apiVersion` isn't validated. `resolveBetaEnabled` is `selected == 'beta'`, so `'Beta'` silently means stable. Same shape: `betas: []` is treated as "send nothing" rather than "send the defaults", so `AnthropicOptions(apiVersion: 'beta', betas: [])` sends no header at all.

Not breaking as far as I can see, and "feat" is right. Curated models do move from forced `return_output` to `output_config.format` on the wire, which changes prompt cache keys and makes schema validation strict, so worth a release note.

cabljac (changes requested):
looks the right direction, see my long review comment above

adesinah:
Reproduced all five request-shaping bugs on the wire before fixing — the schema ones offline, so they're pinned by tests now rather than only by your live run. Thanks for the server responses, they made each one unambiguous.

**Nullable objects.** Fixed. `_isObjectType` accepts `"object"` or a list containing it, used for both the inference and the closing.

**Recursion into containers.** Fixed properly rather than patched. `_toAnthropicSchema` now follows JSON Schema structure — `properties`/`$defs`/`patternProperties` recurse into values, `allOf`/`anyOf`/`oneOf`/`prefixItems` into elements, `items`/`not`/`if`/`then`/`else`/`contains`/`propertyNames` as single schemas, everything else untouched. Your `{"required": {"type": "boolean"}}` case is a test, along with fields named `type` and `properties`.

**Validation keywords.** Stripped on the native path only — the tool path never validated them and I didn't want to change what it sends. Covers the numeric, string, array and object constraint families, not just the three you hit.

**Uncurated models.** Took your inversion rather than adding the two names: unknown goes native, the fallback is opt-out via `structuredOutputs: false`. Worth flagging that no curated entry sets it false today, so the fallback and `_assertToolOutputAllowed` are now unreachable in practice and their tests had to go. Options are a curated opt-out for a genuinely pre-structured-outputs model, or deleting the fallback. I didn't want to guess at either — which would you prefer?

**`tool_choice` with `tools: null`.** Fixed, and it also fired for a curated model, not just the caller case. Test covers both.

**Betas.** Trimmed to the two the plugin consumes, `effort-2025-11-24` and `structured-outputs-2025-11-13`, with the reasoning and the `betas:` escape hatch in the dartdoc. `files-api` and `task-budgets` buy nothing here and are two fewer names to rot.

**`apiVersion`.** Now throws INVALID_ARGUMENT on anything but `beta`/`stable`. Left `betas: []` meaning "send none" — it reads as explicit to me, and I added a comment saying so, but happy to flip it.

Release note on curated models moving to `output_config.format`: agreed, will add to the body.

116 pass, `--fatal-infos` clean, full workspace green. Rebased onto current main too — it had moved and the branch predated it.

cabljac (commented):
All five fixes check out at cd0fcc1, and the tests pin each one.

**Fallback.** Delete it. `knownClaudeModelFor(name)?.structuredOutputs ?? true` is a constant `true`, so `return_output` and `_assertToolOutputAllowed` are unreachable. The gap it leaves belongs to core, not the plugin: #433 and #426.

**`betas: []`.** Keep it as "send none".

One question: is `structured-outputs-2025-11-13` load-bearing? Stable sends `output_config.format` with no beta header and your live run passed, so the name may be inert.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -531,11 +613,171 @@ sdk.ThinkingConfig? _mapThinkingConfig(
...
+/// currently lists does, and assuming otherwise sent a newly released model
+/// down the tool fallback, which fails on exactly the models that do not
+/// accept a forced tool choice:
+///
+///     400 tool_choice: type "tool" and "any" are not supported for this model
+///
+/// So the fallback is opt-out now: a curated entry sets `structuredOutputs:
+/// false` to claim it, rather than every unknown name inheriting it.
+bool _supportsNativeStructuredOutput(String modelName) =>
+    knownClaudeModelFor(modelName)?.structuredOutputs ?? true;
```
cabljac, line 628: This is `true` for every input. No `KnownClaudeModel` sets `structuredOutputs: false` and the field defaults true at known_models.dart:75, so the else branch at :309-326, `_assertToolOutputAllowed` and its INVALID_ARGUMENT throw are all unreachable. I'd delete the fallback and let #433 own the gap. More in the top-level comment.

packages/genkit_anthropic/test/test_live.dart
```diff
@@ -110,6 +111,64 @@ void main() {
...
+          thinking: ThinkingConfig(type: 'enabled', budgetTokens: 1024),
+        ),
+      );
+
+      expect(response.output, isNotNull);
+      expect(response.output!.name, 'John Doe');
+      expect(response.output!.age, 30);
+    }, timeout: Timeout(Duration(minutes: 2)));
+
+    test('should generate structured output via the forced-tool '
```
cabljac, line 132: This no longer covers the fallback. `knownClaudeModelFor(id) == null` selects an uncurated model, and after the inversion at plugin_impl.dart:628 an uncurated model goes native, so this runs the same path as the test above it. Nothing exercises `return_output` now, live or offline. If the fallback goes, this goes with it.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -531,11 +613,171 @@ sdk.ThinkingConfig? _mapThinkingConfig(
...
+          field.key.toString(): _asSchema(field.value),
+      };
+    } else if (_schemaListKeywords.contains(key) && value is List) {
+      out[key] = value.map(_asSchema).toList();
+    } else if (_schemaValuedKeywords.contains(key)) {
+      // `items` may be a list of schemas in older drafts.
+      out[key] = value is List
+          ? value.map(_asSchema).toList()
+          : _asSchema(value);
+    } else if (key == 'additionalProperties' && value is Map) {
```
cabljac, line 715: nit: this recursion is discarded whenever the node is an object, since :733 overwrites `additionalProperties` with `false`. It only survives on a node with no `type` and no `properties`/`required`, where `additionalProperties` means nothing. Probably just drop the branch.

adesinah:
Two things from #453, which makes `supports.constrained` something core actually reads at generate time.

Once that lands the forced-tool branch here is unreachable. Core simulates constrained generation for any model declaring no `constrained`, and the simulation strips `output.schema` before the plugin sees the request. `baseClaudeSupports` withholds the claim, so every model that would have taken the `else` branch now arrives with no schema at all. That takes `_assertToolOutputAllowed` and the `toolChoice`-unforcing fix with it, since both need `output.schema != null` on an unclaimed model. Probably worth deleting the branch rather than shipping it with a better error.

The other is that two defaults disagree. `_supportsNativeStructuredOutput` reads an uncurated name as supporting native structured output, but `commonModelInfo` withholds `constrained` for those same names. Under #453 the declaration wins, so an uncurated model gets simulated for and never reaches the native path. The reasoning in this PR is the stronger of the two — a newly released model shouldn't drop into the tool fallback and hit `tool_choice: type "tool" and "any" are not supported for this model` — so `commonModelInfo` probably wants `constrained: true`, leaving `structuredOutputs: false` as the opt-out.

adesinah:
One more, and this one I haven't verified — worth a check before merge.

`_supportsNativeStructuredOutput` isn't gated on `resolveBetaEnabled`, but `structured-outputs-2025-11-13` is in `defaultAnthropicBetas` and `apiVersion` defaults to `'stable'`. So a curated model on default config sends `output_config.format` with no `anthropic-beta` header.

If the stable surface rejects the field, the native path is unreachable unless the caller opts into `apiVersion: 'beta'`, and the models that declare `constrained: true` would 400 on a plain structured-output request. If it accepts it, nothing to do. Either way it's one live call against a curated model with default config to settle.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -183,10 +254,34 @@ class AnthropicPluginImpl extends GenkitPlugin {
...
+            req,
+            modelName,
+            options,
+            // The header is what unlocks `output_config.format`, and `betas`
+            // replaces the curated list rather than adding to it - so a caller
+            // opting into some other beta by name takes this one away, and the
+            // schema has to travel by tool instead. Reading the resolved list
+            // rather than the surface keeps the request and its header from
+            // disagreeing.
+            nativeOutput: betas.any(_isStructuredOutputsBeta),
```
cabljac: This gate is not needed. Live on `claude-sonnet-4-5` with no beta header, `output_config.format` returns 200 on its own, with `thinking: {type: enabled}`, and with a tool present. Docs: "beta headers are no longer required". With the gate, default config (stable) falls back to `return_output` and refuses two things the API serves: schema + manual thinking (INVALID_ARGUMENT at :878) and schema + tools (the `'no-tools'` claim makes core strip the schema).

Fix: send `output_config.format` on both surfaces, claim `constrained: true` everywhere, delete the `return_output` branch (:375-405) and the `_assertForcedOutputToolAllowed` checks. `structured-outputs-2025-11-13` can leave `defaultAnthropicBetas` too; it is only accepted for a transition period. The "beta-only" wording at :376, :824-825, README:133 and structured_output_wire_test.dart:27 needs the same fix.

packages/genkit_anthropic/lib/src/schema.dart
```diff
@@ -0,0 +1,140 @@
...
+          field.key.toString(): _asSchema(field.value),
+      };
+    } else if (_schemaListKeywords.contains(key) && value is List) {
+      out[key] = value.map(_asSchema).toList();
+    } else if (_schemaValuedKeywords.contains(key)) {
+      // `items` may be a list of schemas in older drafts.
+      out[key] = value is List
+          ? value.map(_asSchema).toList()
+          : _asSchema(value);
+    } else if (key == 'additionalProperties' && value is Map) {
```
cabljac: Live 400: "For 'object' type, 'additionalProperties: object' is not supported. Please set 'additionalProperties' to false". Same for `true` and `{type: string}`. schemantic emits every `Map<String, T>` field as `additionalProperties: {}` (genkit/lib/src/types.g.dart:370), so the dictionary case at test :403 fails on the wire. JS sets it to `false` (js/plugins/anthropic/src/runner/beta.ts:107-109); follow that, or document `Map` fields as unsupported natively.

Same pass-through at :35: `oneOf` 400s live ("Schema type 'oneOf' is not supported"), `anyOf` with the same content is accepted. `SchemanticType.nullable()` emits `oneOf`, so rewrite it to `anyOf`.

cabljac (changes requested):
Ran the live suite and some direct calls today. Stable accepts `output_config.format` with no header, with manual thinking and with tools, and the docs mark structured outputs GA (https://platform.claude.com/docs/en/docs/build-with-claude/structured-outputs). That settles the 18 Sep question and the fallback from last round: the native path works on both surfaces. Two blockers inline. Analyze and 146 unit tests pass; live 10 of 12, the two failures are `claude-3-5-haiku-latest` no longer existing, same on main.

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: You're right, and I verified it myself before taking it: `output_config.format` on the default surface, no beta header, both with manual thinking and with a caller tool, 200 in each case. Two live tests now assert exactly that rather than trusting either of our word for it.

Taken in full. The `return_output` branch is gone, `_assertForcedOutputToolAllowed` and `_supportsNativeStructuredOutput` with it, and `structured-outputs-2025-11-13` is out of `defaultAnthropicBetas` (leaving only `effort-2025-11-24`). One tier now: curated models claim `constrained: true` regardless of `apiVersion`, so `KnownClaudeModel.infoFor(beta:)` collapses back to a plain `info`. The whole surface-dependent claim I built for this is gone - it was answering a question that turned out not to exist. Net -257 lines.

Uncurated names still claim nothing and are simulated for: whether a model is on the Structured Outputs list is per-model, and the plugin has only checked what it curates. Wording fixed at the four places you listed.

packages/genkit_anthropic/lib/src/schema.dart
adesinah: Both fixed, and the first one reverses a change I made earlier in this PR - I had made `additionalProperties` preserve a dictionary's value schema, reasoning that overwriting it would constrain the model to `{}`. Your 400 settles it: the value schema is not sendable at all, so `false` it is, matching JS. The cost is real and now documented - a `Map<String, T>` field loses its value schema natively - so the README says so and points at modelling it as a string field if that matters.

`oneOf` is rewritten to `anyOf` on the way out, with the reasoning in the dartdoc: the two differ only in whether more than one branch may match, which `SchemanticType.nullable()` does not rely on. Tests for both, plus the existing nullable-object and validation-keyword cases.

140 unit tests pass, analyze clean. Live: everything green except the two simulated-path tests for `claude-3-5-haiku-latest`, which fail identically on the branch point - it just does not reliably return parseable JSON from prompt instructions. Not from this change, and worth its own look.

packages/genkit_anthropic/lib/src/schema.dart
```diff
@@ -0,0 +1,156 @@
...
+/// Anthropic rejects `$schema`, rejects the validation keywords above, and
+/// requires `additionalProperties: false` on every object, including ones
+/// nested under `$defs`.
+///
+/// Recursion follows JSON Schema structure rather than descending into every
+/// map it meets. A `properties` map is not itself a schema - descending into
+/// it applied the object inference and the closed marker to the map of field
+/// names, which corrupted any schema with a field called `type`, `properties`
+/// or `required`.
+Map<String, dynamic> toAnthropicSchema(Map<String, dynamic> schema) {
```
cabljac: A `dynamic` or `Object?` field is emitted by schemantic as `$Schema.any()`, which is `{}`, and this passes it through to `output_config.format.schema`. Anthropic rejects that:

```dart
@Schema()
abstract class $Record {
  String get name;
  dynamic get extra; // Object? behaves the same
}
```

On main (forced `return_output` tool, `{}` accepted in `input_schema`) this returns 200 with `extra` populated, sync and streaming, on claude-sonnet-4-5 and claude-sonnet-5. On this PR both fail with:

`output_config.format.schema: Empty schema ({}) that accepts any JSON value is not supported. Please specify a concrete type.`

Options I can see: detect `{}` and send that request down the simulated path, or reject it up front with INVALID_ARGUMENT naming the field. Which do you prefer? Either way a wire test pinning it would be good.

cabljac (changes requested):

packages/genkit_anthropic/lib/src/schema.dart
adesinah: Neither, in the end — I probed the API to see whether `{}` had a faithful rewrite, and the answer decided it.

`{"type": ["string","number","boolean","object","array","null"]}` is accepted, but only with `additionalProperties: false` on it, and the model then answers an object value as a *string* of JSON:

```
200  union6 + addProps:false -> {"name":"Ada","extra":"{\"city\":\"London\"}"}
```

So the concrete rewrite silently changes a field's type, which is worse than either of your options. Rejecting is honest but loses something that works on main.

So: a schema carrying an empty sub-schema anywhere is not sent as a constraint; the shape goes in the system prompt instead, in the same wording core's own simulation uses. The request still goes out, and `extra` comes back as the value it is. Live:

```
a dynamic field survives, as the value it is ... 200, extra == {city: London}
```

`isNativelyExpressible` walks schema positions only, so `{}` under `properties`, `$defs`, `anyOf`, `items` and friends is all found — four wire tests, including one that the caller's own system prompt is kept and one that an ordinary schema still goes native, plus the live one above.

Unrelated but in the same commit: `genkit_anthropic` had no `build.yaml`, alone among the plugins, so regenerating its `.g.dart` stripped the licence headers and would have failed `check-license`. Copied the config from `genkit_openai` verbatim.

144 unit tests, analyze and format clean. Live green except the two simulated-path tests for `claude-3-5-haiku-latest`, which fail identically at the branch point.

packages/genkit_anthropic/lib/src/schema.dart
```diff
@@ -0,0 +1,207 @@
...
+/// There is no faithful rewrite. A union of every concrete type is accepted,
+/// but its object branch has to carry `additionalProperties: false`, and the
+/// model then answers an object value as a *string* of JSON rather than an
+/// object - checked against the live API. Silently changing a field's type is
+/// worse than not constraining it, so a schema this returns false for is sent
+/// through the prompt instead.
+bool isNativelyExpressible(Map<String, dynamic> schema) =>
+    !_hasEmptySchema(schema);
+
+bool _hasEmptySchema(Object? node) {
```
cabljac: With this fallback in place, I think `Map` fields may want the same route, could you check? As far as I can tell the walker skips `additionalProperties`, so a `Map<String, T>` field goes native, `toAnthropicSchema` closes it to `additionalProperties: false`, and the model can only return an empty object:

```dart
@Schema()
abstract class $Record {
  String get name;
  Map<String, int> get scores; // Map<String, dynamic> behaves the same
}
```

On this PR the call succeeds with `scores == {}`, sync and streaming, on claude-sonnet-4-5 and claude-sonnet-5. On main (forced `return_output` tool) claude-sonnet-5 returns a populated map. This is the `additionalProperties` trade-off we accepted last round, before the prompt fallback existed.

Something like this in the walker might cover it:

```dart
} else if (key == 'additionalProperties' && value != false) {
  return true;
}
```

With that, the map comes back populated on both models, and a closed object still goes native. "a dictionary field is closed, like every other object" pins the current behaviour, so it would need flipping.

It does mean any schema containing a map loses native enforcement, so I am not sure it is the right call. main's `return_output` tool is not `strict`, so these fields were not API-enforced there either. Does that match what you see, and is it a trade-off you are happy with?

cabljac (changes requested):

packages/genkit_anthropic/lib/src/schema.dart
adesinah: Matches what I see, and yes — happy with it. I reproduced both halves before changing anything:

```
200 native, map closed  -> {"name": "Ada", "scores": {}}
200 prompt fallback     -> {"name": "Ada", "scores": {"maths": 90, "art": 70}}
```

So the current behaviour is silent data loss, which is the same failure I turned down the concrete `{}` rewrite for last round — there it stringified an object, here it empties a map. Consistency says route it the same way, so `additionalProperties` with anything but `false` now makes a schema non-expressible and the whole request goes through the prompt. Your walker snippet, essentially.

On the trade-off you flagged: losing native enforcement for any schema containing a map is real, but it is not a regression. `return_output` was never `strict`, so those fields were not enforced on main either — the difference is only that main returned the data and this PR was dropping it. Given the choice between an unenforced correct map and an enforced empty one, the first is plainly better, and a caller who wants enforcement can model the field concretely.

Two wire tests either way — an open map goes to the prompt, an explicitly closed object still goes native — and a live one asserting `scores["maths"] == 90`. `a dictionary field is closed, like every other object` is flipped, as you expected. README now lists both non-expressible shapes together with the reason.

145 unit tests, analyze and format clean; live green but for the two pre-existing `claude-3-5-haiku-latest` simulation tests.

cabljac (approved):

PR #403, opened 2026-08-27
fix(a2ui): reconstruct prior surfaces as a2ui blocks in history

pavelgj:
Convert inbound a2ui surface envelopes back into canonical `a2ui` fenced blocks instead of summarizing them to a `[rendered UI surface]` sentinel. Replaying the model's own UI output in its original format reinforces correct behavior; the sentinel taught the model to emit that literal string in place of a real block.

Concrete surface ids are rewritten to the placeholder so the parser mints a fresh id per turn, preventing the model from reusing and overwriting a prior surface. Client-synthesized action envelopes remain short text summaries, and consecutive surface envelopes are grouped into one block.

### DIFF
--- a/packages/genkit_a2ui/lib/src/a2ui_middleware.dart
+++ b/packages/genkit_a2ui/lib/src/a2ui_middleware.dart
@@ -434,46 +434,87 @@ List<Part> _partsFromSegments(List<ParseSegment> segments) {
   return out;
 }
 
-/// Summarizes a list of a2ui envelopes / actions into a short text string.
+/// Converts a list of a2ui envelopes from an inbound message part back into
+/// model-readable text - the inverse of the outbound block-to-part transform.
+///
+/// The two envelope kinds are handled differently on purpose:
+///
+/// - Assistant-authored surface envelopes (`createSurface`, `updateComponents`,
+///   `updateDataModel`, `deleteSurface`) are reconstructed as the canonical
+///   `a2ui` fenced block the model originally emitted. Replaying a prior turn's
+///   surface as history therefore shows the model its own UI output in the exact
+///   format it is asked to produce, reinforcing correct behavior. (Summarizing
+///   it to a sentinel like `[rendered UI surface]` instead taught the model to
+///   emit that literal string in place of a real block.)
+/// - Client-synthesized `action` envelopes never had a block form, so they
+///   become a short text summary the model can reason about.
+///
+/// Consecutive surface envelopes are grouped into a single block (one surface is
+/// usually several envelopes: create + update(s)). Unknown envelope shapes are
+/// dropped; the caller keeps the message non-empty when everything drops.
 String _summarizeA2uiPart(List<A2uiEnvelope> envelopes) {
-  final lines = <String>[];
+  final out = <String>[];
+  final pendingSurface = <A2uiEnvelope>[];
+
+  void flushSurface() {
+    if (pendingSurface.isEmpty) return;
+    // Rewrite concrete surface ids back to the placeholder the model originally
+    // wrote (the middleware swapped in a real id on the way out). Replaying the
+    // real id would let the model copy it into a fresh `createSurface`, and the
+    // parser passes an explicit pre-existing id through unchanged - so a brand
+    // new answer would reuse (and overwrite, in place) the prior surface instead
+    // of rendering a new one. Using the placeholder makes the parser mint a
+    // fresh id per turn, so each turn is a distinct surface.
+    final placeheld = pendingSurface
+        .map(_withPlaceholderSurfaceId)
+        .toList(growable: false);
+    out.add('```a2ui\n${JsonEncoder.withIndent('  ').convert(placeheld)}\n```');
+    pendingSurface.clear();
+  }
+
   for (final env in envelopes) {
     final action = env['action'];
     if (action is Map) {
+      // Emit any buffered surface block before the action, preserving order.
+      flushSurface();
       final name = action['name'];
       final surfaceId = action['surfaceId'];
       final context = action['context'];
       final ctx = (context is Map && context.isNotEmpty)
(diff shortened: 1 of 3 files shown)


packages/genkit_a2ui/lib/src/a2ui_middleware.dart
```diff
@@ -434,47 +434,91 @@ List<Part> _partsFromSegments(List<ParseSegment> segments) {
...
+A2uiEnvelope _withPlaceholderSurfaceId(A2uiEnvelope env) {
+  final copy = <String, dynamic>{...env};
+  for (final key in const [
+    'createSurface',
+    'updateComponents',
+    'updateDataModel',
+    'deleteSurface',
+  ]) {
+    final payload = copy[key];
+    if (payload is Map && payload['surfaceId'] is String) {
```
apascal07: The rewrite is unconditional, so it also erases ids the model targeted deliberately.

Two kinds of id reach here: one the middleware minted (the model wrote `SURFACE_ID` and `_normalizeEnvelope` swapped in a real id), and one the model wrote explicitly to address a live surface, which `_finalizeBlock` case 2 treats as a genuine incremental update. Only the first should go back to the placeholder.

Concretely: turn 1 mints `abc-123`. The user taps a button, so the next request carries the action envelope, summarized as `[UI action "submit" on surface abc-123]`, plus the surface part replayed with `surfaceId: "SURFACE_ID"`. This change exists so the model copies the block it sees, so it emits `updateComponents` on `SURFACE_ID`. That mints a fresh id, takes the `isFreshRender` branch, synthesizes a `createSurface`, and enforces `_validateRoot`. The user gets a second surface instead of an update to the one they just touched, and a partial subtree update carrying no `root` component is rejected and dropped under `warn`.

History also now shows the model two handles for one surface: `abc-123` in the action line and `SURFACE_ID` in the block. Whichever is meant to win, only that one should appear.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
```diff
@@ -434,47 +434,91 @@ List<Part> _partsFromSegments(List<ParseSegment> segments) {
...
+    // wrote (the middleware swapped in a real id on the way out). Replaying the
+    // real id would let the model copy it into a fresh `createSurface`, and the
+    // parser passes an explicit pre-existing id through unchanged - so a brand
+    // new answer would reuse (and overwrite, in place) the prior surface instead
+    // of rendering a new one. Using the placeholder makes the parser mint a
+    // fresh id per turn, so each turn is a distinct surface.
+    final placeheld = pendingSurface
+        .map(_withPlaceholderSurfaceId)
+        .toList(growable: false);
+    out.add('```a2ui\n${JsonEncoder.withIndent('  ').convert(placeheld)}\n```');
```
apascal07: `withIndent` roughly doubles the tokens against `jsonEncode` and no model reads the result differently.

The `model` hook runs once per raw model call, so a three-step tool loop re-serializes the full history of surfaces three times in a single turn, and nothing caps how many accumulate. Ten turns of few-hundred-component surfaces means all ten replayed on every call. One recent surface reinforces the format as well as ten do.

apascal07: The fence is fixed at three backticks and nothing checks the encoded JSON for backtick runs. A `Text` component whose text is a code sample closes the block partway through the component list and spills the remainder as prose, so the model is shown a malformed example of the exact format it is being asked to reproduce.

apascal07 (approved):
`packages/genkit_a2ui/README.md:330` still says inbound a2ui parts "are summarized into plain text before the underlying model sees them". That now holds only for `action` envelopes.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
pavelgj: Good catch, fixed. Reworked so the guard lives at the true chokepoint instead of erasing ids in history. The parser now forces a fresh id onto every envelope in a `createSurface` block (`createSurface` means "new surface" by definition), so a copied id can never reuse/overwrite a prior surface. That let me drop the placeholder rewrite entirely: history keeps the real surface ids verbatim on both the reconstructed block and the `[UI action ... on surface <id>]` line, so the model sees one consistent handle per surface and can correlate an action with the surface it targeted (important with multiple surfaces on screen). Your `updateComponents` incremental-update path is untouched (a bare update to an explicit existing id still passes through as case 2). Added a regression test ("a new render never reuses a surface id copied from history") in Dart + JS.

packages/genkit_a2ui/lib/src/a2ui_middleware.dart
pavelgj: Dropped `withIndent` for compact `jsonEncode` (JS already used compact `JSON.stringify`), so the token doubling is gone. On the "no cap / re-serialized N times" part: I intentionally did not add a surface-count or history-window cap here. Deciding which past surfaces are still relevant, and how far back to look, is a context-management concern that a dedicated context-compression middleware should own; this middleware does not know enough to make that call safely. Happy to revisit if you feel it belongs here.

pavelgj: Addressed in practice, with the residual documented. Switching to compact `jsonEncode` collapses each reconstructed surface to a single JSON line, and JSON escapes any interior newline as `\n`, so a `Text` value containing a fenced code sample can never place a line-leading ``` inside the block. Combined with the parser's close fence being line-anchored (`(^|\n)[ \t]*```), an inner backtick run can no longer prematurely close the block. The block text can still contain ``` mid-line; a fully robust emitter would use a variable-length (CommonMark-style) fence, but that also requires the parser's fixed three-backtick open fence to become count-aware or it would not re-recognize a copied 4-backtick fence, so I left that as a documented follow-up (comment added in both Dart and JS).

Issue #360, opened 2026-08-24
fix(genkit_openai): init() requires network + API key at startup


cabljac:
`init()` calls `models.list()` on the default host and rethrows any failure (`openai_plugin.dart:78-102`), and `_resolveClientConfig` throws without an API key. Plugin init is lazy (`registry.dart:32-40`), so this fires on the first lookup of any model in the plugin's namespace, or on any Dev-UI action listing (`/api/__health` and `/api/actions` call `registry.listActions()`), even for `customModels`-only setups on the default host. JS/Go initialize offline and resolve lazily, with network only in their listActions paths; both fall back to `OPENAI_API_KEY` env for the key. Also: `list()` runs against custom `baseUrl` hosts unconditionally and omits `customModels`.

PR #405, opened 2026-09-01
feat(genkit_openai): add text-to-speech support

### LINKED ISSUE #168: Support TTS on openai

adesinah:
This PR registers `tts-1`, `tts-1-hd` and `gpt-4o-mini-tts` as speech models, referenced via `openAI.speechModel()`. It also returns audio as a `MediaPart` holding a base64 data URL, matching the JS compat-oai plugin.  The SDK's SpeechRequest caps voice at six legacy values and has no instructions field, so the request body subclasses it and overrides toJson(). This keeps every call on the SDK transport (auth, retries, error mapping) while allowing free-form voices and instructions.

Closes #168


cabljac (commented):
Wire format and MIME map look right. The problems are routing and input selection, left inline.

Cross-cutting, since it isn't a line: this is the seam #422 redraws. I merged the two locally. `resolve()` is fine and TTS by name still works, but `init()` misroutes custom speech models, because `Registry.lookupAction` prefers eager registrations and #422 registers them as chat. #422's `list()` also gives speech ids the chat options schema. Four files conflict. Probably worth agreeing the modality-routing shape once across both PRs rather than resolving it twice.

Dart tests are red at head, though that looks like the blank-prompt case rather than anything here.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -353,6 +419,118 @@ class OpenAIPlugin extends GenkitPlugin {
...
+  /// JS plugin's `toTTSRequest`.
+  String _speechInputText(ModelRequest request) {
+    if (request.messages.isEmpty) {
+      throw GenkitException(
+        'Speech models require a prompt, but no messages were provided.',
+        status: StatusCodes.INVALID_ARGUMENT,
+      );
+    }
+
+    final text = request.messages.first.text;
```
cabljac, line 524: `messages.first` picks the system message, since core appends `system` before `messages` and `prompt` (`generate.dart:684-702`):

```dart
ai.generate(model: openAI.speechModel('tts-1'),
            system: 'You are a narrator.', prompt: 'Read this aloud.');
// -> {"model":"tts-1","input":"You are a narrator.","voice":"alloy"}
```

The prompt never reaches the API, and in a multi-turn conversation only the first message survives. JS has the same bug (`audio.ts:126`), so taking the first non-system message fixes ours without diverging.

packages/genkit_openai/lib/src/speech.dart
```diff
@@ -0,0 +1,123 @@
...
+/// signal available for those models.
+bool declaresMediaOutput(ModelInfo? info) {
+  final output = info?.supports?['output'];
+  return output is List && output.contains('media');
+}
+
+/// Returns true when [modelId] rejects the `speed` parameter.
+bool speechModelRejectsSpeed(String modelId) {
+  return modelId.toLowerCase().contains('gpt-4o-mini-tts');
+}
```
cabljac, line 95: `gpt-4o-mini-tts` accepts `speed`. I ran it:

```
gpt-4o-mini-tts + speed 2.0  -> 200, 12288 bytes
gpt-4o-mini-tts, no speed    -> 200, 14592 bytes
```

Shorter audio, so it took effect. The OpenAPI spec gives `speed` a 0.25-4.0 range with no model restriction, and states restrictions where they exist (`instructions` "does not work with tts-1 or tts-1-hd").

This function and its three call-out sites can go: `openai_plugin.dart:461`, `speech.dart:42`, README:237, README:277, plus the testapp comment.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -353,6 +419,118 @@ class OpenAIPlugin extends GenkitPlugin {
...
+            input: input,
+            voiceName: options.voice ?? speech.defaultSpeechVoice,
+            instructions: options.instructions,
+            responseFormat: options.responseFormat == null
+                ? null
+                : sdk.SpeechResponseFormat.fromJson(format),
+            // gpt-4o-mini-tts rejects `speed` outright.
+            speed: speech.speechModelRejectsSpeed(resolvedModel)
+                ? null
+                : options.speed,
```
cabljac, line 464: Drops a parameter the model accepts. See the note on `speechModelRejectsSpeed`.

packages/genkit_openai/lib/src/speech.dart
```diff
@@ -0,0 +1,123 @@
...
+  String? get voice;
+
+  /// Tone and delivery guidance, e.g. 'Speak in a calm, warm tone.'
+  ///
+  /// Only honored by `gpt-4o-mini-tts`; the `tts-1` family ignores it.
+  String? get instructions;
+
+  /// Playback speed multiplier (0.25 - 4.0).
+  ///
+  /// Not supported by `gpt-4o-mini-tts`, which rejects the field outright.
```
cabljac, line 42: `gpt-4o-mini-tts` does not reject `speed`, verified against the live API. This line needs removing along with the function below.

packages/genkit_openai/lib/src/speech.dart
```diff
@@ -0,0 +1,123 @@
...
+/// Returns true when [info] declares audio output.
+///
+/// Speech models on OpenAI-compatible providers are not always named
+/// `*tts*`, so a caller registering one through `CustomModelDefinition` says
+/// so with `supports: {'output': ['media']}`. That declaration is the only
+/// signal available for those models.
+bool declaresMediaOutput(ModelInfo? info) {
+  final output = info?.supports?['output'];
+  return output is List && output.contains('media');
+}
```
cabljac, line 90: Membership, not exclusivity. A chat model declaring `output: ['text','json','media']` hits only `/audio/speech`; `response.text` comes back empty. `supports.output` is free-form in core, so that declaration is valid.

Requiring an exclusive `['media']` fixes it and breaks no test, though it still catches custom image models. An explicit modality field would settle it properly.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -187,6 +223,12 @@ class OpenAIPlugin extends GenkitPlugin {
         );
       }
 
+      for (final modelId in speech.knownSpeechModels) {
+        if (listed.add(modelId)) {
+          modelMetadataList.add(_speechModelMetadata(modelId));
+        }
+      }
```
cabljac, line 230: Appends `knownSpeechModels` unconditionally, unlike `init()`, which gates on a null `baseUrl`. Groq-shaped instance:

```
list() -> [groq/llama-3.3-70b-versatile, groq/tts-1, groq/tts-1-hd, groq/gpt-4o-mini-tts]
```

Three ids that host 404s on, offered to the Dev UI. The same guard fixes it.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -353,6 +419,118 @@ class OpenAIPlugin extends GenkitPlugin {
...
+          final resolvedModel = options.version ?? modelName;
+
+          final body = _SpeechRequestBody(
+            model: resolvedModel,
+            input: input,
+            voiceName: options.voice ?? speech.defaultSpeechVoice,
+            instructions: options.instructions,
+            responseFormat: options.responseFormat == null
+                ? null
+                : sdk.SpeechResponseFormat.fromJson(format),
```
cabljac, line 460: `responseFormat: 'ogg'` throws a `FormatException` inside the SDK and surfaces as `INTERNAL`/13 rather than a 400-class error. The `enumValues` annotation is decorative at runtime, as is `minimum`/`maximum` on `speed` (`speed: 99.0` reaches the wire). Unknown keys also drop silently, so `OpenAIChatOptions(temperature: 0.7)` against a tts id loses `temperature`.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -353,6 +419,118 @@ class OpenAIPlugin extends GenkitPlugin {
...
+              role: Role.model,
+              content: [
+                MediaPart(
+                  media: Media(
+                    contentType: contentType,
+                    url: 'data:$contentType;base64,${base64Encode(bytes)}',
+                  ),
+                ),
+              ],
+            ),
```
cabljac, line 483: Treats any 200 body as audio and hardcodes `finishReason: stop`, so a zero-byte response becomes `data:audio/mpeg;base64,` reported as success, and a caller following the README idiom writes an empty `.mp3`.

Content-type sniffing isn't reachable here: OpenAI does set a distinct type per format (I checked all six), but `openai_dart` returns bytes with no header access. Length or magic bytes is what's available.

packages/genkit_openai/lib/src/speech.dart
```diff
@@ -0,0 +1,123 @@
...
+
+/// Returns true when [modelId] is a text-to-speech model.
+///
+/// [getModelType] lumps every audio model together, so the extra `tts` check
+/// is what separates speech synthesis from transcription (`whisper-1`,
+/// `gpt-4o-transcribe`), realtime, and audio-in chat models.
+bool isSpeechModel(String modelId) {
+  return getModelType(modelId) == 'audio' &&
+      modelId.toLowerCase().contains('tts');
+}
```
cabljac, line 79: nit: effectively just `contains('tts')`, since `getModelType` already returns `audio` for those. The doc comment credits the other conjunct. Harmless now, but #406 splits the same bucket.

packages/genkit_openai/README.md
```diff
@@ -213,6 +213,47 @@ final response = await ai.generate(
...
+    instructions: 'Speak in a calm, warm tone.',
+  ),
+);
+
+final media = response.media!;            // contentType: audio/mpeg
+final bytes = base64Decode(media.url.split(',').last);
+await File('speech.mp3').writeAsBytes(bytes);
+```
+
+`tts-1` and `tts-1-hd` also accept `speed`. `gpt-4o-mini-tts` rejects it, so the
```
cabljac, line 237: `gpt-4o-mini-tts` does not reject `speed`, verified live.

packages/genkit_openai/README.md
```diff
@@ -229,6 +270,14 @@ The `OpenAIChatOptions` class supports the following options:
 - `visualDetailLevel` (String?, 'auto'|'low'|'high') - Visual detail level for images
 - `version` (String?) - Model version override
 
+The `OpenAISpeechOptions` class supports the following options:
+
+- `voice` (String?) - Voice name, e.g. `'alloy'`, `'sage'`, `'coral'` (defaults to `'alloy'`). Free-form, so new OpenAI voices work without a plugin update
+- `instructions` (String?) - Tone and delivery guidance (`gpt-4o-mini-tts` only)
+- `speed` (double?, 0.25-4.0) - Playback speed (not supported by `gpt-4o-mini-tts`)
```
cabljac, line 277: Same here.

packages/genkit_openai/lib/src/speech.dart
adesinah: Removed `speechModelRejectsSpeed` and every call site — thanks for running it. The live suite now asserts 2x comes back shorter than 1x, so the claim is tested rather than assumed.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Removed `speechModelRejectsSpeed` and every call site — thanks for running it. The live suite now asserts 2x comes back shorter than 1x, so the claim is tested rather than assumed.

packages/genkit_openai/lib/src/speech.dart
adesinah: Removed `speechModelRejectsSpeed` and every call site — thanks for running it. The live suite now asserts 2x comes back shorter than 1x, so the claim is tested rather than assumed.

packages/genkit_openai/README.md
adesinah: Removed `speechModelRejectsSpeed` and every call site — thanks for running it. The live suite now asserts 2x comes back shorter than 1x, so the claim is tested rather than assumed.

packages/genkit_openai/README.md
adesinah: Removed `speechModelRejectsSpeed` and every call site — thanks for running it. The live suite now asserts 2x comes back shorter than 1x, so the claim is tested rather than assumed.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Fixed — takes the last non-system message now, so a system instruction is no longer what gets read aloud and a multi-turn conversation speaks its newest turn. Two tests cover both.

packages/genkit_openai/lib/src/speech.dart
adesinah: Now requires exclusivity — a single-element `['media']`. Agreed an explicit modality field is the real fix; this at least stops a chat model declaring `['text','json','media']` from being routed to /audio/speech.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Fixed — the speech ids are behind the same `baseUrl == null` guard as the chat catalog now, for the reason you gave.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Added real validation for both: an unknown container and an out-of-range speed are now INVALID_ARGUMENT before the request, rather than a FormatException surfacing as INTERNAL. Unknown keys dropping silently is schemantic's behaviour across every options class here, so I've left that one alone.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Empty body is now INTERNAL rather than a success carrying zero bytes. Agreed on sniffing — openai_dart hands back bytes with no headers, so length is what's available.

packages/genkit_openai/lib/src/speech.dart
adesinah: Reworded — the doc now says the name is the whole test and that the getModelType half is a guard for when #406 splits the audio bucket.

adesinah:
Rebased onto main, which settles the cross-cutting point: #422 is merged, so the modality routing is resolved once here rather than twice.

Four files conflicted as you predicted. The decisive one was `init()` — this branch did network I/O at startup, which main has since fixed (#360), so I took main's shape and added speech routing for custom models only. That also fixes the misrouting you described: a custom speech model is now registered as speech eagerly, so the registry's preference for eager registrations no longer beats `resolve()`. `list()` gives speech ids the speech options schema, behind the same `baseUrl` guard as the chat catalog.

The red tests at head were the blank-prompt case, as you thought — `generate()` reports request errors on the response since #413 rather than throwing. Fixed along with two other stale expectations.

All twelve inline points addressed; replies inline. 183 tests pass.

adesinah:
CI is fixed — `speech.g.dart` was stale after the `speed` dartdoc change. Schemantic copies doc comments into the generated class verbatim, and `OpenAISpeechOptions` there is the public type, so the IDE tooltip was telling users the opposite of the README.

Four other things while I was in here:

- `list()` could emit a custom speech model twice — once with the chat options schema — when the host's `/models` also advertised it. Discovery doesn't know about a caller's `models:` declaration, so the two sets overlapped. Deduped, with a test.
- `opus` was labelled `audio/opus` and `pcm` `audio/L16`. OpenAI returns Ogg-encapsulated Opus, and `audio/L16` is defined big-endian at 8 kHz while OpenAI sends little-endian 24 kHz — so a consumer decoding to spec got static. Now `audio/ogg` and `audio/pcm`. Worth a look since the contentType is derived from the requested format, not the response.
- A test asserting `declaresMediaOutput` exclusivity was still named after the `speed` behaviour it used to check.

One I've left, and it's worth your call: routing on `output: ['media']` means anyone already registering a media-producing chat model on a compat host silently moves from `/chat/completions` to `/audio/speech`. Nothing in the repo does it, but the capability flag is generic. That's the explicit modality field you mentioned — happy to add an opt-in here instead if you'd rather not ship the overload.

cabljac (approved):
lgtm

PR #406, opened 2026-09-01
feat(genkit_openai): support whisper on OpenAI stt

adesinah:
## Summary

Adds OpenAI speech-to-text to `genkit_openai`, the input-side counterpart to the
TTS support in #405. `whisper-1`, `gpt-4o-transcribe` and `gpt-4o-mini-transcribe`
were classified as `'audio'` by `getModelType()` but skipped or routed to chat by
every registration path, so audio could not be transcribed at all.

```dart
final response = await ai.generate(
  model: openAI.transcriptionModel('whisper-1'),
  promptParts: [MediaPart(media: recording)],
  config: OpenAITranscriptionOptions(language: 'en'),
);
print(response.text);
```

`whisper-1` also accepts `translate: true`, routing to `/audio/translations` for
English output from audio in any language.

`response.text` is the transcript; the decoded response — segments, timestamps,
logprobs — is on `response.raw`, so nothing `verbose_json` or `include` was asked
for is dropped. Asking for `outputFormat: 'json'` sends the JSON object through
whole instead, so `response.output` parses rather than coming back null.

A model is routed to transcription by its name (`*whisper*`, `*transcribe*`) or,
for a compatible provider that names things differently, by an explicit
`CustomModelDefinition(kind: OpenAIModelKind.transcription)`. Capability metadata
deliberately does not decide it: `supports: {'media': true}` describes a vision
chat model as readily as a transcription one, and reading it as a signal sent
plain text prompts on registered chat models to `/audio/transcriptions`. Speech
keeps its metadata signal, since `output: ['media']` is unambiguous.

> [!NOTE]
> Stacked on #405 — this PR targets `feat/openai-tts-support`, so its diff is STT
> only. Review #405 first. GitHub will retarget this to `main` once #405 merges.

## Why the request is hand-built

Unlike TTS, this path does not go through `openai_dart`. Two blockers:

- `TranscriptionRequest` has no `toJson()` and the multipart builder is private,
  so the `toJson()` subclass trick used for TTS is unavailable.
- `transcriptions.create()` unconditionally runs `jsonDecode(response.body)`, so
  `response_format: text|srt|vtt` throws a `FormatException` instead of returning
  subtitles.

Hand-building the upload also unlocks `chunking_strategy` (needed for long audio
on the `gpt-4o-transcribe` family), `include[]`, and a future `stream: true`.
Auth, error mapping and `baseUrl`/`headers` are resolved through the existing
`_resolveClientConfig()`, so behaviour matches the SDK paths.

## Parity with the JS plugin

Ports `compat-oai/src/audio.ts`, `src/openai/whisper.ts` and `src/translate.ts`.
Two deliberate differences: a blank `prompt` is omitted rather than sent as `''`,
and `response_format` defaults to `json` rather than `text`.

## Testing

35 unit/wire tests and 6 live tests (`skip`-gated on `OPENAI_API_KEY`).

`MockClient` never encodes a real multipart body, so the wiring was also checked
against a local `HttpServer`. Two things a mock would have missed:

- `audio/x-m4a` → `input.m4a` — a generic `split('/').last` would send `x-m4a`,
  which OpenAI rejects
- `timestamp_granularities[]` sent as genuinely repeated fields — the encoding
  `openai_dart` gets wrong with indexed `[$i]` keys, which its own source comment
  admits is unverified

Live tests cover TTS→STT round trip, Spanish→English translation, SRT subtitles
and `verbose_json` with timestamp granularities — all passing against the real
API. Two further tests exercising `chunking_strategy` (string and object form) on
the `gpt-4o-transcribe` family are included but have not yet been run.

No audio fixtures are committed — the live tests generate their input with `tts-1`.

### DIFF
--- a/packages/genkit_openai/lib/src/openai_plugin.dart
+++ b/packages/genkit_openai/lib/src/openai_plugin.dart
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import 'dart:convert';
+import 'dart:typed_data';
 
 import 'package:genkit/plugin.dart';
 import 'package:http/http.dart' as http;
@@ -21,6 +22,7 @@ import 'package:openai_dart/openai_dart.dart' as sdk;
 import '../genkit_openai.dart';
 import 'chat.dart' as chat;
 import 'speech.dart' as speech;
+import 'transcription.dart' as transcription;
 
 /// Core Genkit plugin implementation for OpenAI-compatible APIs.
 ///
@@ -28,6 +30,9 @@ import 'speech.dart' as speech;
 /// [baseUrl] is set) and registers them in the Genkit action registry.
 /// Additional models can be provided via [customModels].
 class OpenAIPlugin extends GenkitPlugin {
+  /// Base URL used when the plugin was not given one.
+  static const String _defaultBaseUrl = 'https://api.openai.com/v1';
+
   final String _pluginName;
 
   @override
@@ -100,6 +105,17 @@ class OpenAIPlugin extends GenkitPlugin {
             continue;
           }
 
+          if (transcription.isTranscriptionModel(modelId)) {
+            registered.add(modelId);
+            actions.add(
+              _createTranscriptionModel(
+                modelId,
+                transcription.transcriptionModelInfo(modelId),
+              ),
+            );
+            continue;
+          }
+
           final modelType = getModelType(modelId);
 
           if (modelType != 'chat' && modelType != 'unknown') {
@@ -119,6 +135,17 @@ class OpenAIPlugin extends GenkitPlugin {
             );
           }
         }
+
+        for (final modelId in transcription.knownTranscriptionModels) {
+          if (registered.add(modelId)) {
+            actions.add(
+              _createTranscriptionModel(
+                modelId,
+                transcription.transcriptionModelInfo(modelId),
+              ),
+            );
+          }
+        }
       } catch (e) {
         throw GenkitException(
           'Error fetching available models from $_pluginName: $e',
@@ -132,6 +159,9 @@ class OpenAIPlugin extends GenkitPlugin {
       if (speech.isSpeechModel(model.name) ||
           speech.declaresMediaOutput(model.info)) {
         actions.add(_createSpeechModel(model.name, model.info));
+      } else if (transcription.isTranscriptionModel(model.name) ||
+          transcription.declaresMediaInput(model.info)) {
+        actions.add(_createTranscriptionModel(model.name, model.info));
       } else {
         actions.add(_createModel(model.name, model.info));
       }
@@ -209,6 +239,12 @@ class OpenAIPlugin extends GenkitPlugin {
           continue;
         }
 
+        if (transcription.isTranscriptionModel(modelId)) {
+          listed.add(modelId);
+          modelMetadataList.add(_transcriptionModelMetadata(modelId));
+          continue;
+        }
+
         final modelType = getModelType(modelId);
         if (modelType != 'chat' && modelType != 'unknown') {
           continue;
@@ -229,6 +265,12 @@ class OpenAIPlugin extends GenkitPlugin {
         }
       }
 
+      for (final modelId in transcription.knownTranscriptionModels) {
+        if (listed.add(modelId)) {
+          modelMetadataList.add(_transcriptionModelMetadata(modelId));
+        }
+      }
+
       return modelMetadataList;
     } catch (e, stackTrace) {
       throw GenkitException(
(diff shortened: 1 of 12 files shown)


packages/genkit_openai/lib/src/transcription.dart
```diff
@@ -0,0 +1,164 @@
...
+bool supportsTranslation(String modelId) {
+  return modelId.toLowerCase().contains('whisper');
+}
+
+/// Returns true when [info] declares audio input.
+///
+/// The input-side mirror of `declaresMediaOutput`: a caller registering a
+/// transcription model on an OpenAI-compatible provider whose name is not
+/// `*whisper*` or `*transcribe*` says so with `supports: {'media': true}`.
+bool declaresMediaInput(ModelInfo? info) {
```
cabljac: Could we use an explicit transcription flag here? `media: true` also describes vision chat models. Registering `gpt-4o` with `multimodalSupports` now makes a plain text prompt fail with “Transcription models require an audio media part.” I reproduced this with a mocked client.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -751,6 +797,222 @@ class OpenAIPlugin extends GenkitPlugin {
...
+
+          if (response.statusCode < 200 || response.statusCode >= 300) {
+            throw GenkitException(
+              'OpenAI API error: HTTP ${response.statusCode}',
+              status: StatusCodes.fromHttpStatus(response.statusCode),
+              details: response.body,
+            );
+          }
+
+          return ModelResponse(
```
cabljac, line 929: Looks like we lose timestamps and logprobs here: `_transcriptText` keeps only `text`, and `raw` stays null. Could we preserve the decoded response in `raw`, as the options docs promise? A mocked `verbose_json` response confirmed the timestamps disappear.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -751,6 +797,222 @@ class OpenAIPlugin extends GenkitPlugin {
...
+
+  /// Pulls the transcript out of a response body.
+  ///
+  /// `json` and `verbose_json` wrap it in an object; `text`, `srt` and `vtt`
+  /// are already the transcript.
+  String _transcriptText(String body, String format) {
+    if (format != 'json' && format != 'verbose_json') {
+      return body;
+    }
+    final decoded = jsonDecode(body);
```
cabljac: Could we keep the JSON object when the caller explicitly sets `outputFormat: 'json'`? This currently unwraps it to plain text too, so Genkit reports success but `response.output` is null. Returning just the transcript seems right for the default text case.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -751,6 +797,222 @@ class OpenAIPlugin extends GenkitPlugin {
...
+            transcription.supportsTranslation(resolvedModel);
+
+        final resolvedConfig = await _resolveClientConfig();
+        final client = httpClient ?? http.Client();
+
+        try {
+          final endpoint = translate
+              ? '/audio/translations'
+              : '/audio/transcriptions';
+          final url = Uri.parse(
```
cabljac: Could we append the endpoint to the URI’s path instead? With a base URL like `/v1?api-version=...`, this puts `/audio/transcriptions` inside the query value. A trailing slash also produces `/v1//audio/transcriptions`, which some hosts may reject. The SDK already handles both cases for chat and TTS.

packages/genkit_openai/lib/src/transcription.dart
adesinah: Agreed, and fixed. `CustomModelDefinition` takes an `OpenAIModelKind` now, and `declaresMediaInput` is gone - `media: true` was never a transcription signal. Routing runs through one `_kindOf` resolver: a declared kind wins, otherwise the name decides. The speech side still reads `info`, since `output: ['media']` does mean audio out and nothing else. Your repro is a test: gpt-4o with media support gets a chat completion for a text prompt.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Fixed. The decoded body now goes on `ModelResponse.raw`, so segments, timestamps and logprobs survive; the text part is still the transcript. `text`, `srt` and `vtt` leave `raw` unset - there is nothing in them beyond the transcript, and `raw` takes an object. Test covers a verbose_json response with segments.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Done. With `outputFormat: 'json'` the object goes through whole so Genkit has something to parse; without it the bare transcript is still what you get. Both cases have a test.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Fixed - the endpoint is appended to the base URI's path segments instead of concatenated, so a query survives and a trailing slash does not double up. Tests for `/v1?api-version=...` and `/v1/`.

cabljac (approved):
lgtm

PR #410, opened 2026-09-01
feat(genkit)!: pass request context to agent session-store operations

JanKubicek:
## Summary

The server-side agent runtime (`defineAgent` / `defineCustomAgent`) dropped the ambient request context when calling `SessionStore`, so context-scoped stores like `FirestoreSessionStore` with `snapshotPathPrefix` always resolved the default `"global"` prefix instead of the per-tenant one derived from the call context.

This change forwards the action context into every session-store operation so multi-tenant isolation works as documented.

## Changes

- `packages/genkit/lib/src/ai/agents/agent.dart`
  - Pass action context to session resolution, snapshot saves (completed/failed/recovery), detached `pending` writes, heartbeats, and `onSnapshotStateChange` subscriptions.
  - `getSnapshotDataAction` / `abortAgentAction` now forward their action context to the store.
  - In-process snapshot/abort operations route through the registered actions, so they inherit ambient action context like `send()` does (existing `Action.run` zone behavior) instead of bypassing it.
- `packages/genkit/lib/src/ai/agents/agent_core.dart`
  - `AgentTransport.getSnapshot` / `abort` accept an optional `context`.
  - `DetachedTask` retains the context passed to `detach()` so `poll()` / `wait()` / `abort()` stay tenant-scoped after the originating action completes.
- `packages/genkit/lib/src/ai/agents/remote_agent.dart`
  - Snapshot/abort reject client-supplied context, matching the existing turn behavior (remote context is derived server-side from request headers).

## Breaking change

`AgentTransport.getSnapshot` and `abort` gained an optional `context` parameter. Third-party `AgentTransport` subclasses must update their overrides to accept it.

## Tests

- Context reaches all session-store operations on send/resume/abort/get-snapshot.
- Ambient action context is inherited by nested in-process calls.
- Detached tasks retain explicit context across polling and abort.
- Remote agents reject non-empty client-supplied context for snapshot and abort.

## Note for reviewers

This PR was mostly written by GPT-5.6 "Sol" (assisted with the codebase exploration). It works for my use case and the changes seem correct on inspection, but I'm not an expert in this codebase, so a thorough review would be appreciated.

### DIFF
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -165,6 +165,7 @@ void _requireStore(SessionStore? store, String operation, String agentName) {
 Future<String?> _abortSnapshotInStore(
   SessionStore store,
   String snapshotId,
+  Map<String, dynamic>? context,
 ) async {
   String? previousStatus;
   await store.saveSnapshot(snapshotId, (current) {
@@ -178,7 +179,7 @@ Future<String?> _abortSnapshotInStore(
 
     current.status = SnapshotStatus.aborted;
     return current;
-  });
+  }, context: context);
   return previousStatus;
 }
 
@@ -272,6 +273,7 @@ class SessionRunner<State> {
     this.inputCh, {
     SessionSnapshot? lastSnapshot,
     SessionStore? store,
+    this.context,
     this.cancel,
     this.onEndTurn,
     this.onDetach,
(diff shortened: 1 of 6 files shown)


pavelgj:
Thanks for the PR. This is a good catch. Overall looks good. Just checking a few things.

pavelgj (approved):

pavelgj:
released https://github.com/genkit-ai/genkit-dart/releases/tag/genkit-v0.16.1
thanks for the fix again

PR #411, opened 2026-09-03
feat(agents): `aborting` snapshot status + rerunnable aborted snapshots (Go parity)

pavelgj:
Stacked on #397 (cancellation). Brings Dart's agent abort lifecycle and resume rules in line with the Go reference implementation (`go/ai/exp/agent.go`), which is the source of truth for the shared agent spec.

### a. `aborting` snapshot status
Adds `aborting` to `SnapshotStatus` and adopts Go's two-write abort protocol so Dart both *emits* and *reads* the mid-flight window correctly.

Before, Dart settled an abort in a single write (`pending -> aborted`). Now it matches Go:

1. **Flip** (`_abortSnapshotInStore`): abort turns a `pending` row into `aborting`, stopping the work without stamping state. It leaves `heartbeatAt` where the worker left it, so a dead worker's row reads as `expired` at once rather than after a heartbeat timeout.
2. **Finalize** (`SessionRunner.maybeSnapshot`): the worker observes the flip, cancels its turn, and settles the row to `aborted` *with* the last-good state. The abort-aware mutator deliberately treats `aborting` as non-terminal so this finalize is allowed through.

Supporting changes for parity:
- `abort()` now returns the *resulting* status (`pending -> aborting`, already terminal -> existing status, missing -> null), matching Go's `Agent.Abort`. **This is a behavior change** from the prior "return previous status".
- `_isHeartbeatExpired` treats a stale `pending` *or* `aborting` beat as `expired` on read (both are in-flight, still-heartbeating states).
- `_windDownHeartbeatBudget` (5 min) bounds how long a winding-down worker keeps beating, so a wedged abort still goes stale and reads as `expired` instead of winding down forever.

### b. Rerunnable aborted snapshots
Resume is no longer restricted to `completed`. The resume guard now also accepts terminal snapshots that carry state, so an `aborted` run can be rerun: whether to continue from a run that was stopped is the caller's judgement, not the framework's. This pairs with the cancellation work in #397, where a cancelled generate resolves to a response with `finishReason: aborted` and the last-good history, which persists as a resumable `aborted` snapshot.
- `_assertResumable` accepts `completed`/`failed`/`aborted` rows that carry state (matching Go's `resumeSessionFrom`) and rejects `pending`/`aborting` with heartbeat-aware messaging (live beat: wait/retry; stale beat: resume from parent).
- `_isResumableLeaf` drives the `sessionId` parent-walk to the last-good leaf.

Note: the guard accepts `failed` for Go parity, but *producing* a useful `failed` snapshot (a model/tool error resolving to a response with `finishReason: failed`, `error`, and the last-good history, the way cancellation already does for `aborted`) is a follow-up. This PR only lands the wire types for it.

### Supporting: `RuntimeError` + `FinishReason.failed` + `GenerateResponse.error`
Regenerated `types.dart` / `types.g.dart` to pull in the error wire shape the updated schema now references directly from `ModelResponse.error` / `GenerateResponse.error` (previously only aliased via `AgentErrorInfo`), plus `FinishReason.failed`. These are the groundwork for the follow-up that makes a model/tool error resolve gracefully (mirroring cancellation's `aborted` response). Allowlisted `RuntimeError` in the schema generator so it compiles. (This also pulls in `metadataOnly` on `GetSnapshotDataInput` from the canonical schema; wiring it up is out of scope here.)

## Reading / handling `aborting`

`aborting` is transient and non-terminal: keep polling until a terminal status. `DetachedTask.wait()`/`poll()` already do this (`_terminalStatuses` excludes `aborting`).

```dart
final task = await chat.detach(text: 'do the long thing');
final status = await task.abort(); // may be `aborting` while it winds down
if (status?.value == SnapshotStatus.aborting.value) {
  final settled = await task.wait(); // ...ride the wind-down to the terminal row
  print(settled.status?.value); // 'aborted'
}
```

Reading a snapshot directly, branch on the status and re-read the non-terminal ones (a stale heartbeat surfaces as `expired`):

```dart
final snap = await agent.getSnapshotData(snapshotId: id);
switch (snap?.status?.value) {
  case 'completed':
  case 'aborted':
    break; // terminal - use as-is (and rerunnable, since it carries state)
  case 'pending':
  case 'aborting':
    break; // in-flight - poll again shortly
  case 'expired':
    break; // worker died mid-run - fall back to the last-good snapshot
}
```

## Tests
- `getSnapshotData reports a stale aborting snapshot as expired` (cross-runtime read: a Go server's stale `aborting` beat reads as `expired`).
- Updated the conformance harness (`_executeAbort`) to read the pre-abort status before aborting, since `abort()` now returns the *resulting* status - mirroring the Go and JS harnesses. Full conformance suite green.

## Note for reviewers
`abort()` returning the resulting status (vs the previous status) is an intentional Go-parity choice and diverges from JS's `agent.abort()`. Worth a CHANGELOG callout on release.

### DIFF
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -203,15 +203,21 @@ const Duration _defaultHeartbeatInterval = Duration(seconds: 30);
 /// [_defaultHeartbeatInterval] so a single missed beat does not trip expiry.
 const Duration _defaultHeartbeatTimeout = Duration(seconds: 60);
 
-/// Returns `true` when [snapshot] is a `pending` (detached, in-flight) snapshot
-/// whose heartbeat is older than [timeout] - i.e. its background worker is
-/// presumed dead. A pending snapshot that has not yet written a first heartbeat
-/// is not considered expired (the beat may simply not have fired yet).
+/// Returns `true` when [snapshot] is a detached, in-flight snapshot whose
+/// heartbeat is older than [timeout] - i.e. its background worker is presumed
+/// dead. Both `pending` and `aborting` are in-flight, still-heartbeating states
+/// (a worker keeps beating while it winds down an abort), so a stale beat in
+/// either reads as `expired`; this mirrors the shared spec, where a `pending`
+/// *or* `aborting` row with a stale beat surfaces as `expired`. A snapshot that
+/// has not yet written a first heartbeat is not considered expired (the beat may
+/// simply not have fired yet).
 bool _isHeartbeatExpired(
   SessionSnapshot snapshot, [
   Duration timeout = _defaultHeartbeatTimeout,
 ]) {
-  if (snapshot.status?.value != 'pending' || snapshot.heartbeatAt == null) {
+  final status = snapshot.status?.value;
+  if ((status != 'pending' && status != 'aborting') ||
+      snapshot.heartbeatAt == null) {
     return false;
   }
   final last = DateTime.tryParse(snapshot.heartbeatAt!);
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -147,11 +147,21 @@ class AgentInitError extends GenkitException {
 // ---------------------------------------------------------------------------
 
 /// Builds an abort-aware [SnapshotMutator]: it skips the write (returns `null`)
-/// when the current snapshot was concurrently aborted, otherwise writes
-/// [input]. This prevents a "completed"/"failed" write from clobbering an
-/// "aborted" status set by a concurrent abort.
+/// when the current snapshot has already *settled*, otherwise writes [input].
+/// This prevents a "completed"/"failed" write from clobbering a terminal status.
+///
+/// An `aborting` row is deliberately NOT treated as settled: it sits between the
+/// abort flip (which stops the work) and the finalize (which stamps the state
+/// on), so the finalize that settles it to `aborted` with state must be allowed
+/// through.
 SnapshotMutator _abortAwareMutator(SessionSnapshot input) {
-  return (current) => current?.status?.value == 'aborted' ? null : input;
+  return (current) {
+    final status = current?.status?.value;
+    if (status == 'completed' || status == 'failed' || status == 'aborted') {
+      return null;
+    }
+    return input;
+  };
 }
 
 /// Asserts that an operation requiring a persistent store is not being invoked
@@ -166,28 +176,39 @@ void _requireStore(SessionStore? store, String operation, String agentName) {
   }
 }
 
-/// Sets a snapshot's status to `aborted` (unless it already reached a terminal
-/// state) and returns its previous status, or `null` when the snapshot does
-/// not exist.
+/// Flips a `pending` snapshot to `aborting` via an ordinary `saveSnapshot`, and
+/// returns the *resulting* status: `aborting` when the row was pending (or was
+/// already aborting, since a second abort is an idempotent verbatim rewrite),
+/// the existing terminal status when the row had already settled, or `null` when
+/// the snapshot does not exist.
+///
+/// This is the first of the abort protocol's two writes: it stops the work (the
+/// detached worker observes the flip and cancels its turn) but does not stamp
+/// state - the worker's own finalize write later settles the row to `aborted`
+/// *with* the last-good state (see [SessionRunner.maybeSnapshot]). The flip
+/// deliberately leaves `heartbeatAt` where the worker left it: the beat is the
+/// worker's liveness signal, so a dead worker's row reads as `expired` at once
+/// rather than a heartbeat timeout later.
 Future<String?> _abortSnapshotInStore(
   SessionStore store,
   String snapshotId,
   Map<String, dynamic>? context,
 ) async {
-  String? previousStatus;
+  String? resultStatus;
   await store.saveSnapshot(snapshotId, (current) {
-    if (current == null) return null;
-    previousStatus = current.status?.value;
-    if (previousStatus == 'completed' ||
-        previousStatus == 'failed' ||
-        previousStatus == 'aborted') {
-      return null; // Already terminal - don't override.
+    if (current == null) return null; // not found
+    final status = current.status?.value;
+    if (status != 'pending') {
+      // Settled, or already aborting: re-persist verbatim so the returned
+      // status carries the existing value (idempotent).
+      resultStatus = status;
+      return current;
     }
-
-    current.status = SnapshotStatus.aborted;
+    current.status = SnapshotStatus.aborting;
+    resultStatus = 'aborting';
     return current;
   }, context: context);
-  return previousStatus;
+  return resultStatus;
 }
 
 // ---------------------------------------------------------------------------
@@ -202,6 +223,16 @@ const Duration _defaultHeartbeatInterval = Duration(seconds: 30);
 /// has not advanced is reported as `expired` on read. Comfortably larger than
 /// [_defaultHeartbeatInterval] so a single missed beat does not trip expiry.
 const Duration _defaultHeartbeatTimeout = Duration(seconds: 60);
+
+/// Bounds how long an `aborting` invocation keeps heartbeating while its worker
+/// drains toward the finalize write. The beats are what keep a live,
+/// winding-down worker's row reading as `aborting` instead of `expired` the
+/// moment the abort flip lands: the flip stops the work, not the worker, and a
+/// drain through a slow tool call can outlast [_defaultHeartbeatTimeout]. A
+/// worker that has not finalized within the budget is presumed wedged; its
+/// beats stop, the row goes stale and reads as `expired`, and the task stays
+/// recoverable instead of winding down forever.
+const Duration _windDownHeartbeatBudget = Duration(minutes: 5);
 
 /// Returns `true` when [snapshot] is a detached, in-flight snapshot whose
 /// heartbeat is older than [timeout] - i.e. its background worker is presumed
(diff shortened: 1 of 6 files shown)


packages/genkit/lib/src/types.dart
```diff
@@ -540,6 +550,7 @@ abstract class $Artifact {
 abstract class $GetSnapshotDataInput {
   String? get snapshotId;
   String? get sessionId;
+  bool? get metadataOnly;
```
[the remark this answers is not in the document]
pavelgj: later PR

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -475,10 +594,11 @@ class SessionRunner<State> {
...
-        // `aborted` rather than `failed`: the abort path already persisted the
-        // `aborted` status (the abort-aware mutator would skip a `failed` write
-        // anyway), so we record the finish reason and skip the failed snapshot
-        // write entirely instead of reporting a spurious error.
+        // `aborted` rather than `failed`: the settling `aborted` snapshot was
+        // already written on the resolve path above (or, for a wedged run, is
+        // owed by the finalize), and the abort-aware mutator would skip a late
+        // `failed` write anyway, so we record the finish reason and skip the
+        // failed snapshot write entirely instead of reporting a spurious error.
         if (cancel?.isCancelled ?? false) {
```
apascal07, line 610: A cancelled turn that *throws* lands here and breaks without writing a snapshot, so nothing settles the `aborting` row. The trailing `runner.maybeSnapshot()` in `flow()` returns early at the detached guard (`snapshotId` is null, `_lastSnapshot.snapshotId` is the pending id), and `ensureRecoverySnapshot()` only runs on the `failed` output path.

The throw path is reachable: `core/action.dart:268` calls `cancel?.throwIfCancelled()` on every action invoked inside the turn, so aborting a detached run mid tool-loop throws rather than resolving. The row then stays `aborting`, `stopHeartbeat()` runs in the `finally`, and 60s later every read shapes it to `expired`. `wait()` returns `expired` instead of `aborted`, and `_assertResumable` sends the caller to the parent, dropping the last-good state.

Go has no equivalent gap: `finalizePendingSnapshot` runs after fn returns regardless of how it ended and unconditionally settles an `aborting` row to `aborted` with the state. The comment above ("is owed by the finalize") names a step Dart does not have. Before this PR the row was already terminal here, since `_abortSnapshotInStore` wrote `aborted` directly.

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -147,11 +147,21 @@ class AgentInitError extends GenkitException {
...
+///
+/// An `aborting` row is deliberately NOT treated as settled: it sits between the
+/// abort flip (which stops the work) and the finalize (which stamps the state
+/// on), so the finalize that settles it to `aborted` with state must be allowed
+/// through.
 SnapshotMutator _abortAwareMutator(SessionSnapshot input) {
-  return (current) => current?.status?.value == 'aborted' ? null : input;
+  return (current) {
+    final status = current?.status?.value;
+    if (status == 'completed' || status == 'failed' || status == 'aborted') {
```
apascal07: `aborting` falls through to `return input`, so whatever status the runner decided lands, not `aborted`.

The notification that cancels the token is asynchronous (a microtask for `InMemorySessionStore`, up to the 2s poll for `FirestoreSessionStore`, a filesystem event for `FileSessionStore`). If the abort flip lands after the turn's `if (cancel?.isCancelled ?? false)` check at line 527 but before the turn-end write, `maybeSnapshot(status: 'completed')` sees `aborting`, writes through, and the row settles `completed`. `abort()` already returned `aborting`, so the caller polls and watches the abort disappear. `tests/specs/agent.yaml` ("abort pending agent") requires that row to settle `aborted`. The `failed` write at line 611 can clobber it the same way.

Go closes this in the mutator rather than at the call site: `finalizePendingSnapshot` has `case existing.Status == SnapshotStatusAborting`, which stamps `aborted`, `finishReason: aborted`, and the state whatever the fn's outcome was.

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -788,7 +903,7 @@ _resolveSession<State>(
       context: context,
     );
     final visited = <String>{};
-    while (snapshot != null && snapshot.status?.value != 'completed') {
+    while (snapshot != null && !_isResumableLeaf(snapshot)) {
```
apascal07, line 937: The walk treats `pending` and `aborting` as "keep going", so a live in-flight leaf is skipped and the resume forks off its parent.

Abort a detached run and immediately `chat(sessionId:)`: the leaf is `aborting` with no state yet, so the walk resumes from the parent. The worker's finalize then stamps the aborted state onto the leaf, and the parent has two children. With `rejectBranchingSessions` set, `FileSessionStore._latestSnapshotForSession` fails the session outright; without it the pointer picks a leaf arbitrarily.

Go rejects instead of walking: `resumeSessionFrom` returns FAILED_PRECONDITION for both statuses, on the grounds that sending the caller to an earlier snapshot forks the run away from work about to be committed.

Separately, the comment above this loop still says only `completed` snapshots are resumable and that the walk targets "the last-good (`completed`) snapshot".

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -1112,330 +1230,351 @@ Agent<State> defineCustomAgent<State>(
...
+                  // The runtime's own abort flips the row to `aborting`; a
+                  // foreign writer that lands `aborted` directly still means
+                  // stop.
+                  final status = snap.status?.value;
+                  if (status != 'aborting' && status != 'aborted') return;
+                  // The flip stops the work, not the worker: the run is
+                  // winding down toward its finalize write now, and the
+                  // heartbeat keeps running so readers see a live `aborting`
+                  // row instead of presuming the worker dead the moment the
+                  // flip lands. The flow's `finally` stops the beats right
```
apascal07: The `finally` in `flow()` runs after `await runner.maybeSnapshot()`, so the beats stop *after* the finalize, not before it. Go orders it the other way (`stopHeartbeat()`, then `finalizePendingSnapshot`).

Harmless today, because the heartbeat mutator no-ops on a terminal row. Worth correcting so that guard is not later read as redundant.

packages/genkit/test/ai/agents/agents_spec_test.dart
```diff
@@ -842,7 +842,13 @@ Future<void> _executeAbort(
     throw _SpecError('abort invocation requires snapshotId');
   }
 
-  final previousStatus = await agent.abort(snapshotId);
+  // Capture the status before aborting so we can assert `expectPreviousStatus`.
+  // `abort()` returns the status *after* the attempt (a pending row reads back
+  // as `aborting`), while the spec asserts the *previous* status, so the harness
+  // reads it first. Mirrors the Go and JS conformance harnesses.
+  final before = await agent.getSnapshotData(snapshotId: snapshotId);
```
apascal07: `getSnapshotData` applies the `expired` heartbeat shaping, so this reads a shaped status where the Go harness reads the raw stored one (`store.GetSnapshot`).

A detached row beats every 30s against a 60s timeout. If the beat is stale when this step runs, the read returns `expired` and every `expectPreviousStatus: pending` case fails, even though the stored status is `pending` and the abort itself succeeded.

packages/genkit/test/ai/agents/agent_test.dart
```diff
@@ -598,6 +600,60 @@ void main() {
...
     );
 
+    test(
+      'getSnapshotData reports a stale aborting snapshot as expired',
+      () async {
+        // A cross-runtime read: a Go server settles its abort in two writes
+        // (flip -> `aborting`, later finalize -> `aborted`) and keeps
+        // heartbeating while it winds down. A stale `aborting` beat means the
+        // draining worker died, so - like a stale `pending` - it must read as
+        // `expired`. Dart itself settles aborts in one write and never emits
```
apascal07: No longer true. `_abortSnapshotInStore` now writes `SnapshotStatus.aborting`, so Dart both emits and reads the mid-flight window. The test is right, only the rationale is stale.

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -166,28 +176,39 @@ void _requireStore(SessionStore? store, String operation, String agentName) {
...
+    final status = current.status?.value;
+    if (status != 'pending') {
+      // Settled, or already aborting: re-persist verbatim so the returned
+      // status carries the existing value (idempotent).
+      resultStatus = status;
+      return current;
     }
-
-    current.status = SnapshotStatus.aborted;
+    current.status = SnapshotStatus.aborting;
```
apascal07, line 207: The flip does not advance `updatedAt`. Dart stores are caller-managed (only `maybeSnapshot` and `ensureRecoverySnapshot` stamp it), so an aborting row keeps the timestamp from before the abort, and a client ordering or displaying by `updatedAt` shows no activity. Go sets `updated.UpdatedAt = now` on this same transition. The already-settled branch above correctly leaves it alone, matching Go's verbatim re-persist.

apascal07 (approved):

packages/genkit/lib/src/ai/agents/agent.dart
pavelgj: Good catch, fixed. The cancel branch in the catch now writes the settling `aborted` snapshot (with the last-good state) before breaking, instead of just recording the finish reason. That is the finalize Dart was missing versus Go's `finalizePendingSnapshot`: previously a turn aborted via a thrown `throwIfCancelled()` left the row stuck `aborting`, which later shaped to `expired` and dropped state on resume. Rewrote the stale comment too, and added a regression test ("a detached run aborted via a thrown cancel settles as aborted") since the existing coverage only exercised the resolve-on-cancel path.

packages/genkit/test/ai/agents/agent_test.dart
pavelgj: You're right, the test is correct but the rationale was stale now that the abort flip emits `aborting`. Updated the comment to describe the two-write protocol and that this test exercises the read side (a stale `aborting` beat shaping to `expired`, whatever runtime wrote the row).

packages/genkit/lib/src/ai/agents/agent.dart
pavelgj: Fixed. The flip branch now stamps `updatedAt = now`, matching Go's transition. Left the already-settled / verbatim re-persist branch untouched, as you noted Go does.

packages/genkit/lib/src/ai/agents/agent.dart
pavelgj: Fixed with strict Go parity: the `sessionId` resume path now rejects a `pending`/`aborting` leaf with FAILED_PRECONDITION instead of walking to the parent, matching `resumeSessionFrom`. That removes the fork-off-parent hazard you described. Also fixed the stale comment above the loop that still claimed only `completed` snapshots are resumable.

packages/genkit/test/ai/agents/agents_spec_test.dart
pavelgj: Fixed. `_executeAbort` now reads the raw stored status via the store (`store.getSnapshot`) rather than `getSnapshotData`, so the `expired` heartbeat shaping no longer leaks into the `expectPreviousStatus` assertion. Mirrors the Go harness reading `store.GetSnapshot`. Threaded the store instances through `_setupHarness` (it now returns `(agents, stores)`).

PR #413, opened 2026-09-03
feat(agents)!: graceful `failed` responses (model + tool errors) + rerunnable failed snapshots

pavelgj:
Stacked on #411 (`aborting` status + rerunnable aborted snapshots). Adds the `failed` finish-reason counterpart, matching the Go reference (`go/ai/exp` `failurePartial` + failed-turn commit).

## ⚠️ Breaking changes
1. **A model error during `generate` no longer throws.** It resolves to a `GenerateResponse` with `finishReason: failed`, carrying the structured `error` and the last-good history (the failing turn's own partial output is dropped). Mirrors how cancellation already resolves to `finishReason: aborted`, and how Go returns a failed partial.
2. **A throwing tool now fails the generation.** Previously a tool `fn` that threw was swallowed into an `Error: ...` tool response and fed back to the model. Now a throwing tool - along with a tool-not-found, or a tool's own unrelated `CancelledException` - fails the generation with `finishReason: failed` and the last-good history, exactly like a model failure. A caller-token cancellation resolves to `aborted`.

Callers that previously wrapped `generate()` in try/catch to observe a model/tool error must now branch on `response.finishReason == FinishReason.failed` and read `response.error` (and, in-process, `response.cause`).

### a. Graceful `failed` generate responses
- `_failedResponse` (new): builds a `finishReason: failed` response with `error` (a `RuntimeError` mapped from the cause; a `GenkitException` keeps its status, anything else is `INTERNAL`) and the turn's last-good `messages` as the resume point. Mirrors `_abortedResponse`.
- The model-call, tool-execution, and restart-tool `catch` sites now return `_failedResponse(...)` for a genuine (non-cancel) error instead of rethrowing / swallowing. The cancel path is unchanged (still `aborted`).
- `GenerateResponseHelper` forwards the new `error` field through from the `ModelResponse`, and gains a Dart-only `cause` field exposing the original thrown object for in-process inspection (`res.cause is SocketException`). `cause` does not cross the reflection/HTTP boundary; `error` is the serializable view.

### b. Throwing tools fail the loop
- `_executeTools` now only catches `ToolInterruptException` (a turn outcome, not a failure). Every other throw propagates to `_runGenerateLoop`, which converts it into a `failed` response (or `aborted` when this turn's token was cancelled).

### c. Agent persists a rerunnable `failed` snapshot
- `TurnResult` gains an `error` field; the prompt-backed agent loop maps a `finishReason: failed` generate response into a `TurnResult` carrying that error and keeps the last-good history (same handling the aborted path already had).
- `SessionRunner.run` gains a `failed` branch: it commits a `failed` snapshot with the error and the last-good state, so a client can branch on the status and rerun the same snapshot id (the resume guard from #411 already accepts `failed` rows that carry state).
- `ensureRecoverySnapshot` reports a first-turn failure's own persisted `failed` snapshot as the resume point.
- `AgentResponse.error` (new getter) exposes the structured error to callers.

### Supporting: `RuntimeError` + `FinishReason.failed` + `GenerateResponse.error`
Regenerated `types.dart` / `types.g.dart` to pull in the error wire shape the updated schema references directly from `ModelResponse.error` / `GenerateResponse.error` (previously only aliased via `AgentErrorInfo`), plus `FinishReason.failed`. Allowlisted `RuntimeError` in the schema generator so it compiles. (This also pulls in `metadataOnly` on `GetSnapshotDataInput` from the canonical schema; wiring it up is out of scope here.)

## Handling a failed response

```dart
final res = await ai.generate(model: m, prompt: 'hi', toolNames: ['myTool']);
switch (res.finishReason) {
  case FinishReason.failed:
    print(res.error?.status);  // e.g. 'UNAVAILABLE' (model) or the tool's status
    print(res.cause);          // the raw thrown exception (in-process only)
    // res.messages is the last-good history; retry/resume from it.
  case FinishReason.aborted:
    // cancelled; res.messages is the resume point.
  default:
    print(res.text);
}
```

Agent side, a `failed` turn persists a rerunnable snapshot:

```dart
try {
  await chat.send(text: 'hi'); // throws AgentError on a failed turn
} on AgentError catch (e) {
  final snap = await agent.getSnapshotData(snapshotId: e.snapshotId);
  // snap.status == 'failed', snap.state carries the resumable history.
  final rerun = agent.chat(snapshotId: e.snapshotId);
  await rerun.send(text: 'hi'); // re-drives from the last-good state
}
```

## Tests
- generate: model error -> `failed` with error + resumable history + `cause`; a throwing tool -> `failed` (last-good history preserved, model not re-called); tool-not-found -> `failed` (`NOT_FOUND`); non-`GenkitException` -> `INTERNAL`; a failure after a successful tool-call turn preserves that turn as last-good.
- cancel: updated the one test that encoded the old swallow-and-recover behavior to assert a tool's unrelated cancellation now fails the generation.
- agent: a prompt-backed turn whose model fails persists a `failed` snapshot with rerunnable history, and reruns to success once the model recovers.
- Full `packages/genkit` suite green (643 tests), `dart analyze` clean.

## Note for reviewers
Both breaks are intentional Go-parity / correct-default choices; Dart is in preview. Worth a CHANGELOG callout on release.

### DIFF
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -47,9 +47,16 @@ import 'session.dart';
 /// ended (e.g. `interrupted`, `length`). When omitted, no per-turn reason is
 /// reported.
 class TurnResult {
-  TurnResult({this.finishReason});
+  TurnResult({this.finishReason, this.finishMessage});
 
   final AgentFinishReason? finishReason;
+
+  /// An optional human-readable message describing why the turn ended.
+  ///
+  /// Carries, for example, the "Reached max turns of N" text from a generation
+  /// that overran its turn limit, so the [SessionRunner] can surface it rather
+  /// than dropping it.
+  final String? finishMessage;
 }
 
 /// Per-turn context handed to the handler passed to [SessionRunner.run].
@@ -404,14 +411,50 @@ class SessionRunner<State> {
           if (cancel?.isCancelled ?? false) {
             lastTurnFinishReason = AgentFinishReason.aborted;
             lastTurnError = null;
+            // Persist the turn as `aborted`. The detached route already wrote
+            // `aborted` via `_abortSnapshotInStore`, in which case the
+            // abort-aware mutator makes this a no-op; but the attached route
+            // (`AgentTurn.abort()` -> token cancel) has written nothing, so
+            // without this the trailing `invocationEnd` snapshot would persist a
+            // half-finished turn as `completed` and later be picked as a resume
+            // point.
+            final snapshotId = await maybeSnapshot(
+              status: 'aborted',
+              snapshotId: turnSnapshotId,
+              finishReason: AgentFinishReason.aborted,
+            );
             _notifyEndTurn(
-              _lastSnapshot?.snapshotId,
+              snapshotId ?? _lastSnapshot?.snapshotId,
               AgentFinishReason.aborted,
             );
             return true;
           }
 
           final finishReason = turnResult?.finishReason;
+
+          // A turn that resolved `aborted` *without* the token being cancelled
+          // is not a cooperative cancel: it is an overrun (e.g. the generate
+          // loop hit `maxTurns`). Route it to the failure path so the reason
+          // (e.g. "Reached max turns of N") surfaces as an error instead of
+          // being silently dropped as a success with a null message.
+          if (finishReason == AgentFinishReason.aborted) {
+            lastTurnFinishReason = AgentFinishReason.failed;
+            lastTurnError = toErrorDetails(
+              GenkitException(
+                turnResult?.finishMessage ?? 'Turn aborted.',
+                status: StatusCodes.ABORTED,
+              ),
+            );
+            final snapshotId = await maybeSnapshot(
+              status: 'failed',
+              error: lastTurnError,
+              snapshotId: turnSnapshotId,
+              finishReason: AgentFinishReason.failed,
+            );
+            _notifyEndTurn(snapshotId, AgentFinishReason.failed);
+            return true;
+          }
+
           lastTurnFinishReason = finishReason;
           lastTurnError = null;
 
@@ -898,7 +941,7 @@ final class _InProcessTransport extends AgentTransport {
   TurnStream runTurn(
     AgentInput input,
     AgentInit init, {
-    required CancellationToken cancel,
+    CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) {
     final bidi = _startBidi(input, init, context: context, cancel: cancel);
@@ -909,7 +952,7 @@ final class _InProcessTransport extends AgentTransport {
   Future<AgentOutput>? run(
     AgentInput input,
     AgentInit init, {
-    required CancellationToken cancel,
+    CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) {
     return _startBidi(input, init, context: context, cancel: cancel).onResult;
@@ -1137,7 +1180,9 @@ Agent<State> defineCustomAgent<State>(
           // Capture the disposer so a reused, long-lived `ctx.cancel` token
           // doesn't accumulate one stranded listener (pinning this turn's
           // controller and closure) per turn. Disposed when the flow settles.
-          final unlinkCancel = ctx.cancel.onCancel(cancelController.cancel);
+          // `link` forwards the cancellation reason so a caller-supplied
+          // `controller.cancel('...')` survives the hop into this turn's token.
+          final unlinkCancel = ctx.cancel?.link(cancelController);
           final cancelToken = cancelController.token;
           void Function()? unsubscribe;
           // Background heartbeat timer for the detached snapshot. Started in
@@ -1302,7 +1347,7 @@ Agent<State> defineCustomAgent<State>(
               // The turn has settled (the snapshot reached a terminal status),
               // so stop refreshing its heartbeat.
               stopHeartbeat();
-              unlinkCancel();
+              unlinkCancel?.call();
               unsubscribe?.call();
               offArtifactAdded();
               offArtifactUpdated();
@@ -1597,10 +1642,16 @@ Agent<State> definePromptAgent<State>(
         context: options.context,
         inputStream: null,
         init: null,
-        cancel: options.cancel ?? CancellationToken.none,
+        cancel: options.cancel,
       ));
 
-      // Keep everything that is NOT a prompt-template message.
+      final aborted = res.finishReason == FinishReason.aborted;
+
+      // Keep everything that is NOT a prompt-template message. On an aborted
+      // turn `res.message` is null; the abort path already carries the last-good
+      // history in `res.modelRequest`, so leaving `keep` without a trailing
+      // model message preserves the user turn as the resume point rather than
+      // treating the user's own message as the model reply.
       final reqMessages = res.modelRequest?.messages;
       if (reqMessages != null) {
         final keep = reqMessages
@@ -1628,13 +1679,18 @@ Agent<State> definePromptAgent<State>(
       final reason = res.finishReason;
       return TurnResult(
         finishReason: reason != null ? AgentFinishReason(reason.value) : null,
+        finishMessage: aborted ? res.finishMessage : null,
       );
     });
 
     final msgs = sess.getMessages();
+    // Only surface the last message as the agent's reply when it is actually a
+    // model turn. An aborted turn leaves the user message last (see above), so
+    // returning it would echo the user's prompt back as the agent's reply.
+    final lastMessage = msgs.isNotEmpty ? msgs.last : null;
     return AgentResult(
       artifacts: sess.getArtifacts(),
-      message: msgs.isNotEmpty ? msgs.last : null,
+      message: lastMessage?.role == Role.model ? lastMessage : null,
       finishReason: sess.lastTurnFinishReason,
     );
   }
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -203,15 +203,21 @@ const Duration _defaultHeartbeatInterval = Duration(seconds: 30);
 /// [_defaultHeartbeatInterval] so a single missed beat does not trip expiry.
 const Duration _defaultHeartbeatTimeout = Duration(seconds: 60);
 
-/// Returns `true` when [snapshot] is a `pending` (detached, in-flight) snapshot
-/// whose heartbeat is older than [timeout] - i.e. its background worker is
-/// presumed dead. A pending snapshot that has not yet written a first heartbeat
-/// is not considered expired (the beat may simply not have fired yet).
+/// Returns `true` when [snapshot] is a detached, in-flight snapshot whose
+/// heartbeat is older than [timeout] - i.e. its background worker is presumed
+/// dead. Both `pending` and `aborting` are in-flight, still-heartbeating states
+/// (a worker keeps beating while it winds down an abort), so a stale beat in
+/// either reads as `expired`; this mirrors the shared spec, where a `pending`
+/// *or* `aborting` row with a stale beat surfaces as `expired`. A snapshot that
+/// has not yet written a first heartbeat is not considered expired (the beat may
+/// simply not have fired yet).
 bool _isHeartbeatExpired(
   SessionSnapshot snapshot, [
   Duration timeout = _defaultHeartbeatTimeout,
 ]) {
-  if (snapshot.status?.value != 'pending' || snapshot.heartbeatAt == null) {
+  final status = snapshot.status?.value;
+  if ((status != 'pending' && status != 'aborting') ||
+      snapshot.heartbeatAt == null) {
     return false;
   }
   final last = DateTime.tryParse(snapshot.heartbeatAt!);
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -1134,7 +1134,10 @@ Agent<State> defineCustomAgent<State>(
           // transport token to it, so an attached `runTurn(cancel:)` also
           // cooperatively stops this turn's `generate`.
           final cancelController = CancellationController();
-          ctx.cancel.onCancel(cancelController.cancel);
+          // Capture the disposer so a reused, long-lived `ctx.cancel` token
+          // doesn't accumulate one stranded listener (pinning this turn's
+          // controller and closure) per turn. Disposed when the flow settles.
+          final unlinkCancel = ctx.cancel.onCancel(cancelController.cancel);
           final cancelToken = cancelController.token;
           void Function()? unsubscribe;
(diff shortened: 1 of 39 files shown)


packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -1771,9 +1809,27 @@ Agent<State> definePromptAgent<State>(
...
       final reason = res.finishReason;
+      // A failed generate response carries its structured `error`; surface it on
+      // the TurnResult so the runtime persists a `failed` snapshot with the
+      // last-good history (rerunnable), rather than treating it as a completed
+      // turn.
+      final error = failed
+          ? (res.error != null
+                ? AgentErrorDetails(
+                    status: res.error!.status ?? StatusCodes.INTERNAL.name,
+                    message: res.error!.message,
```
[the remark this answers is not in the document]
pavelgj: Not needed here: this is the `res.error != null` branch, and `AgentErrorInfo.message` is a non-nullable `String` (generated getter: `String get message => _json['message'] as String`), so `res.error!.message` can't be null. Adding `?? 'Generation failed.'` would be a dead null-aware expression that `dart analyze` flags. The `status` field above genuinely is nullable, which is why the `??` fallback is correct there.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -460,16 +515,23 @@ Future<GenerateResponseHelper> _runGenerateLoop(
...
+    // `failurePartial`.
     final aborted = _abortResponseIfCancelled(
       e,
       ctx.cancel,
       history: options.messages,
       config: options.config,
     );
     if (aborted != null) return aborted;
-    rethrow;
+    return _failedResponse(
```
apascal07, line 591: Go's parity shape is `(partial, err)`: the partial is a convenience, the error stays mandatory. Returning only the response drops that half, so a failure is observable only if the caller remembers to branch on `finishReason`.

The two test edits in this PR are the visible cost. An unregistered a2ui catalog id (a `StateError`) and an invalid OpenAI tool schema (`INVALID_ARGUMENT`) are developer mistakes that used to surface immediately. They now resolve with `res.text == ''` and no signal at all.

It is also not applied consistently. `_runGenerateLoop` still throws for `model == null` (L380) and model-not-found (L403), which are the same category of mistake. Which one throws now depends on how deep in the call stack it is raised, not on what kind of failure it is. Either throw for caller-fault statuses (INVALID_ARGUMENT, NOT_FOUND, UNIMPLEMENTED) and resolve `failed` for the rest, or resolve those too, but the split as it stands is arbitrary.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -514,17 +576,24 @@ Future<GenerateResponseHelper> _runGenerateLoop(
...
+    // history, so the caller can inspect `response.error` and resume.
     final aborted = _abortResponseIfCancelled(
       e,
       ctx.cancel,
       history: options.messages,
       config: options.config,
     );
     if (aborted != null) return aborted;
-    rethrow;
+    return _failedResponse(
```
apascal07, line 679: Go reclassifies every tool failure rather than passing the tool's own status through. `ErrToolFailed = status.ErrInternal.Subtype("tool failed")` in `go/ai/errors.go`, documented as "the status is INTERNAL because a tool's failure is not a failure of the caller's request", and `toolFailureError` wraps the message as `tool %q failed: %w`.

`_toRuntimeError` keeps the tool's status verbatim. A tool throwing `GenkitException('no auth', status: UNAUTHENTICATED)` makes the failed response report `error.status == 'UNAUTHENTICATED'` for the whole generation, so a retry client re-authenticates the caller for a fault that happened inside a tool. The new test asserting `FAILED_PRECONDITION` pins that behavior.

The tool name is lost too. With several tool requests in one turn, `error.message` does not say which one failed.

I can see an argument for either way here... but returning UNAUTHENTICATED to a user application is probably pointless.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -254,6 +254,61 @@ GenerateResponseHelper _abortedResponse({
...
+  Map<String, dynamic>? config,
+  GenerationUsage? usage,
+}) {
+  final error = _toRuntimeError(cause);
+  final request = ModelRequest(messages: history, config: config);
+  return GenerateResponseHelper(
+    ModelResponse(
+      finishReason: FinishReason.failed,
+      finishMessage: error.message,
+      error: error,
```
apascal07, line 300: `failurePartial` sets `p.Error = responseError(cause)` before it branches on failed vs aborted, so a Go aborted partial carries an error as well. `ErrMaxTurnsExceeded` is `status.ErrAborted.Subtype(...)`, giving `Error.Status == ABORTED`.

`_abortedResponse` (L234) still leaves `error` null. Now that this PR makes `error` the documented way to read a failure, `generate(maxTurns: 1)` against a tool-looping model resolves with `finishReason: aborted` and `error == null`, and a caller writing `res.error!.status` after seeing an abnormal finish reason hits a null check error.

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -634,9 +667,12 @@ class SessionRunner<State> {
       return _lastSnapshot?.snapshotId;
     }
 
-    // First-turn failure: the last-good state is the seed the client holds.
+    // First-turn failure: the last-good state is the seed the client holds,
+    // so there is no prior turn snapshot to recover to. If this turn itself
+    // persisted a terminal (e.g. `failed`) snapshot, that row is the resume
+    // point; report it. Otherwise there is nothing to resume.
     if (turnIndex == 0) {
```
apascal07: This reports the turn's own `failed` row on a first-turn failure, but a failure at turn >= 1 never reaches here.

`lastGoodStateVersion` is seeded in the constructor and `maybeSnapshot` sets `_lastSnapshotVersion = session.getVersion()` after the failed write, so the version guard above never matches on a failed turn. On turn 0 that lands on this new `return`. On turn 2 it falls through and writes a fresh `completed` snapshot from `lastGoodState`, which is the state as of turn 1, before the failed turn's user message.

So `AgentError.snapshotId` is a `failed` row on the first turn and a `completed` row with a shorter history after that. The rerun pattern in the PR description (`snap.status == 'failed'`, then `agent.chat(snapshotId: ...)`) only holds for the first turn.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -681,18 +750,24 @@ Future<GenerateResponseHelper> _runGenerateAction(
...
+        // `CancelledException` to the caller.
         final aborted = _abortResponseIfCancelled(
           e,
           c.cancel,
           history: opts.messages,
           config: opts.config,
         );
         if (aborted != null) return aborted;
-        rethrow;
+        return _failedResponse(
```
apascal07, line 857: This is the third `_failedResponse` site, but the tail of `_runGenerateAction` (L850, `return composedGenerate(...)`) has no catch. A middleware whose `generate` hook throws before delegating to `next` still escapes `generate()` as a throw, while a middleware whose `model` hook throws (the a2ui catalog case) is caught at L516 and resolves `failed`.

Go covers both, at the end of `generateWithRequest`: `if err != nil && resp == nil { resp = lastPartial ... else failurePartial(ctx, nil, lastReq, err) }`. One extension point with two contracts depending on which hook broke.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -254,6 +254,61 @@ GenerateResponseHelper _abortedResponse({
...
 
+/// Maps a thrown value to the structured [RuntimeError] carried on a failed
+/// response's `error` field. Preserves a [GenkitException]'s status; anything
+/// else is reported as `INTERNAL`.
+RuntimeError _toRuntimeError(Object cause) {
+  if (cause is GenkitException) {
+    return RuntimeError(
+      status: cause.status.name,
+      message: cause.message,
+      details: cause.details != null ? {'details': cause.details} : null,
```
apascal07: `GenkitException.details` is a `String?`, so this produces `error.details == {'details': 'body'}`, and agent.dart:1865 copies it straight onto `AgentErrorDetails.details`. The persisted snapshot row ends up holding `error.details.details`.

The sibling normalizer `toErrorDetails` (agent.dart:116) writes the same field unwrapped, with a `?? underlyingException ?? message` fallback. Both shapes reach `AgentErrorInfo.details` depending on whether the error resolved through the generate loop or was thrown out of the turn, and a client reading `snapshot.error.details` cannot handle both. `underlyingException` is dropped on this path too.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -460,16 +515,23 @@ Future<GenerateResponseHelper> _runGenerateLoop(
...
     if (aborted != null) return aborted;
-    rethrow;
+    return _failedResponse(
+      history: options.messages,
+      config: options.config,
+      cause: e,
+    );
   }
 
   final parser = format
```
apascal07, line 598: Two gaps against Go on the parse path, both more visible now that everything else resolves.

Go skips output parsing when the finish reason is abnormal (`if resp.FinishReason.isAbnormal()`), so a model that itself returns `blocked` or `failed` passes through untouched. Here the parser is built unconditionally and applied at L558, so a blocked response with an empty message is still run through the JSON parser.

Go also returns a parse failure as a response with `Error` set and the original message and finish reason intact, never as a partial. `_parseOutput` throws instead, which leaves a schema mismatch as the one remaining throw out of `generate()`.

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -514,17 +576,24 @@ Future<GenerateResponseHelper> _runGenerateLoop(
...
     );
     if (aborted != null) return aborted;
-    rethrow;
+    return _failedResponse(
+      history: options.messages,
+      config: options.config,
+      cause: e,
+      // The model already answered this turn; carry its usage onto the failed
+      // response so the tokens the turn spent are still reported.
+      usage: response.usage,
```
apascal07: `failurePartial` copies the whole base response (`p = *base`) and clears only `Message`, so `Custom`, `LatencyMs`, `Raw` and `Operation` ride along with `Usage`.

Carrying `usage` alone drops the rest. A provider that reports cache-read or billing detail in `custom` rather than in `usage` loses it whenever a tool in that turn fails, which is the same accounting this argument exists to preserve.

apascal07 (approved):

packages/genkit/lib/src/ai/generate.dart
pavelgj: Fixed. `_toRuntimeError` no longer nests `{'details': ...}`; it carries just the classified status and message, matching Go's `responseError` and the sibling `toErrorDetails` shape. The agent-side mapping now sees a flat `error`, so `snapshot.error.details` is consistent across both paths.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Agreed, matched Go. New `_toolFailureError` reclassifies a tool failure to INTERNAL with `tool "<name>" failed: <msg>`, mirroring `toolFailureError`/`ErrToolFailed`. The original exception is wrapped as `underlyingException` so it's still reachable via `res.cause`, and a context/token cancel still takes the aborted path rather than being blamed on the tool. Updated the test that pinned `FAILED_PRECONDITION` to assert INTERNAL + the tool name, with the original status reachable through `cause`.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Good catch, fixed. Folded `_abortedResponse` and `_failedResponse` into a single `_abnormalResponse` that always sets `error`, so both finishes carry a payload the same way Go's `failurePartial` sets `p.Error` before branching. `maxTurns`/cancel now resolve with an ABORTED-classed `error`; a genuine exception that raced a cancel keeps its own status. Added a `maxTurns` test asserting `res.error!.status == 'ABORTED'`.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Both handled. Output parsing is now skipped on an abnormal finish via a local `FinishReason.isAbnormal` (blocked/aborted/failed/interrupted/other), matching Go, so a blocked/empty response passes through untouched. A schema mismatch now resolves with the original message and finish reason intact under an INTERNAL error instead of throwing, mirroring the `ErrInvalidOutput` path. Added tests for both (blocked passthrough, and schema-mismatch resolving rather than throwing).

packages/genkit/lib/src/ai/generate.dart
pavelgj: Fixed. `_abnormalResponse` takes the whole base response and copies `usage`, `custom`, `raw`, `latencyMs`, and `operation`, clearing only the message, like Go's `p = *base`. The tool-failure site passes `base: response` so provider cache/billing detail in `custom`/`raw` survives. Added a test asserting `custom`/`raw` ride onto the failed response.

packages/genkit/lib/src/ai/generate.dart
pavelgj: Right, that was the one remaining escape. Wrapped the tail call in `_runGenerateAction` so a hook that throws before delegating to `next` resolves the same way the loop's own catch sites do: aborted when this turn's token was cancelled, otherwise failed with the entry history. Matches Go's `GenerateWithRequest` tail synthesizing a `failurePartial` for an error raised outside a turn. Added a test with a middleware whose `generate` hook throws.

packages/genkit/lib/src/ai/agents/agent.dart
pavelgj: You're right, and the fix is simpler than the old code. The failed turn already commits its own `failed` row (with the last-good history as its state) via `maybeSnapshot`, so `ensureRecoverySnapshot` now just reports `_lastSnapshot?.snapshotId` on any turn, dropping the fresh-`completed`-write fallback that pulled stale prior-turn state. Mirrors Go's `failedOutput` returning `lastSnapshotID`. Removed the now-dead `_lastGoodFinishReason`. Added a test: a failure on turn 2 reports its own `failed` snapshot (carrying the turn-2 user message) and reruns to success.

packages/genkit/lib/src/ai/generate.dart
pavelgj: I dug into the Go source on this one (`GenerateWithRequest`, L438-532, and the `Generate` doc at L1086-1087): the rule there is "errors before the request resolves throw with a nil response; everything after resolves to a failed/aborted partial." Model-nil (L452), model-not-found (L459), tool-not-found declared on the request (L478), and bad options all throw; the a2ui catalog `StateError` and the invalid tool schema are raised inside the model call, i.e. after the request resolved, so they resolve to `failed`. So the split is by "caller mistake before the run starts" vs "the run broke," not by call-stack depth, and Dart already lines up with it. The fair part of your point was that this was implicit, so I added a comment at the setup-phase throw sites documenting the rule. Happy to reconsider if you think the a2ui/schema cases should be reclassified to a caller-fault status and thrown, but that felt like a separate change from Go-parity.

PR #414, opened 2026-09-03
fix(genkit_openai): do not require network or an API key at startup

### LINKED ISSUE #360: fix(genkit_openai): init() requires network + API key at startup

adesinah:
## Summary

`OpenAIPlugin.init()` called `models.list()` whenever no custom `baseUrl` was
set, and resolved the API key there too. Both requirements landed at startup:

- **`init()` failure is fatal and uncached.** The registry initializes plugins
  *outside* the try/catch that guards per-plugin `list()`, and only records
  success — so the throw recurred on every subsequent call.
- **It took the Dev UI down.** `/api/__health` and `/api/actions` both go
  through `listActions()`, and one plugin's `init()` throw returns HTTP 500 for
  the whole request. With no key, the Dev UI could not connect at all.

`genkit_openai` was the only plugin in the repo doing I/O in `init()` —
`genkit_anthropic` and `genkit_google_genai` never override it.

After this change, `Genkit(plugins: [openAI()])` starts offline with no key, and
a missing or invalid key surfaces at `generate()` time instead.

### Also included: vision metadata fix

`_supportsVisionByHeuristics` matched vision only via a literal `gpt-4o`,
`vision`, `gpt-4-turbo`, or an `o` immediately after the version digits. Once
OpenAI moved that marker into the version number, every modern multimodal
family reported `media: false`:

```
gpt-4.1    media=false
gpt-5      media=false
gpt-5.5    media=false
```

Nothing in `packages/genkit` reads `supports`, so this misreports rather than
blocks: the models still take images, but every consumer of the metadata — the
Dev UI's model picker, `listActions` output, anything selecting a model by
capability — is told the current multimodal families are text-only. Replaced
with a version comparison (4.1 and newer, all of 5+); plain `gpt-4` stays
text-only. This is a separate bug, but the curated catalog can't carry correct
metadata without it.

## Testing

```bash
cd packages/genkit_openai
OPENAI_API_KEY=sk-... dart test test/integration_test.dart -n 'OPENAI_API_KEY|enriches'
```

Fixes #360

### DIFF
--- a/packages/genkit_openai/lib/src/known_models.dart
+++ b/packages/genkit_openai/lib/src/known_models.dart
@@ -1,0 +1,60 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Chat models the plugin curates so they stay listable without network access.
+///
+/// The plugin normally discovers models from `GET /models`, but that needs both
+/// connectivity and a valid key. This list is what gets listed when discovery
+/// is unavailable, and it is merged with the discovered set when it is.
+///
+/// It is a convenience, never a gate: the plugin resolves any model id, so a
+/// model absent from this list still works when named explicitly. That is
+/// deliberate - it keeps just-released models usable without a plugin release.
+/// Capability metadata is not duplicated here either; it comes from the
+/// heuristics behind `modelInfoFor`.
+///
+/// Curation rules, to keep this list from sprawling as OpenAI ships:
+///
+/// - Aliases only. Dated snapshots (`gpt-4o-2024-08-06`) and `-chat-latest`
+///   aliases are omitted; discovery surfaces them when online, and pinning a
+///   snapshot is an explicit act the plugin's resolver already supports.
+/// - Chat models only. Embeddings, image, audio and moderation models are not
+///   served by this plugin's chat path.
+/// - Recent generations plus what apps still pin in practice, not every
+///   generation ever shipped.
+const List<String> knownChatModels = <String>[
+  // GPT-5 family, newest first. `-pro` is the higher-reasoning tier;
+  // `-mini`/`-nano` trade capability for latency and cost.
+  'gpt-5.5',
+  'gpt-5.5-pro',
+  'gpt-5.4',
+  'gpt-5.4-mini',
+  'gpt-5.4-nano',
+  'gpt-5.4-pro',
+  'gpt-5',
+  'gpt-5-mini',
+  'gpt-5-nano',
+
+  // GPT-4.1 and 4o remain widely pinned by existing apps.
+  'gpt-4.1',
+  'gpt-4.1-mini',
+  'gpt-4.1-nano',
+  'gpt-4o',
+  'gpt-4o-mini',
+
+  // o-series reasoning models.
+  'o3',
+  'o3-mini',
+  'o4-mini',
+];
--- a/packages/genkit_openai/lib/src/openai_plugin.dart
+++ b/packages/genkit_openai/lib/src/openai_plugin.dart
@@ -15,6 +15,7 @@
 import 'package:genkit/plugin.dart';
 import 'package:http/http.dart' as http;
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:openai_dart/openai_dart.dart' as sdk;
 
 import '../genkit_openai.dart';
@@ -25,9 +26,9 @@ final _logger = Logger('genkit_openai');
 
 /// Core Genkit plugin implementation for OpenAI-compatible APIs.
 ///
-/// Automatically discovers models from the OpenAI API (when no custom
-/// [baseUrl] is set) and registers them in the Genkit action registry.
-/// Additional models can be provided via [customModels].
+/// Registers nothing up front beyond [customModels]: [resolve] builds a model
+/// on demand for any id. Discovery runs in [list], against whichever host
+/// [baseUrl] names, and returns metadata rather than registered actions.
 class OpenAIPlugin extends GenkitPlugin {
   final String _pluginName;
 
@@ -55,6 +56,15 @@ class OpenAIPlugin extends GenkitPlugin {
   /// Optional HTTP client for dependency injection and testing.
   final http.Client? httpClient;
 
+  /// Looks up an environment variable, for the API key fallback.
+  ///
+  /// Injectable so a test can run against a known-empty environment: the
+  /// package's own suite needs `OPENAI_API_KEY` exported for
+  /// `integration_test.dart`, which would otherwise make the no-key tests
+  /// pass or fail depending on the developer's shell.
+  @visibleForTesting
+  final String? Function(String name) configVar;
+
   /// Creates an [OpenAIPlugin].
   ///
   /// Provide either [apiKey] or [apiKeyProvider], but not both.
@@ -66,6 +76,7 @@ class OpenAIPlugin extends GenkitPlugin {
     this.customModels = const [],
     this.headers,
     this.httpClient,
+    this.configVar = getConfigVar,
   }) : _pluginName = name {
     if (name.isEmpty || name.contains('/')) {
       throw GenkitException(
@@ -83,12 +94,10 @@ class OpenAIPlugin extends GenkitPlugin {
 
   /// Registers actions that need neither network access nor a key.
   ///
-  /// Deliberately does no I/O. The registry treats an `init()` failure as
-  /// fatal and does not cache it (`registry.dart:32-40`), and every
-  /// `listActions()` call initializes plugins outside its per-plugin
-  /// try/catch - so throwing here takes down the whole Dev UI, including
-  /// `/api/__health`. Model discovery belongs in [list], where a failure
-  /// degrades instead.
+  /// Deliberately does no I/O. A throw here is not cached and is not caught
+  /// per-plugin, so it fails every `listActions()` call for the whole
+  /// registry - taking down the Dev UI, `/api/__health` included. Model
+  /// discovery belongs in [list], where a failure degrades instead.
   ///
   /// Models are not registered up front: [resolve] builds them on demand for
   /// any id, so nothing is lost by staying offline here.
@@ -130,7 +139,7 @@ class OpenAIPlugin extends GenkitPlugin {
     if (config == null) {
       throw GenkitException(
         '[$_pluginName] API key is required. Provide it via apiKey or apiKeyProvider '
-        'in the plugin constructor, or set the OPENAI_API_KEY environment variable.',
+        'in the plugin constructor, or set the $_apiKeyEnvVar environment variable.',
         status: StatusCodes.INVALID_ARGUMENT,
       );
     }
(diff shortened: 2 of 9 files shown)


packages/genkit_openai/test/openai_plugin_startup_test.dart
```diff
@@ -0,0 +1,461 @@
...
+      );
+
+      final actions = await plugin.init();
+
+      expect(actions, hasLength(1));
+      expect(actions.first.name, 'openai/my-model');
+      expect(recorder.requests, isEmpty);
+    });
+
+    test('listActions succeeds with no key and no network', () async {
```
cabljac, line 116: These two tests (this one and "a missing key still fails at generate time") only pass with `OPENAI_API_KEY` unset. With it exported:

```
OPENAI_API_KEY=sk-dummy dart test test/openai_plugin_startup_test.dart
+16 -2: Some tests failed.
```

`integration_test.dart` in the same package needs the var exported, so anyone running the package suite the documented way gets two red. CI is green only because `dart_ci.yml` never sets it. Suggest making the env lookup injectable on `OpenAIPlugin` (defaulting to `getConfigVar`) so these can pass an empty one.

Related: `apiKey: ''` short-circuits the env fallback at `openai_plugin.dart:150`, and the resulting error still says "or set the OPENAI_API_KEY environment variable" even when it is set. Worth a test either way.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -141,60 +126,105 @@ class OpenAIPlugin extends GenkitPlugin {
...
       );
     }
+
+    // Curated models are listed even when discovery omits them, and custom
+    // models are always listed - they need no discovery to be valid.
+    final ids = <String>{
+      ...discovered,
+      ...knownChatModels,
+      ...customModels.map((m) => m.name),
+    };
```
cabljac, line 235: This merges the curated catalog for every instance, so `openAI(name: 'groq', baseUrl: ...)` now lists `groq/gpt-5.5`, `groq/o4-mini` etc. On `main` a custom host listed exactly what its `/models` returned plus `customModels`. JS keeps the curated set in the `openai` vendor plugin only (`openai/index.ts:197-208`), Go likewise (`openai/openai.go:346-350`).

#419 already has the fix (`ddfdb8e`, `baseUrl == null` guard). Can it move down here so this PR is correct on its own, and the assertion at `startup_test.dart:399` flips?

packages/genkit_openai/lib/src/utils.dart
```diff
@@ -120,6 +123,20 @@ bool _supportsToolsByHeuristics(String id) {
...
 }
 
+/// Whether a `gpt-N[.M]` id is version 4.1 or newer.
+bool _isVisionCapableGptVersion(String id) {
+  final match = _gptVersionPattern.firstMatch(id);
+  if (match == null) return false;
+
+  final major = int.tryParse(match.group(1)!);
+  if (major == null) return false;
+  if (major >= 5) return true;
```
cabljac, line 133: `gpt-35-turbo` (Azure deployment name for 3.5) parses major 35 here and now reports `media: true`. On `main` it was false.

```
HEAD: gpt-35-turbo media=true   gpt-35-turbo-16k media=true
MAIN: gpt-35-turbo media=false  gpt-35-turbo-16k media=false
```

Bounding major (`< 10`) or special-casing 35 fixes it, plus `expect(supportsVision('gpt-35-turbo'), false)`.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -141,60 +126,105 @@ class OpenAIPlugin extends GenkitPlugin {
...
-    return apiKey;
+    return apiKey ?? _apiKeyEnvVars.map(getConfigVar).nonNulls.firstOrNull;
+  }
+
+  /// Client config for discovery, or null when no key is available.
+  ///
+  /// Lets [list] skip a request it knows would 401, without duplicating the
+  /// key resolution that [_resolveClientConfig] does - notably without
+  /// invoking [apiKeyProvider] twice, which for a provider that mints a token
+  /// per call would double the cost of every Dev UI poll.
```
cabljac: The registry caches the `list()` future for the life of the process (`registry.dart:164-174`, cleared only on throw), so this per-poll cost does not exist. Same for the test name at `startup_test.dart:327` ("once per listing, not twice").

The flip side: because `list()` now never throws, an app that starts offline or before the key is exported is stuck on the curated catalog until restart. I checked with a client that fails the first listing and serves `gpt-new` after: second `listActions()` made zero requests, never surfaced it. JS caches success only, Go relists every poll. `genkit_anthropic` has the same behaviour so this is consistent, not worse, and it may really be a registry issue (do not cache a degraded result). Do you agree? If so I am happy to file that separately and we just fix the comments and test name here.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -77,45 +81,26 @@ class OpenAIPlugin extends GenkitPlugin {
...
-    }
-
-    // Register custom models
-    for (final model in customModels) {
-      actions.add(_createModel(model.name, model.info));
-    }
-
-    return actions;
-  }
+  Future<List<Action>> init() async => [
```
cabljac, line 105: Small behaviour question. With no key, JS throws at `init()` (openai-node constructor, `openai/index.ts:197` -> `index.ts:133-138`) and Go panics in `Init` (`openai.go:324-326`). Here the key is only needed at `generate()`. I think lazy is the better default for the Dev UI case, and it matches `genkit_anthropic`, but the README should not call it parity since it is more permissive than both. Just flagging as a deliberate divergence to state rather than a bug.

(On a bad key the difference is smaller than it looks: JS/Go also keep their curated set registered, they just log at error and retry next poll.)

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -77,45 +81,26 @@ class OpenAIPlugin extends GenkitPlugin {
     }
   }
 
+  /// Registers actions that need neither network access nor a key.
+  ///
+  /// Deliberately does no I/O. The registry treats an `init()` failure as
+  /// fatal and does not cache it (`registry.dart:32-40`), and every
```
cabljac: nit: the citation is accurate today but line numbers into another package will rot. Say what the constraint is ("a throw here is not cached, and fails every `listActions()` call") and drop the reference.

Also the class dartdoc at `:28-29` still says discovery happens "when no custom baseUrl is set" and "registers them", neither true now: discovery is in `list()`, runs for every host, and returns metadata only.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -362,6 +392,9 @@ class OpenAIPlugin extends GenkitPlugin {
   }
 }
 
+/// Environment variables consulted for the API key, in order.
+const _apiKeyEnvVars = ['OPENAI_API_KEY'];
```
cabljac: nit: one element documented "in order". A `const _apiKeyEnvVar` string is simpler unless a second is coming.

packages/genkit_openai/test/integration_test.dart
```diff
@@ -283,6 +284,62 @@ void main() {
...
+
+      expect(
+        names,
+        containsAll(knownChatModels.map((id) => 'openai/$id')),
+        reason: 'the curated catalog must survive discovery',
+      );
+      expect(
+        names.length,
+        greaterThan(knownChatModels.length),
+        reason: 'discovery should contribute models beyond the catalog',
```
cabljac, line 382: nit: this is tautological, `list()` merges the catalog in regardless of what `/models` returned, so `containsAll` always passes. If the intent is to check the curated ids are real, assert them against the discovered set before the merge.

cabljac (changes requested):
Direction is right, `init()` doing I/O was the actual bug and the Dev UI fix is real. Three things I would like sorted before this lands as the stack base: the startup tests only pass when `OPENAI_API_KEY` is unset, custom hosts now advertise the OpenAI catalog (#419 already fixes this, it belongs here), and `gpt-35-turbo` picks up vision. Two lifecycle questions inline that I do not think are blockers but want a deliberate answer on since the README frames this as parity.

PR body: I could not find the "Genkit rejects image inputs when `media: false`" gating anywhere in `packages/genkit` (`supports` is set in `model.dart:102` and `remote_model.dart:70`, never read). Might be the Dev UI, worth softening to that if so.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Moved it down — `a536e67` is now the second commit here, and `startup_test.dart:399` flips with it. 414 is correct standing alone.

419 and 422 rebased on top, all green.

packages/genkit_openai/lib/src/utils.dart
adesinah: Good catch. Fixed in #422 rather than here — that PR deletes this heuristic for a curated catalog, so bounding the major would have been dead code a commit later.

`gpt-35-turbo` and `gpt-35-turbo-16k` are now alternate spellings on the `gpt-3.5-turbo` entry, so they resolve to text-only. They stay out of `versions`, since OpenAI doesn't serve those names. Test added.

packages/genkit_openai/test/openai_plugin_startup_test.dart
adesinah: Fixed. `configVar` is now injectable on `OpenAIPlugin` (defaults to `getConfigVar`), and the two no-key tests pass a stub that returns null.

`OPENAI_API_KEY=sk-dummy dart test test/openai_plugin_startup_test.dart` → +20, was +16 -2.

Also fixed the `apiKey: ''` case you mentioned — a blank key is treated as absent now, so it falls through to the env var instead of short-circuiting, and the error message interpolates the var name.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: You're right, and I'd forgotten the registry caches it. Reworded to per-listing, and dropped the poll framing from the test comment.

Agree on the degraded-result caching too — that's a registry bug, not something to work around here. Please do file it separately.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Agreed, stated it in the README now rather than implying parity — says it's deliberately more permissive than JS and Go, and why.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Both fixed — the doc now states the constraint instead of citing line numbers, and the class dartdoc says what init/list/resolve actually do now.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Done, it's a `const _apiKeyEnvVar` string now.

packages/genkit_openai/test/integration_test.dart
adesinah: Right, that always passed. Dropped it and added a real check instead: a second listing with `baseUrl` set gets no catalog, so it's pure discovery, and the curated ids have to overlap it. Kept the `greaterThan`.

packages/genkit_openai/test/openai_plugin_startup_test.dart
adesinah: One more from chasing this: `openai_plugin_tool_schema_wire_test.dart` had a live test in it, so that file needed a real key too. Moved it into `integration_test.dart` next to the other tool test — it covers the schema-less case, which the existing one doesn't.

Every non-integration file now passes with a key exported: `OPENAI_API_KEY=sk-dummy dart test $(ls test/*_test.dart | grep -v integration)` → +93.

packages/genkit_openai/lib/src/known_models.dart
```diff
@@ -0,0 +1,60 @@
...
+///   generation ever shipped.
+const List<String> knownChatModels = <String>[
+  // GPT-5 family, newest first. `-pro` is the higher-reasoning tier;
+  // `-mini`/`-nano` trade capability for latency and cost.
+  'gpt-5.5',
+  'gpt-5.5-pro',
+  'gpt-5.4',
+  'gpt-5.4-mini',
+  'gpt-5.4-nano',
+  'gpt-5.4-pro',
```
cabljac, line 44: some of these seem to be Responses API only? Which would 400?

cabljac: Answered upstack: #422 drops every `-pro` tier and says why. Fine here since the stack lands together.

cabljac (approved):
Approving. Two nits, neither blocking.

PR body still says Genkit rejects image inputs when `media: false`. Nothing in `packages/genkit` reads `supports`, so if that gating exists it is the Dev UI; worth saying that instead, since the body becomes the squash message.

nit: `414217e`'s headline is the vision change and its body carries the startup subject, looks swapped. Moot if this squashes.

#423 hooks into `forModel`, which #422 removes, so I'll redo it as a catalog entry on top of #422 once the stack lands. I'll file the registry issue for the cached degraded listing as agreed.

adesinah:
Both fixed/answered.

Body: you're right, nothing in `packages/genkit` reads `supports` — I grepped for reads and there are none, only writes. Reworded to say it misreports rather than blocks, and named the actual consumers (Dev UI picker, `listActions`).

Commit headline: leaving it. Confirmed the repo is squash-only with `squash_merge_commit_title=PR_TITLE`, so that commit never reaches main — and rewriting it means a force-push across an approved PR and two stacked on top. Happy to do it if you'd rather.

pavelgj (approved):
Sanity-checked the full stack locally: dart analyze clean, tests pass, mergeable. init() staying offline and the vision fix both look right. LGTM.

PR #415, opened 2026-09-03
fix(genkit_openai)!: send json_object for schemaless JSON output

### LINKED ISSUE #359: fix(genkit_openai): no json_object mode; documented jsonMode option is dead code
`output.format=json` without a schema sends no response_format at all (`chat.dart:79`); JS falls back to `json_object`. The `jsonMode` option is declared (`chat.dart:55`), documented in the README (`:49`), and never read. Also consider whether `strict: true` + `additionalProperties: false` (`chat.dart:82-86`) is too strict vs JS.

adesinah:
## Problem

Three defects in `buildOpenAIResponseFormat` (`lib/src/chat.dart`):

1. **Schemaless JSON asked for nothing.** `outputFormat: 'json'` without a
   schema sent no `response_format`. Genkit adds no prompt instructions in that
   case either, so nothing told the model to emit JSON.
2. **`jsonMode` was dead code** — declared, documented in the README, never
   read. Its test only checked that the option parsed.
3. **`strict: true` rejected ordinary schemas.** Strict mode needs
   `additionalProperties: false` on every object and every property in
   `required`; the plugin injected it only at the top level, and schemantic
   omits optional fields from `required`. Any schema with an optional field
   returned a 400.

## Solution

One function covering the whole output config, mirroring the JS plugin:

- `format: 'json'` + schema → `json_schema`, schema verbatim, `strict: false`
- `format: 'json'`, no schema → `json_object`
- `format: 'text'` → `text`
- `jsonMode: true` alone → `json_object`
- nothing requested → no `response_format` key, so compatible hosts still work

Output config wins over `jsonMode`, and the caller's schema is no longer
mutated.

## Breaking change

`strict: true` -> `strict: false` on `json_schema`. With strict, OpenAI
guaranteed the reply conformed to the schema; it no longer does, and the schema
becomes advisory. Schemas that already satisfied strict lose that guarantee —
schemas that did not stop returning 400.

JS omits the flag entirely, which is what we want, but
`ResponseFormat.jsonSchema` declares `bool strict = true` and always
serializes it, so `false` is the closest Dart gets.

The title carries `!` so `tools/version.dart` bumps the major rather than
shipping this as a patch — a caller relying on guaranteed conformance needs to
see it in the changelog.

Second behaviour change, non-breaking but worth stating: a schemaless
`outputFormat: 'json'` request now sends `json_object` to **every** host,
compat backends included, where before it sent no `response_format` at all.
Hosts vary in what they accept, so there is a wire test pinning it.

## Testing

175 pass, `dart analyze --fatal-infos` clean, and the full workspace
(`melos exec --dir-exists=test -- dart test`) is green. There were zero wire
tests asserting `response_format` before this — which is how three defects
coexisted. The new wire suite fails against the pre-fix code, so it isn't
decoration.

**Live tests run against the real API: 14/14 pass.** That includes
`an output schema with an optional field is accepted`, which is the only thing
that can verify defect 3 — a mock cannot prove OpenAI stopped rejecting the
schema. Defect 1's `schemaless json output returns parseable JSON` passed too.

```bash
OPENAI_API_KEY=sk-... dart test test/integration_test.dart
# 00:25 +14: All tests passed!
```

## Notes

- `jsonMode` is a divergence from JS, which has no such option. Wiring it up beat
  deleting it: a documented option that silently does nothing is worse.
- Dropping `strict` means output is best-effort JSON matching the schema, as in
  JS — no exact-conformance guarantee.

Fixes #359

### DIFF
--- a/packages/genkit_openai/README.md
+++ b/packages/genkit_openai/README.md
@@ -67,7 +67,6 @@ final response = await ai.generate(
   config: OpenAIChatOptions(
     temperature: 0.7,
     maxTokens: 100,
-    jsonMode: false,
   ),
 );
 ```
@@ -297,9 +296,49 @@ The `OpenAIChatOptions` class supports the following options:
 - `frequencyPenalty` (double?, -2.0 to 2.0) - Frequency penalty
 - `seed` (int?) - Seed for deterministic sampling
 - `user` (String?) - User identifier for abuse detection
-- `jsonMode` (bool?) - Enable JSON mode
+- `jsonMode` (bool?) - Forces `{"type": "json_object"}`. Only consulted when Genkit's own output config does not already imply JSON; see [JSON output](#json-output)
 - `visualDetailLevel` (String?, 'auto'|'low'|'high') - Visual detail level for images
 - `version` (String?) - Model version override
+
+### JSON output
+
+There are three ways to get JSON back, in order of preference:
+
+```dart
+// 1. A schema - the model is constrained to the shape and `output` is typed.
+final response = await ai.generate(
+  model: openAI.model('gpt-4o'),
+  prompt: 'Describe a book.',
+  outputSchema: Book.$schema,
+);
+print(response.output!.title);
+
+// 2. JSON with no particular shape.
+final response = await ai.generate(
+  model: openAI.model('gpt-4o'),
+  prompt: 'Return a JSON object with keys "name" and "age".',
+  outputFormat: 'json',
+);
+
+// 3. The provider flag directly, for callers not using Genkit's output config.
+final response = await ai.generate(
+  model: openAI.model('gpt-4o'),
+  prompt: 'Reply with a JSON object.',
+  config: OpenAIChatOptions(jsonMode: true),
+);
+```
+
+`outputSchema` sends `response_format: {"type": "json_schema"}`; the other two
+send `{"type": "json_object"}`. Output config wins when both are set, so
+`jsonMode` never overrides a schema.
+
+OpenAI rejects `json_object` unless the conversation also asks for JSON, so
+options 2 and 3 need the prompt to say so. Option 1 does not.
+
+Schemas are sent as authored. The plugin does not set `strict`, because
+OpenAI's strict mode requires every property to appear in `required` and
+`additionalProperties: false` on every object — which rejects ordinary schemas
+that have optional fields.
 
 ## Custom Headers
 
--- a/packages/genkit_openai/README.md
+++ b/packages/genkit_openai/README.md
@@ -296,7 +296,7 @@ The `OpenAIChatOptions` class supports the following options:
 - `frequencyPenalty` (double?, -2.0 to 2.0) - Frequency penalty
 - `seed` (int?) - Seed for deterministic sampling
 - `user` (String?) - User identifier for abuse detection
-- `jsonMode` (bool?) - Forces `{"type": "json_object"}`. Only consulted when Genkit's own output config does not already imply JSON; see [JSON output](#json-output)
+- `jsonMode` (bool?) - Forces `{"type": "json_object"}`. Only consulted when Genkit's own output config says nothing about the format; any explicit `outputFormat` wins, `'text'` included. See [JSON output](#json-output)
 - `visualDetailLevel` (String?, 'auto'|'low'|'high') - Visual detail level for images
 - `version` (String?) - Model version override
 
--- a/packages/genkit_openai/test/integration_test.g.dart
+++ b/packages/genkit_openai/test/integration_test.g.dart
@@ -143,3 +143,77 @@ base class _PersonSchemaTypeFactory extends SchemanticType<PersonSchema> {
     dependencies: [],
   );
 }
+
+/// A schema with an optional field.
+///
+/// `nickname` is nullable, so schemantic leaves it out of `required` - which
+/// is precisely what OpenAI's strict mode rejects. The plugin used to send
+/// `strict: true`, so this shape returned a 400.
+base class ProfileSchema {
+  /// Creates a [ProfileSchema] from a JSON map.
+  factory ProfileSchema.fromJson(Map<String, dynamic> json) =>
+      $schema.parse(json);
+
+  ProfileSchema._(this._json);
+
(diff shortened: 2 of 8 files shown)


packages/genkit_openai/test/integration_test.g.dart
```diff
@@ -143,3 +143,77 @@ base class _PersonSchemaTypeFactory extends SchemanticType<PersonSchema> {
...
+/// `strict: true`, so this shape returned a 400.
+base class ProfileSchema {
+  /// Creates a [ProfileSchema] from a JSON map.
+  factory ProfileSchema.fromJson(Map<String, dynamic> json) =>
+      $schema.parse(json);
+
+  ProfileSchema._(this._json);
+
+  ProfileSchema({required String name, String? nickname}) {
+    _json = {'name': name, 'nickname': ?nickname};
```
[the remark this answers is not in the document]
adesinah: Thanks — this one is a false positive, so no change here.

`?nickname` inside a collection literal is Dart's **null-aware element** syntax, not a stray prefix. It omits the entry entirely when the expression is null, which is exactly the semantics your suggestion spells out longhand. This package targets `sdk: ^3.10.0` and the file analyzes clean:

```
$ dart analyze --fatal-infos test/integration_test.g.dart
Analyzing integration_test.g.dart...
No issues found!
```

cabljac:
Core fix is right. `jsonMode` is genuinely wired now (`openai_plugin.dart:252`, pinned by the wire test at `:159`) and it matches JS. Few things before it lands.

The `text` branch at `chat.dart:116` is the main one. It sends `response_format` on a path where nothing needs it, and no test covers it. Either drop the branch or add a wire test pinning what it sends.

`chat.dart:100`: on a compat `baseUrl`, `gpt-4` now gets `json_object` where pre-PR it got nothing. OpenAI 400s on that alias, and compat hosts vary.

```
pre:  (no response_format)
post: response_format: {type: json_object}
```

Just flagging as a behaviour change, but given genkit_openai targets compat providers I'd want a test pinning it.

`chat.dart:112`: `strict:false` is always serialized, so the conformance guarantee is gone. There's no version bump in the PR (pubspec still 0.4.1), and this wants one plus a note.

nit: `openai_plugin_chat_test.dart:92`, the `r'\$defs'` assertion is vacuous. I removed the flatten logic and `resolves $defs` still passed, only the line-74 test failed.

Two smaller ones:

- `chat.dart:55-60` dartdoc and `README:227` both say jsonMode wins unless output implies JSON, but `outputFormat: 'text'` silently suppresses it. Docs and code disagree.
- `integration_test.dart:250-275` asserts `response.output!.name` under `strict:false`. If the field is omitted that throws a TypeError inside `expect` rather than failing as an assertion, so the test misreports its own failure.

adesinah:
All six done.

**`text` branch.** Kept it rather than dropped — it mirrors JS (`compat-oai/src/model.ts:622`), so removing it would be a divergence. Added two wire tests: one pinning `{type: text}`, one pinning that it beats `jsonMode`.

**Compat `json_object`.** You're right that it's a behaviour change, and it applies to every host, not just `gpt-4`. Wire test added against a `baseUrl` backend, and called out in the body.

**`strict: false`.** Retitled `fix(genkit_openai)!:` so `version.dart` bumps the major instead of a patch, with a Breaking change section explaining what's lost. Confirmed we can't do better: `ResponseFormat.jsonSchema` declares `bool strict = true`, non-nullable, so JS's omission isn't reachable through the factory.

**`\$defs` assertion.** Good catch, it was checking for a literal backslash. Fixed to `r'$defs'` and verified it now fails when I remove the flatten — both that test and the line-74 one go red.

**Docs vs code.** Code was right, docs were wrong. Both the dartdoc and README now say any explicit `outputFormat` wins, `'text'` included.

**Integration TypeError.** Asserting on `toJson()` fields instead of the typed getter, so a missing field fails as an assertion rather than crashing inside `expect`.

Also rebased onto current main — it had moved five commits, including #413, which turns model errors into failed responses rather than throws. Nothing in this PR asserted on that, so no test changes, but it was stale.

175 pass, `--fatal-infos` clean, and the full `melos exec -- dart test` is green.

adesinah:
Live tests are run now — 14/14 against the real API, including `an output schema with an optional field is accepted`. That's the one that matters: it's the only way to confirm `strict: false` actually stopped the 400 on schemas with optional fields, since a mock can't prove it. `schemaless json output returns parseable JSON` passed too. Body updated; it previously said they hadn't been run.

Also fixed a CI break I caused while addressing your review: `OpenAIChatOptions` is schemantic-annotated, so rewriting the `jsonMode` dartdoc changed `chat.g.dart` and I hadn't regenerated. "Verify clean git tree" caught it — `analyze`, `test` and `format` all pass with a stale .g.dart, which is worth knowing for doc-only edits. Regenerated in `9e2461a`, all checks green.

packages/genkit_openai/README.md
```diff
@@ -297,10 +296,50 @@ The `OpenAIChatOptions` class supports the following options:
...
+```
+
+`outputSchema` sends `response_format: {"type": "json_schema"}`; the other two
+send `{"type": "json_object"}`. Output config wins when both are set, so
+`jsonMode` never overrides a schema.
+
+OpenAI rejects `json_object` unless the conversation also asks for JSON, so
+options 2 and 3 need the prompt to say so. Option 1 does not.
+
+Schemas are sent as authored. The plugin does not set `strict`, because
```
cabljac, line 338: Schemas aren't quite sent as authored: they go through `flatten()`, which inlines `$ref` and strips `$defs` (schemantic/lib/src/flatten.dart:47-73). And we do set `strict`, to `false` (chat.dart:119), though that half is behaviourally moot since false is OpenAI's default anyway. Maybe "The plugin sends `strict: false`. Schemas are flattened ($ref/$defs resolved) but otherwise unmodified."

cabljac (approved):
One sentence in the README is out of date, flagged inline. Otherwise this looks good.

packages/genkit_openai/README.md
adesinah: Fixed in fefbadf — took your wording.

ifielker (approved):

PR #416, opened 2026-09-03
feat(agent): support metadataOnly snapshot reads

pavelgj:
Honor GetSnapshotDataInput.metadataOnly: the getSnapshotData action and the
typed read surfaces (Agent.getSnapshot/getSnapshotData, AgentApi.getSnapshot,
DetachedTask.poll) can now read a snapshot's shaped metadata without its state
payload. The runtime reads through the optional SnapshotMetadataReader store
capability when available (loading no state) and falls back to a full read
otherwise, then drops the state and skips the client transform. Heartbeat
expiry shaping still applies. Both bundled stores implement the capability.

Ports the Go reference (SnapshotMetadataReader, WithMetadataOnly).

### DIFF
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -1649,15 +1649,25 @@ Agent<State> defineCustomAgent<State>(
     String? sessionId,
     Map<String, dynamic>? context,
   ) async {
-    if (store is SnapshotMetadataReader) {
-      final reader = store as SnapshotMetadataReader;
-      return snapshotId != null
-          ? reader.getSnapshotMetadata(snapshotId, context: context)
-          : reader.getLatestSnapshotMetadata(sessionId!, context: context);
-    }
-    return store.getSnapshot(
+    // Normalize first so the metadata path enforces the same "exactly one of
+    // snapshotId / sessionId" contract as the full read, rather than crashing
+    // on a null-assertion or silently ignoring one id.
+    final normalized = normalizeGetSnapshotOptions(
       snapshotId: snapshotId,
       sessionId: sessionId,
+    );
+    if (store is SnapshotMetadataReader) {
+      final reader = store as SnapshotMetadataReader;
+      return normalized.snapshotId != null
+          ? reader.getSnapshotMetadata(normalized.snapshotId!, context: context)
+          : reader.getLatestSnapshotMetadata(
+              normalized.sessionId!,
+              context: context,
+            );
+    }
+    return store.getSnapshot(
+      snapshotId: normalized.snapshotId,
+      sessionId: normalized.sessionId,
       context: context,
     );
   }
--- a/packages/genkit/lib/src/ai/agents/agent.dart
+++ b/packages/genkit/lib/src/ai/agents/agent.dart
@@ -47,9 +47,16 @@ import 'session.dart';
 /// ended (e.g. `interrupted`, `length`). When omitted, no per-turn reason is
 /// reported.
 class TurnResult {
-  TurnResult({this.finishReason});
+  TurnResult({this.finishReason, this.finishMessage});
 
   final AgentFinishReason? finishReason;
+
+  /// An optional human-readable message describing why the turn ended.
+  ///
+  /// Carries, for example, the "Reached max turns of N" text from a generation
+  /// that overran its turn limit, so the [SessionRunner] can surface it rather
+  /// than dropping it.
+  final String? finishMessage;
 }
 
 /// Per-turn context handed to the handler passed to [SessionRunner.run].
@@ -404,14 +411,50 @@ class SessionRunner<State> {
           if (cancel?.isCancelled ?? false) {
             lastTurnFinishReason = AgentFinishReason.aborted;
             lastTurnError = null;
+            // Persist the turn as `aborted`. The detached route already wrote
+            // `aborted` via `_abortSnapshotInStore`, in which case the
+            // abort-aware mutator makes this a no-op; but the attached route
+            // (`AgentTurn.abort()` -> token cancel) has written nothing, so
+            // without this the trailing `invocationEnd` snapshot would persist a
+            // half-finished turn as `completed` and later be picked as a resume
+            // point.
+            final snapshotId = await maybeSnapshot(
+              status: 'aborted',
+              snapshotId: turnSnapshotId,
+              finishReason: AgentFinishReason.aborted,
+            );
             _notifyEndTurn(
-              _lastSnapshot?.snapshotId,
+              snapshotId ?? _lastSnapshot?.snapshotId,
               AgentFinishReason.aborted,
             );
             return true;
           }
 
           final finishReason = turnResult?.finishReason;
+
+          // A turn that resolved `aborted` *without* the token being cancelled
+          // is not a cooperative cancel: it is an overrun (e.g. the generate
+          // loop hit `maxTurns`). Route it to the failure path so the reason
+          // (e.g. "Reached max turns of N") surfaces as an error instead of
+          // being silently dropped as a success with a null message.
+          if (finishReason == AgentFinishReason.aborted) {
+            lastTurnFinishReason = AgentFinishReason.failed;
+            lastTurnError = toErrorDetails(
+              GenkitException(
+                turnResult?.finishMessage ?? 'Turn aborted.',
+                status: StatusCodes.ABORTED,
+              ),
+            );
+            final snapshotId = await maybeSnapshot(
+              status: 'failed',
+              error: lastTurnError,
+              snapshotId: turnSnapshotId,
+              finishReason: AgentFinishReason.failed,
+            );
+            _notifyEndTurn(snapshotId, AgentFinishReason.failed);
+            return true;
+          }
+
           lastTurnFinishReason = finishReason;
           lastTurnError = null;
 
@@ -898,7 +941,7 @@ final class _InProcessTransport extends AgentTransport {
   TurnStream runTurn(
     AgentInput input,
     AgentInit init, {
-    required CancellationToken cancel,
+    CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) {
     final bidi = _startBidi(input, init, context: context, cancel: cancel);
@@ -909,7 +952,7 @@ final class _InProcessTransport extends AgentTransport {
   Future<AgentOutput>? run(
     AgentInput input,
     AgentInit init, {
-    required CancellationToken cancel,
+    CancellationToken? cancel,
     Map<String, dynamic>? context,
   }) {
     return _startBidi(input, init, context: context, cancel: cancel).onResult;
@@ -1137,7 +1180,9 @@ Agent<State> defineCustomAgent<State>(
           // Capture the disposer so a reused, long-lived `ctx.cancel` token
           // doesn't accumulate one stranded listener (pinning this turn's
           // controller and closure) per turn. Disposed when the flow settles.
-          final unlinkCancel = ctx.cancel.onCancel(cancelController.cancel);
+          // `link` forwards the cancellation reason so a caller-supplied
+          // `controller.cancel('...')` survives the hop into this turn's token.
+          final unlinkCancel = ctx.cancel?.link(cancelController);
           final cancelToken = cancelController.token;
           void Function()? unsubscribe;
           // Background heartbeat timer for the detached snapshot. Started in
@@ -1302,7 +1347,7 @@ Agent<State> defineCustomAgent<State>(
               // The turn has settled (the snapshot reached a terminal status),
               // so stop refreshing its heartbeat.
               stopHeartbeat();
-              unlinkCancel();
+              unlinkCancel?.call();
               unsubscribe?.call();
               offArtifactAdded();
               offArtifactUpdated();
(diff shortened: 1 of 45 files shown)


packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -1623,26 +1639,66 @@ Agent<State> defineCustomAgent<State>(
...
+      return snapshotId != null
+          ? reader.getSnapshotMetadata(snapshotId, context: context)
+          : reader.getLatestSnapshotMetadata(sessionId!, context: context);
+    }
+    return store.getSnapshot(
       snapshotId: snapshotId,
       sessionId: sessionId,
       context: context,
     );
+  }
```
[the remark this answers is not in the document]
pavelgj: Done in b1c74625. `readSnapshotMetadata` now runs `normalizeGetSnapshotOptions` first and dispatches on the normalized ids, so the capability path enforces the same 'exactly one of snapshotId/sessionId' contract (no null-assertion crash, no silently-ignored id).

packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -365,6 +399,24 @@ final class InMemorySessionStore
...
+  }
+
+  @override
+  Future<SessionSnapshot?> getLatestSnapshotMetadata(
+    String sessionId, {
+    Map<String, dynamic>? context,
+  }) async {
+    final snap = await getSnapshot(sessionId: sessionId, context: context);
+    return snap != null ? stripSnapshotState(snap) : null;
+  }
```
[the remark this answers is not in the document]
pavelgj: Done in b1c74625. `getSnapshotMetadata`/`getLatestSnapshotMetadata` now read `_snapshots` directly and `stripSnapshotState` once, skipping the full-getSnapshot deep clone of the state payload.

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -281,6 +282,24 @@ class FileSessionStore implements SessionStore, SnapshotChangeNotifier {
...
+  }
+
+  @override
+  Future<SessionSnapshot?> getLatestSnapshotMetadata(
+    String sessionId, {
+    Map<String, dynamic>? context,
+  }) async {
+    final snap = await getSnapshot(sessionId: sessionId, context: context);
+    return snap != null ? stripSnapshotState(snap) : null;
+  }
```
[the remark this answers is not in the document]
pavelgj: Done in b1c74625. `getSnapshotMetadata` drops 'state' off the decoded map before parsing (no double serialize). `getLatestSnapshotMetadata` takes the pointer fast-path, but I kept the self-heal guard: it only honors the pointer when the leaf still exists and belongs to the session, otherwise falls through to the scan + strip (matching `_latestSnapshotForSession`).

packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -50,6 +50,13 @@ SessionSnapshot _cloneSnapshot(SessionSnapshot snapshot) =>
       _deepClone(snapshot.toJson()) as Map<String, dynamic>,
     );
 
+/// Returns a copy of [snapshot] with its `state` dropped, for a metadata-only
+/// read. Every other field is carried through as stored.
```
apascal07: Not every other field. `snapshotSessionId()` falls back to `state.sessionId`, so dropping `state` erases the session identity of a row that carries it only there. Both bundled stores support such rows (`FileSessionStore._saveSnapshotUnlocked` writes the pointer off `snapshotSessionId(result)`), and the new heartbeat-expiry test writes one.

Two consequences. `FileSessionStore.getLatestSnapshotMetadata` evaluates `snapshotSessionId(meta) == sessionId` as `null == sessionId`, judges the pointer stale on every call, and falls back to the full prefix scan: the O(n) cost the fast path exists to avoid, paid on every poll tick. And the row that comes back has `sessionId` null, so a client branching on it sees nothing.

Go has no equivalent hazard because `SessionSnapshot.SessionID` is a top-level field that survives `State = nil`.

packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -50,6 +50,13 @@ SessionSnapshot _cloneSnapshot(SessionSnapshot snapshot) =>
...
+SessionSnapshot stripSnapshotState(SessionSnapshot snapshot) {
+  final json = Map<String, dynamic>.from(snapshot.toJson())..remove('state');
```
apascal07, line 65: `Map.from` copies only the top level, so every nested value except `state` stays shared with the argument. `InMemorySessionStore.getSnapshotMetadata` calls this on the row still held in `_snapshots`, where the full read deep-clones via `_cloneSnapshot`.

So `(await store.getSnapshotMetadata(id))!.error!.message = '...'` writes into the stored row: `toJson()` returns the live `_json`, and `AgentErrorInfo` wraps the same nested map. A later full read returns the mutated error.

Go guards this explicitly: `snapshotMetadata` copies `HeartbeatAt` and `Error` "so a caller that edits either result never reaches the store's row".

packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -115,6 +122,33 @@ abstract interface class SessionStore {
   });
 }
 
+/// Optional capability layered on [SessionStore] for answering a metadata-only
+/// read (`GetSnapshotDataInput.metadataOnly`) without loading the row's state.
+///
+/// A store that does not implement it is still correct: the runtime reads the
+/// row in full and drops the state, paying to load a conversation history it
+/// will not use. A projecting store (SQL, Firestore) implements this to skip
```
apascal07, line 140: `FirestoreSessionStore` does not implement `SnapshotMetadataReader`: `packages/genkit_google_cloud/lib/firestore_session_store.dart:360` still declares only `SessionStore, SnapshotChangeNotifier`. The store this sentence points at is the one bundled store that does not get the projection.

For Firestore the fallback is not one wasted document read. `getSnapshot` opens a read-only transaction and reconstructs the checkpoint shards plus the diff chain for the whole conversation, and a `poll(metadataOnly: true)` pays that on every tick. Go's `FirestoreSessionStore` implements both methods as single document reads outside any transaction.

packages/genkit/lib/src/ai/agents/session.dart
```diff
@@ -365,6 +399,36 @@ final class InMemorySessionStore
...
+    Map<String, dynamic>? context,
+  }) async {
+    // Strip directly off the stored row so we never deep-clone the (possibly
+    // large) state payload the way a full `getSnapshot` would.
+    final snap = _snapshots[snapshotId];
+    return snap != null ? stripSnapshotState(snap) : null;
+  }
+
+  @override
+  Future<SessionSnapshot?> getLatestSnapshotMetadata(
```
apascal07, line 435: Neither metadata method validates its id, while `getSnapshot` runs `normalizeGetSnapshotOptions` (empty id rejected with INVALID_ARGUMENT, `assertValidSessionId` on a session lookup). `getLatestSnapshotMetadata('')` matches nothing and returns null, which reads as "this session has no snapshots", where `getSnapshot(sessionId: '')` throws. The runtime normalizes first, so this only bites direct store callers, but the interface doc promises "the same resolution, the same null-if-none contract".

Go validates on both paths: `InMemorySessionStore.GetLatestSnapshotMetadata` returns "session ID is empty", and `FileSessionStore.latest` applies `validateSessionID` for the metadata read and the full read alike.

packages/genkit/lib/src/ai/agents/session_io.dart
```diff
@@ -281,6 +282,54 @@ class FileSessionStore implements SessionStore, SnapshotChangeNotifier {
...
+    final file = await _fileFor(snapshotId, context);
+    if (!file.existsSync()) return null;
+    try {
+      // Drop `state` off the decoded map before parsing, so the (possibly
+      // large) state payload is never materialized into a SessionState and we
+      // avoid the re-serialize a `stripSnapshotState` on the full row would do.
+      final json = jsonDecode(await file.readAsString()) as Map<String, dynamic>
+        ..remove('state');
+      return SessionSnapshot.fromJson(json);
+    } catch (_) {
```
apascal07: The full read path propagates a corrupt or unreadable file (`_snapshotById` lets the `jsonDecode` throw); this one turns it into `null`, which the caller cannot tell apart from "no such snapshot".

A truncated file makes `getSnapshotData(snapshotId: id)` throw and `getSnapshotData(snapshotId: id, metadataOnly: true)` return null, so `DetachedTask.poll(metadataOnly: true)` counts 10 misses and raises `StateError('Snapshot <id> not found.')` instead of surfacing the corruption. A permission error on the prefix directory behaves the same way.

Go keeps the two paths identical here: `readMetadataAt` and `readAt` both go through `readJSONAt`, so an unmarshal error comes back either way.

packages/genkit/lib/src/ai/agents/agent_core.dart
```diff
@@ -95,10 +95,16 @@ abstract base class AgentTransport {
...
+  ///
+  /// When [metadataOnly] is set the returned snapshot carries the shaped
+  /// metadata (status, finish reason, parent, session, timestamps, error) with
+  /// no state payload, so a poll that only branches on where a task stands can
+  /// skip loading the conversation history.
   Future<SessionSnapshot?> getSnapshot({
     String? snapshotId,
     String? sessionId,
     Map<String, dynamic>? context,
+    bool metadataOnly = false,
```
apascal07: `AgentTransport` is exported from `genkit.dart` and is `base`, so it can be extended outside the library. An override must declare every named parameter of the method it overrides, so this addition fails to compile every existing subclass, in a 0.16.x release. The three in-repo test transports needed exactly that edit.

If external transports are supported, this wants either a major bump or a seam that leaves the abstract signature alone (a separate `getSnapshotMetadata` hook defaulting to `getSnapshot`).

apascal07 (approved):

packages/genkit/lib/src/ai/agents/session.dart
pavelgj: Fixed. `stripSnapshotState` now promotes the resolved `sessionId` to the top level before dropping `state`, so a row that carried its id only under `state` comes back with `sessionId` set. That restores client branching and keeps the `FileSessionStore` pointer fast path from judging every pointer stale (no more O(n) scan per poll tick). `FileSessionStore.getSnapshotMetadata` does the same promotion on its decoded map. Matches Go, where `SessionID` is a top-level field that survives `State = nil`.

Coverage: `stripSnapshotState` unit test plus per-store metadata tests all assert `meta.sessionId` survives a state-only-sessionId row.

packages/genkit/lib/src/ai/agents/session.dart
pavelgj: Fixed. `stripSnapshotState` now deep-clones the retained fields (the clone runs after `state` is removed, so the large payload is never copied), and `InMemorySessionStore.getSnapshotMetadata` strips directly off the stored row. Mutating a nested field on the result (e.g. `meta.error!.message = ...`) no longer reaches the store's row. This is the equivalent of Go's `snapshotMetadata` copying `HeartbeatAt` and `Error` defensively; deep-cloning the whole state-less map covers every nested field uniformly (Dart wraps a live JSON map, so a value-copy like Go's `meta := *snap` is not available).

Coverage: "getSnapshotMetadata result is isolated from the stored row" mutates `meta.error` and asserts a later read still sees the original.

packages/genkit/lib/src/ai/agents/session.dart
pavelgj: Fixed. `FirestoreSessionStore` now implements `SnapshotMetadataReader`. `getSnapshotMetadata` is a single document read and `getLatestSnapshotMetadata` is a pointer read plus one document read, both outside any transaction and with no shard/diff reconstruction (the snapshot document already carries every metadata field). `_toMetadataSnapshot(_SnapshotDoc)` builds the row with `state` omitted. This drops the per-tick reconstruction cost the fallback would have paid on `poll(metadataOnly: true)`, matching Go's single-document reads. The interface doc now lists all three bundled stores.

Coverage: emulator tests for drop-state, missing -> null, leaf resolution, missing session -> null, and id validation (green against a local Firestore emulator).

packages/genkit/lib/src/ai/agents/session.dart
pavelgj: Fixed. Both metadata methods on all three stores now run the same normalization as `getSnapshot` (`normalizeGetSnapshotOptions`, or Firestore's local equivalent), so an empty/blank id is rejected with `INVALID_ARGUMENT` instead of returning null. `getLatestSnapshotMetadata('')` now throws exactly like `getSnapshot(sessionId: '')`. Matches Go validating on both paths.

Coverage: "metadata reads validate their id like getSnapshot" on each store.

packages/genkit/lib/src/ai/agents/session_io.dart
pavelgj: Fixed. `getSnapshotMetadata` no longer catches decode errors: a corrupt or unreadable file throws, exactly as the full read (`_snapshotById`) does, so a truncated file is distinguishable from "not found" and `poll(metadataOnly: true)` surfaces corruption instead of counting misses to a spurious not-found. Go keeps the two paths identical here via `readJSONAt`.

One deliberate divergence worth flagging: on the latest/pointer path a corrupt target leaf propagates in Dart, whereas Go's `latest` fast path swallows the read error and falls through to the scan. This is a pre-existing property of the Dart file store's full read (`_latestSnapshotForSession` never swallowed a corrupt pointer target); this PR keeps the new metadata path aligned with that full read rather than changing the standing behavior.

Coverage: "getSnapshotMetadata throws on a corrupt file" and "getLatestSnapshotMetadata surfaces a corrupt pointer target like the full read" (asserts both the full read and the metadata read throw).

PR #417, opened 2026-09-03
feat(middleware): async sub-agents and task continuation for agents middleware

pavelgj:
Ports the async sub-agents work from the Go agents middleware (`go/plugins/middleware/exp/agents{,_async,_continue}.go`) to the Dart `genkit_middleware` `agents` middleware. Stacked on top of #416.

## What's new

- **`async` option.** Delegation tools gain a `background` flag, and three shared tools are added: `check_background_tasks`, `wait_for_background_tasks`, `abort_background_tasks`. A background delegation detaches a server-managed sub-agent (`AgentInput.detach`), which persists a pending snapshot and returns a self-contained `"<agent>:<snapshotId>"` taskId immediately; the orchestrator collects results later.
- **taskId / status / name on delegation results.** Server-managed delegations (those whose sub-agent has a `store`) are addressable after the fact; artifacts from such runs are namespaced by the run's snapshot id so re-checks are idempotent (matches Go).
- **`continue_task` tool.** Retries a failed/aborted task from its last saved progress, or follows up on a completed task inside its own session, with the interrupt refusal, empty-follow-up gate, and expired-recovery fence ported over.

## Notes / divergences from Go

- Dart has no blocking wait wire action, so `wait_for_background_tasks` **polls** each task's snapshot on an interval (the same strategy the runtime's own `DetachedTask.wait` uses). It honors `timeoutSeconds` (0/omitted = unbounded, negative = return now) and `waitFor: all|first`.
- Single-isolate, so Go's mutex collapses to plain fields; parallel fetches use `Future.wait`.
- The `continue_task` tool is registered whenever agents are configured (the synchronous `tools` getter can't do async registry lookups to detect server-managed agents up front); it refuses gracefully for a client-managed handle.

## Testing

- 11 new cases in `test/agents_middleware_async_test.dart` (background launch + collect, undetachable refusal, unresolvable/empty taskIds, sync taskId, continue follow-up / empty-follow-up refusal / client-managed rejection, tool presence with/without `async`). Full package suite: 47 passing, analyze clean, formatted.
- Runnable sample `testapps/agents/bin/async_agent_sample.dart`.

## Behavior change to flag

Server-managed sub-agents now namespace artifacts by the shortened snapshot id (`agent_<8hex>/name`) instead of a random 4-char suffix (existing test updated accordingly).


packages/genkit_middleware/lib/src/agents_middleware_async.dart
```diff
@@ -0,0 +1,554 @@
...
+    if (ids.isEmpty) return BackgroundTasksResult(note: _noTaskIdsNote);
+
+    var first = false;
+    switch (input.waitFor) {
+      case null:
+      case _waitForAll:
+        break;
+      case _waitForFirst:
+        first = true;
+      default:
```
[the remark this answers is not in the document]
pavelgj: Not applicable in Dart 3: non-empty `switch` cases no longer fall through and do not need a terminating `break;` (each case implicitly breaks). Adding one would trip this repo's `unnecessary_breaks` lint (enabled in the root `analysis_options.yaml`, treated as fatal by `--fatal-infos`). Confirmed: `dart analyze` is clean and all 47 tests pass, including the `waitFor: "first"` case this switch handles.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
```diff
@@ -0,0 +1,554 @@
...
+        report.error =
+            'The task was aborted before it finished. Continue it with '
+            '$_continueToolName using this taskId to pick up from its last '
+            'saved progress.';
+      case 'expired':
+        report.error =
+            'The background worker stopped reporting progress and is presumed '
+            'dead. Attempt $_continueToolName with this taskId to recover saved '
+            'progress, or delegate the task again.';
+    }
```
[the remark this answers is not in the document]
pavelgj: Same as above: this is valid Dart 3. Non-empty cases do not fall through, so no `break;` is needed; adding them would violate the repo's `unnecessary_breaks` lint. `dart analyze` reports no issues and the `check`/`wait` report-path tests exercise these branches and pass.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
```diff
@@ -0,0 +1,614 @@
...
+        message: Message(
+          role: Role.user,
+          content: [TextPart(text: task)],
+        ),
+        detach: true,
+      );
+      return _foldDetachOutcome(handle, begun.invocationNum, out, words);
+    } catch (e) {
+      return AgentDelegationResult(response: '${words.errPrefix}: $e');
+    }
```
apascal07, line 110: Go has a third arm here that this drops (`maybeDetachRejection`, agents_async.go:270-286). In Dart a rejected detach arrives as a thrown `GenkitException` with `FAILED_PRECONDITION`, so it lands in this catch.

`abortable` falls back to `true` when the agent's metadata does not carry the field, so such an agent passes `_refuseUndetachable`, runs, and is rejected here. The model gets a bare error, the delegation slot stays spent, and nothing tells it the synchronous form would work. Go refunds the slot and names that retry.

packages/genkit_middleware/lib/src/agents_middleware.dart
```diff
@@ -144,79 +351,297 @@ GenerateMiddlewareRef<AgentsOptions> agents({
...
+  /// only case that accepts seeded init state. Unknown metadata is treated as
+  /// not client-managed (the safe default: avoid seeding state an agent might
+  /// reject), matching Go.
+  bool get isClientManaged =>
+      metadata?.stateManagement.value == AgentStateManagement.client.value;
+
+  /// Whether the agent can run in the background (its store supports detach).
+  /// Unknown metadata is treated as abortable (a wrong guess costs a refusal at
+  /// launch time, not a silently missing capability), matching Go.
+  bool get abortable => metadata?.abortable ?? true;
```
apascal07: These two reads are not null-safe. `AgentMetadata.fromJson` is `$schema.parse`, which wraps the map without validating it, so `metadata.abortable` runs `_json['abortable'] as bool` and throws a `TypeError` when the key is absent. `stateManagement` does the same on a missing String.

The `?? true` and the `?.` only cover a null `AgentMetadata`, not a partial map. Both call sites (line 776 below, and agents_middleware_continue.dart:98) sit outside any try, so an agent with partial metadata fails the whole generate call instead of taking the fallback these doc comments promise.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
```diff
@@ -0,0 +1,614 @@
...
+      case _waitForFirst:
+        first = true;
+      default:
+        return BackgroundTasksResult(
+          note:
+              'Unknown waitFor value "${input.waitFor}". Use "$_waitForAll" '
+              '(the default) to wait for every task, or "$_waitForFirst" to '
+              'return when any one settles.',
+        );
+    }
```
apascal07, line 246: Go accepts `""` here alongside `"all"` (agents_async.go:366). An empty string falls to `default`, so a model that sends `"waitFor": ""` gets `Unknown waitFor value ""` back and the tasks are not collected on that turn.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
```diff
@@ -0,0 +1,614 @@
...
+    if (timeoutSeconds < 0) {
+      return _reportTasks(ids, _readSnapshotOnce);
+    }
+
+    // A value large enough to overflow the microsecond arithmetic is treated as
+    // unbounded (the same as 0); without the clamp it wraps into a deadline in
+    // the past and the wait returns instantly (matches Go's maxWaitSeconds).
+    final deadline = (timeoutSeconds > 0 && timeoutSeconds <= _maxWaitSeconds)
+        ? DateTime.now().add(Duration(seconds: timeoutSeconds))
+        : null;
```
apascal07: The clamp bounds `Duration` but not `DateTime`. `_maxWaitSeconds` is about 9.22e12 seconds, and `DateTime`'s valid range ends about 8.64e12 seconds from now.

`timeoutSeconds: 9000000000000` passes the clamp, then `DateTime.now().add(...)` throws `ArgumentError: DateTime is outside valid range` and fails the generate call. The overflow test uses int64 max, which is above the clamp and correctly falls through to unbounded, so this band is untested.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
```diff
@@ -0,0 +1,614 @@
...
+    }
+  }
+
+  // ── Snapshot fetches ──────────────────────────────────────────────────────
+
+  /// Reads a task's snapshot once (the check tool and the wait tool's poll).
+  Future<SessionSnapshot?> _readSnapshotOnce(
+    _AgentHandle handle,
+    String snapshotId,
+  ) => _getSnapshot(handle, snapshotId);
```
apascal07, line 587: The wait poll shares the check tool's fetch, so every 200ms tick is a full read: `resolveSnapshot` with `metadataOnly: false` runs `toClientSnapshot` over the whole conversation. Three tasks over a five second wait is roughly 75 of those, and only the terminal read needs the state.

packages/genkit_middleware/lib/src/agents_middleware.dart
```diff
@@ -225,211 +650,338 @@ class AgentsMiddleware extends GenerateMiddleware {
...
+    }
+
+    if (_async) tools.addAll(_backgroundTaskTools());
+
+    // The continue tool is always registered when agents are configured. Go
+    // registers it only when a sub-agent may be server-managed, but that needs
+    // async agent resolution the synchronous `tools` getter cannot do here, so
+    // it is registered unconditionally and refuses gracefully for a
+    // client-managed handle.
+    tools.add(_continueTool());
```
apascal07, line 715: Registering this unconditionally costs more than a dead tool. `_runContinue` spends a slot in `_beginDelegation` before it checks `isClientManaged`, and that refusal does not call `_releaseDelegation`.

So with `maxDelegations: 5` and store-less sub-agents, five `continue_task` calls exhaust the cap and every real delegation after that is refused. Go cannot reach this, because `anyContinuableAgent` gates the registration.

The system prompt paragraph can be gated even if the tool cannot: `_buildInstructions` runs from the async `generate()` hook, which can resolve handles.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
```diff
@@ -0,0 +1,614 @@
...
+    String taskId, {
+    DateTime? deadline,
+    Future<void>? stopSignal,
+    void Function()? onSettled,
+    CancellationToken? cancel,
+  }) async {
+    var stopped = false;
+    stopSignal?.then((_) => stopped = true);
+
+    BackgroundTaskReport? lastTransient;
```
apascal07: nit: `lastTransient` never survives an iteration. It is assigned `report` on a transient poll and set to null on every other one, both before the returns below, so `lastTransient ?? report` is always `report`. Riding out a blip works because the loop continues; the variable does nothing.

apascal07 (approved):

packages/genkit_middleware/lib/src/agents_middleware_async.dart
pavelgj: Good catch. Added an `on GenkitException` arm in `_launchDelegation`: on `FAILED_PRECONDITION` it now refunds the slot with `_releaseDelegation()` and returns the "cannot run in the background; delegate without background instead" refusal, matching the `_refuseUndetachable` pre-flight wording and Go's `maybeDetachRejection`. Non-precondition exceptions keep the previous bare-error path.

packages/genkit_middleware/lib/src/agents_middleware.dart
pavelgj: Right, `$schema.parse` just wraps the map, so the typed getters throw on a partial map rather than returning null. Switched both `_AgentHandle.isClientManaged` and `abortable` to read the raw map (`metadata?.toJson()['stateManagement' | 'abortable']`) so an absent key takes the documented fallback (`abortable` -> true, `isClientManaged` -> false) instead of a `TypeError`. That covers both out-of-try call sites.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
pavelgj: Added `case '':` alongside `_waitForAll` so an empty string collects on that turn instead of hitting the unknown-value refusal, matching Go's agents_async.go:366.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
pavelgj: Confirmed: the clamp bounds Duration but not DateTime's narrower range. Extracted `_deadlineFor`, which keeps the Duration clamp and also catches the `ArgumentError` from `DateTime.add`, treating that band as unbounded (same as 0). Added a test at 9e12s (under the clamp, over the DateTime range) to cover it, since the existing int64-max test sits above the clamp.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
pavelgj: Split the poll from the settle. Added `_pollSnapshot` (metadataOnly: true) and a `detail` flag on `_reportTask`; wait ticks now read metadata only and skip the state fold and the settled cache, and the loop does a single full `_readSnapshotOnce` on settle to fold the answer. Guarded so a transient blip on that final read falls back to the metadata report we already trust as terminal. The existing "check/wait collect the result" test exercises this path and still asserts the folded response.

packages/genkit_middleware/lib/src/agents_middleware.dart
pavelgj: The tool still registers unconditionally (the sync `tools` getter can't resolve handles), but I fixed both halves of the cost you flagged:

- The client-managed refusal in `_runContinue` now calls `_releaseDelegation()`, so a run of these refusals no longer drains the cap. Added a test: with `maxDelegations: 1`, a client-managed `continue_task` refusal still leaves room for a real delegation.
- Gated the continue system-prompt paragraph behind an `anyContinuable` check computed in the async `generate()` hook (which can resolve handles), so store-less-only setups don't advertise it.

Updated the slot-policy doc comment to reflect that client-managed now refunds. Flagging in case you'd prefer keeping the slot and gating differently.

packages/genkit_middleware/lib/src/agents_middleware_async.dart
pavelgj: Yep, it was always equal to `report` at both returns. Removed it; the loop rides out blips purely by continuing, and both exits just return `report`.

PR #419, opened 2026-09-07
test(genkit_openai): cover the OpenAI-compatible path with a fake host

### LINKED ISSUE #362: test(genkit_openai): compat path has zero test coverage
README advertises Groq/xAI/DeepSeek/Together (`README.md:3,141-201`); no test constructs the plugin with `baseUrl`, `headers`, or registers a `CustomModelDefinition`, and the testapp only uses plain `openAI(apiKey:)`. Either test the compat path (JS has `tests/fake_openai_server.ts` to copy) or soften the README claim. Also: README pins `genkit: ^0.13.0` vs actual `^0.15.1`.

adesinah:
Fixes #362. Stacked on #414.

Ports the JS plugin's `fake_openai_server` (`js/plugins/compat-oai/tests/`) to Dart. It binds a real loopback `HttpServer` rather than a `MockClient`: every other test in this package injects `httpClient`, so the client the plugin builds for itself — the only thing that acts on `baseUrl` — was never covered.

26 tests in `test/openai_plugin_compat_test.dart` covering baseUrl routing, auth and `apiKeyProvider`, custom headers, `CustomModelDefinition`, multiple named backends, streaming, tool calls, and error-status mapping.

Found one bug on the way, fixed in its own commit: `list()` merged the curated OpenAI catalog in regardless of `baseUrl`, so a Groq backend advertised 17 models Groq doesn't serve, each a 404 from the Dev UI. The catalog is now skipped when `baseUrl` is set. `resolve()` still serves any id named explicitly, so nothing becomes unreachable. Worth knowing when reviewing #414, since the bug is in the `list()` that PR introduces.

README: the install block pinned `genkit: ^0.13.0` against an actual `^0.16.1`. The compat section now says what's verified — a local fake host, not each provider's live API.

One thing this surfaced but doesn't fix: listing against an *unreachable* compat host takes ~7s, because discovery is a GET and the SDK retries the refused connection three times. The Dev UI blocks that long per poll. Probably its own issue.

Also adds `testapps/openai_sample/lib/compatible_providers.dart`, since the ticket notes the sample app only ever used plain `openAI(apiKey:)`. Flows for Groq, DeepSeek and xAI; backends without a key are skipped, so it starts up with none set. Each has a `<NAME>_BASE_URL` override for proxies and local servers (llama.cpp, Ollama, LM Studio).

```sh
GROQ_API_KEY=... genkit start -- dart run lib/compatible_providers.dart
```

### DIFF
--- a/packages/genkit_openai/test/fake_openai_server.dart
+++ b/packages/genkit_openai/test/fake_openai_server.dart
@@ -1,0 +1,323 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// A stand-in for an OpenAI-compatible host, ported from the JS plugin's
+/// `tests/fake_openai_server.ts`.
+///
+/// This binds a real socket rather than injecting a `MockClient` because the
+/// compat path is precisely the part a mock cannot vouch for: every other test
+/// in this package hands the plugin an `httpClient`, so the client the plugin
+/// builds for itself - the one real users get, and the only thing that turns
+/// `baseUrl` into an actual connection - is never exercised. Pointing the
+/// plugin at `server.baseUrl` with no `httpClient` proves the URL is dialed,
+/// the auth and custom headers survive the SDK's interceptor chain, and SSE is
+/// parsed off a genuinely chunked socket.
+library;
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+/// One request as the fake host saw it.
+class RecordedRequest {
+  /// HTTP method, e.g. `POST`.
+  final String method;
+
+  /// Request path, e.g. `/v1/chat/completions`.
+  final String path;
+
+  /// Request headers, lower-cased by `dart:io`.
+  final Map<String, String> headers;
+
+  /// The raw request body.
+  final String rawBody;
+
+  RecordedRequest({
+    required this.method,
+    required this.path,
+    required this.headers,
+    required this.rawBody,
+  });
+
+  /// The body decoded as JSON, or an empty map when there was no body.
+  Map<String, dynamic> get body => rawBody.isEmpty
+      ? const {}
+      : (jsonDecode(rawBody) as Map).cast<String, dynamic>();
+
+  @override
+  String toString() => '$method $path';
+}
+
+/// A canned reply for the fake host to serve.
+class FakeResponse {
+  /// HTTP status to write.
+  final int statusCode;
+
+  /// JSON-encodable body, for non-streaming replies.
+  final Object? body;
+
+  /// Extra response headers.
+  final Map<String, String> headers;
+
+  /// Whether to serve this reply as an SSE stream.
+  final bool stream;
+
+  /// SSE payloads, each written as one `data:` frame, followed by `[DONE]`.
+  final List<Object>? chunks;
+
+  /// Verbatim SSE frames, for wire shapes [chunks] cannot express. No `[DONE]`
+  /// frame is appended.
+  final List<String>? rawSse;
+
+  const FakeResponse({
+    this.statusCode = 200,
+    this.body,
+    this.headers = const {},
+    this.stream = false,
+    this.chunks,
+    this.rawSse,
+  });
+
+  /// A JSON reply with the given [body] and [statusCode].
+  const FakeResponse.json(Object body, {int statusCode = 200})
+    : this(body: body, statusCode: statusCode);
+
+  /// An OpenAI-style error reply, as a compatible host returns on failure.
+  factory FakeResponse.error(int statusCode, String message) {
+    return FakeResponse(
+      statusCode: statusCode,
+      body: {
+        'error': {'message': message, 'type': 'invalid_request_error'},
+      },
+    );
+  }
+
+  /// An SSE reply built from [chunks].
+  const FakeResponse.sse(List<Object> chunks)
+    : this(stream: true, chunks: chunks);
+}
+
+/// An OpenAI-compatible host backed by a real [HttpServer] on an ephemeral
+/// port.
+///
+/// Queued replies are served in order; once the queue drains, a path-aware
+/// default keeps tests that do not care about the response body short. When
+/// `expectedApiKey` is set, any request without a matching bearer token gets
+/// the same 401 shape OpenAI returns.
+class FakeOpenAIServer {
+  FakeOpenAIServer._(this._server, this._expectedApiKey) {
+    unawaited(_serve());
+  }
+
+  final HttpServer _server;
+  final String? _expectedApiKey;
+  final List<FakeResponse> _queued = [];
+
+  /// Every request the host received, in order.
+  final List<RecordedRequest> requests = [];
+
+  /// Anything thrown while serving a request.
+  ///
+  /// A handler that dies leaves the client staring at a closed socket, which
+  /// surfaces in the test as an opaque transport error a long way from the
+  /// cause. These are printed as they happen so the real reason is on screen.
+  final List<Object> handlerErrors = [];
+
+  /// Binds a fake host on an ephemeral loopback port.
+  ///
+  /// Pass [expectedApiKey] to have the host reject any other bearer token with
+  /// a 401, the way a real provider would.
+  static Future<FakeOpenAIServer> start({String? expectedApiKey}) async {
+    final server = await HttpServer.bind(InternetAddress.loopbackIPv4, 0);
+    return FakeOpenAIServer._(server, expectedApiKey);
+  }
+
+  /// The base URL to hand the plugin, including the `/v1` prefix that
+  /// OpenAI-compatible hosts conventionally serve under.
+  String get baseUrl => 'http://127.0.0.1:${_server.port}/v1';
+
+  /// Queues [response] to be served to the next request.
+  void enqueue(FakeResponse response) => _queued.add(response);
+
+  /// Stops the host, dropping any in-flight connections.
+  Future<void> stop() => _server.close(force: true);
+
+  /// The bodies of every chat-completions request received.
+  List<Map<String, dynamic>> get chatRequestBodies => [
+    for (final r in requests)
+      if (r.path.endsWith('/chat/completions')) r.body,
+  ];
+
+  Future<void> _serve() async {
+    await for (final request in _server) {
+      try {
+        await _handle(request);
+      } catch (e, stackTrace) {
+        // Usually the client hung up or the test ended, which is harmless. A
+        // bug in this harness lands here too, so make it visible rather than
+        // letting the test fail with an unexplained transport error.
+        handlerErrors.add(e);
+        stderr.writeln(
+          'FakeOpenAIServer failed to serve a request: $e\n$stackTrace',
+        );
+      }
+    }
+  }
+
+  Future<void> _handle(HttpRequest request) async {
+    final rawBody = await utf8.decoder.bind(request).join();
+    final headers = <String, String>{};
+    request.headers.forEach((name, values) => headers[name] = values.join(','));
+
+    requests.add(
+      RecordedRequest(
+        method: request.method,
+        path: request.uri.path,
+        headers: headers,
+        rawBody: rawBody,
+      ),
+    );
+
+    final expected = _expectedApiKey;
+    if (expected != null &&
+        request.headers.value('authorization') != 'Bearer $expected') {
+      await _writeJson(request.response, 401, {
+        'error': {
+          'message': 'Incorrect API key provided',
+          'type': 'invalid_request_error',
+          'param': null,
+          'code': 'invalid_api_key',
+        },
+      });
+      return;
+    }
+
+    final response = _queued.isNotEmpty
+        ? _queued.removeAt(0)
+        : _defaultFor(request.uri.path);
+
+    if (response.stream) {
+      await _writeSse(request.response, response);
+    } else {
+      await _writeJson(
+        request.response,
+        response.statusCode,
+        response.body,
+        extraHeaders: response.headers,
+      );
+    }
+  }
+
+  Future<void> _writeJson(
+    HttpResponse response,
+    int statusCode,
+    Object? body, {
+    Map<String, String> extraHeaders = const {},
+  }) async {
+    response.statusCode = statusCode;
+    response.headers.contentType = ContentType.json;
+    extraHeaders.forEach(response.headers.set);
+    response.write(jsonEncode(body));
+    await response.close();
+  }
+
+  Future<void> _writeSse(HttpResponse response, FakeResponse spec) async {
+    response.statusCode = spec.statusCode;
+    response.headers.set('content-type', 'text/event-stream');
+    response.headers.set('cache-control', 'no-cache');
+    spec.headers.forEach(response.headers.set);
+
+    final rawSse = spec.rawSse;
+    if (rawSse != null) {
+      for (final frame in rawSse) {
+        response.write(frame);
+        await response.flush();
+      }
+      await response.close();
+      return;
+    }
+
+    for (final chunk in spec.chunks ?? const []) {
+      response.write('data: ${jsonEncode(chunk)}\n\n');
+      await response.flush();
+    }
+    response.write('data: [DONE]\n\n');
+    await response.close();
+  }
+
+  FakeResponse _defaultFor(String path) {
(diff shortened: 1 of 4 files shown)


packages/genkit_openai/test/fake_openai_server.dart
```diff
@@ -0,0 +1,323 @@
...
+    'id': 'chatcmpl-fake',
+    'object': 'chat.completion',
+    'created': 0,
+    'model': model,
+    'choices': [
+      {
+        'index': 0,
+        'message': {
+          'role': 'assistant',
+          'content': content,
```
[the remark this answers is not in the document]
adesinah: This is valid Dart — it's the null-aware element syntax (`?expr` in a collection literal), stable since Dart 3.8. `sdk: ^3.10.0` here, and `dart analyze --fatal-infos` on this file reports no issues.

The suggested rewrite doesn't just work equally well — it fails CI. I applied it to check, and the analyzer asks for the original back:

```
info - test/fake_openai_server.dart:284:11 - Use the null-aware marker '?'
       rather than a null check via an 'if'. Try using '?'. - use_null_aware_elements
```

This repo analyzes with `--fatal-infos`, so `if (toolCalls != null)` is the version that breaks the build. The pattern is also already used in 31 files across the repo, including `packages/genkit/lib/src/types.g.dart`. No change made.

packages/genkit_openai/test/fake_openai_server.dart
```diff
@@ -0,0 +1,323 @@
...
+          'tool_calls': ?toolCalls,
+        },
+        'finish_reason': finishReason,
+      },
+    ],
```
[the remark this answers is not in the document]
adesinah: Same as the `tool_calls` thread: `?usage` is null-aware element syntax, valid on `sdk: ^3.10.0` and clean under `dart analyze --fatal-infos`.

Applying the suggestion makes the analyzer emit `use_null_aware_elements` at this exact line, and with `--fatal-infos` that fails CI. No change made.

packages/genkit_openai/test/fake_openai_server.dart
```diff
@@ -0,0 +1,323 @@
...
+}) {
+  return {
+    'id': 'chatcmpl-fake',
+    'object': 'chat.completion.chunk',
+    'created': 0,
+    'model': model,
+    'choices': [
+      {
+        'index': 0,
+        'delta': {'content': ?content},
```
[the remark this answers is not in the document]
adesinah: Same as the two threads above: `{'content': ?content}` is null-aware element syntax, valid and analyzer-clean.

Applying the suggestion produces `use_null_aware_elements` at line 307, which fails CI under `--fatal-infos`. No change made.

packages/genkit_openai/test/openai_plugin_compat_test.dart
```diff
@@ -0,0 +1,652 @@
...
+            baseUrl: server.baseUrl,
+            models: [llama()],
+          ),
+        ],
+      );
+      addTearDown(ai.shutdown);
+      ai.defineTool(
+        name: 'getWeather',
+        description: 'Get the weather for a location',
+        fn: (input, ctx) async => .response({'temperature': 72}),
```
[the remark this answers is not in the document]
adesinah: `.response(...)` is Dart 3.10 dot-shorthand for `ToolResult.response(...)` — the factory at `packages/genkit/lib/src/ai/tool.dart:68`. `ToolResult` is the declared return type of `fn`, so the target resolves and the leading dot is legal.

This suggestion is the one that would actually break the build. I applied it to confirm:

```
error - test/openai_plugin_compat_test.dart:211:35 - The returned type 'Map<String, int>'
        isn't returnable from a 'Future<ToolResult<_>>' function, as required by the
        closure's context. - return_of_invalid_type_from_closure
```

Returning the output directly stopped being correct when the tool API was redesigned around `ToolResult` in #350 — a tool now returns either `.response(output)` or an interrupt, which is why the wrapper exists. The identical line is already checked into `main` three times in `test/openai_plugin_wire_test.dart` (:81, :123, :163) and again in `test/integration_test.dart:106`. No change made.

adesinah:
Heads-up: #422 (curated model catalog) is stacked on this branch.

Worth knowing before this merges — #422 supersedes two things here:

- `knownChatModels` becomes a view derived from the catalog rather than a hand-kept list, so the `-pro` tiers drop out (Responses API only) and the `-chat-latest` snapshots come in (the catalog can express a chat model that takes no tools).
- The vision version-comparison heuristic is removed; `gpt-4.1` and the `gpt-5.x` line get media support from their catalog entries instead. The bug stays fixed, the tests move to asserting against the catalog.

Two assertions in `openai_plugin_startup_test.dart` change as a result. No action needed here — just so review of this PR isn't surprised by them later.

packages/genkit_openai/test/openai_plugin_compat_test.dart
```diff
@@ -0,0 +1,652 @@
...
+      );
+      addTearDown(ai.shutdown);
+
+      // Startup and listing survive a host that is not there at all. This
+      // takes ~7s: discovery is a GET, so the SDK retries the refused
+      // connection three times with exponential backoff before list() gives
+      // up and falls back to the catalog. That is the whole test - do not
+      // "speed it up" by dropping the listActions() call.
+      final actions = await ai.registry.listActions();
+      expect(actions, isNotEmpty);
```
cabljac: This is the whole 7s of the suite (everything else finishes in under 1s) and the assertion does not pin anything: with `baseUrl` set the catalog is withheld now (your `ddfdb8e`), so `listActions()` here returns `[util:generate]` only. It passes on the core action alone. The comment above saying it "falls back to the catalog" is stale for the same reason, and "do not speed it up" is defending an assertion that cannot fail.

Two options: assert the plugin-level state instead (`plugin.list()` is empty for a bare compat backend, or register a model and assert it is listed) and accept the 7s consciously, or drop the `listActions` half and keep generate-fails, which is the part that actually tests something. I would also assert the status rather than bare `isA<GenkitException>()`.

packages/genkit_openai/test/openai_plugin_compat_test.dart
```diff
@@ -0,0 +1,652 @@
...
+      final tools = (bodies.first['tools'] as List)
+          .cast<Map<String, dynamic>>();
+      final function = tools.single['function'] as Map<String, dynamic>;
+      expect(function['name'], 'getWeather');
+      // The second turn must carry the tool result the host asked for.
+      final followUpRoles = (bodies[1]['messages'] as List)
+          .cast<Map<String, dynamic>>()
+          .map((m) => m['role'])
+          .toList();
+      expect(followUpRoles, contains('tool'));
```
cabljac: Only checks a tool-role message exists. A plugin sending the wrong `tool_call_id` (the real failure mode on Groq/DeepSeek) passes this. Go pins it (`xai/xai_test.go:294-302`). Suggest finding the tool message and asserting `tool_call_id == 'call_1'` and content, plus `tool_calls[0].id == 'call_1'` on the re-sent assistant turn. The plugin does send them correctly today, `converters.dart:47-67`.

packages/genkit_openai/test/openai_plugin_compat_test.dart
```diff
@@ -0,0 +1,652 @@
...
+  });
+
+  group('compat host errors map to Genkit statuses', () {
+    const cases = <int, StatusCodes>{
+      400: StatusCodes.INVALID_ARGUMENT,
+      401: StatusCodes.UNAUTHENTICATED,
+      403: StatusCodes.PERMISSION_DENIED,
+      404: StatusCodes.NOT_FOUND,
+      500: StatusCodes.INTERNAL,
+    };
```
cabljac, line 608: nit: 503 -> `UNAVAILABLE` is missing. JS (`compat_oai_test.ts:2020-2028`) and Go (`errors_test.go:72`) both pin it, and it is free here: openai_dart only retries 5xx on idempotent methods (`retry_wrapper.dart:181-184`) and this is a POST. I checked, 12ms, one request.

packages/genkit_openai/test/openai_plugin_compat_test.dart
```diff
@@ -0,0 +1,652 @@
...
+        name: 'groq',
+        apiKey: 'groq-key',
+        baseUrl: compat.baseUrl,
+        customModels: [llama()],
+      );
+
+      final names = modelNames(await plugin.list());
+
+      expect(names.every((n) => n.startsWith('groq/')), isTrue);
+    });
```
cabljac: nit: same setup and assertion as `:474-486`, and both are subsumed by the exact-set test at `:547-569`. Drop this one.

packages/genkit_openai/test/fake_openai_server.dart
```diff
@@ -0,0 +1,323 @@
...
+/// `tests/fake_openai_server.ts`.
+///
+/// This binds a real socket rather than injecting a `MockClient` because the
+/// compat path is precisely the part a mock cannot vouch for: every other test
+/// in this package hands the plugin an `httpClient`, so the client the plugin
+/// builds for itself - the one real users get, and the only thing that turns
+/// `baseUrl` into an actual connection - is never exercised. Pointing the
+/// plugin at `server.baseUrl` with no `httpClient` proves the URL is dialed,
+/// the auth and custom headers survive the SDK's interceptor chain, and SSE is
+/// parsed off a genuinely chunked socket.
```
cabljac: nit: "the only thing that turns `baseUrl` into an actual connection" overstates it. openai_dart sends streaming through the injected client too (`streaming_resource.dart:171`), `openai_plugin_usage_wire_test.dart:134` already streams through a `MockClient`, and `startup_test.dart:364-381` already asserts the `baseUrl` host reaches the client. What this genuinely adds is the self-built `http.Client` and `client.close()` (`openai_plugin.dart:122,321`), real header serialisation, and real chunked SSE off a socket. Say that, so the next reader does not conclude the `MockClient` tests are worthless.

packages/genkit_openai/test/fake_openai_server.dart
```diff
@@ -0,0 +1,323 @@
...
+
+  /// Every request the host received, in order.
+  final List<RecordedRequest> requests = [];
+
+  /// Anything thrown while serving a request.
+  ///
+  /// A handler that dies leaves the client staring at a closed socket, which
+  /// surfaces in the test as an opaque transport error a long way from the
+  /// cause. These are printed as they happen so the real reason is on screen.
+  final List<Object> handlerErrors = [];
```
cabljac, line 130: nit: collected, never asserted, and already goes to stderr. Either add a `tearDown` that fails on non-empty (that is what would make it useful) or drop it. Same for `RecordedRequest.toString`, `FakeResponse.headers`, and the `model:` params on `chatCompletion`/`chatChunk`, none used by a test.

packages/genkit_openai/test/openai_plugin_compat_test.dart
```diff
@@ -0,0 +1,652 @@
...
+CustomModelDefinition llama() => CustomModelDefinition(
+  name: 'llama-3.3-70b-versatile',
+  info: ModelInfo(
+    label: 'Llama 3.3 70B',
+    supports: {'multiturn': true, 'tools': true, 'systemRole': true},
+  ),
+);
+
+void main() {
+  late FakeOpenAIServer server;
```
cabljac: nit: assigned in `setUp` in three groups and inside test bodies in `authentication`, while `multiple backends` and `errors` use locals. Return it from `startServer()` and use a local everywhere.

testapps/openai_sample/lib/compatible_providers.dart
```diff
@@ -0,0 +1,194 @@
...
+  final ai = Genkit(
+    plugins: [
+      for (final backend in configuredBackends)
+        openAI(
+          name: backend.name,
+          apiKey: backend.apiKey,
+          baseUrl: backend.baseUrl,
+          models: backend.models,
+          // Several compatible hosts read headers for attribution or routing;
+          // this is the knob OpenRouter's rankings use, for instance.
```
cabljac: nit: comment says this is OpenRouter's ranking knob but the backends are Groq, DeepSeek, xAI. Either drop the header or fix the comment.

cabljac (changes requested):
Suite is real: drives `ai.generate` through a loopback socket and asserts on bodies, headers, paths, status mapping and streamed output. One blocker, the unreachable-host test, plus a few cheap parity gaps inline.

The `list()` catalog-gating commit (`ddfdb8e`) is the fix for a #414 regression, I have asked for it to move down there so this PR stays test-only.

Spotted in passing, not this PR: `_handleStreaming` (`openai_plugin.dart:344-357`) only emits text deltas, so streaming tool calls never surface as partial `toolRequest` chunks. JS pins that (`compat_oai_test.ts:1590-1800`). Same on `main`. Do you agree? If so I will file it.

Also worth a note on #366: this fake server is a port of the JS one that issue names as the model, so it is the obvious seed for the shared harness. Generic bits (`FakeResponse`, request recording, SSE writer) would lift out, OpenAI builders stay here.

packages/genkit_openai/test/openai_plugin_compat_test.dart
adesinah: Done — asserts `tool_call_id == 'call_1'` on the tool message plus the content, and `tool_calls[0].id` on the re-sent assistant turn.

packages/genkit_openai/test/openai_plugin_compat_test.dart
adesinah: Rewrote it. Registers a custom model and asserts `plugin.list()` returns exactly that, so it pins plugin-level state instead of passing on `util:generate`. Stale comment gone, 7s now called out as accepted rather than defended.

Status is pinned too, but it's `INTERNAL`, not `UNAVAILABLE` — a refused socket never becomes an `ApiException` so there's no HTTP status to map. Left as-is with a note; changing it is behaviour, not test.

packages/genkit_openai/test/openai_plugin_compat_test.dart
adesinah: Added, and you're right that it's free — 12ms, one request.

packages/genkit_openai/test/openai_plugin_compat_test.dart
adesinah: Dropped.

packages/genkit_openai/test/fake_openai_server.dart
adesinah: Fair, that was overstated. Reworded to what it actually adds: the self-built client and its `close()`, header serialisation as the wire sees it, and SSE off genuinely chunked frames — with the MockClient tests credited for the rest.

packages/genkit_openai/test/fake_openai_server.dart
adesinah: Took the first option for `handlerErrors` — `stop()` now throws if it's non-empty, so a quietly broken host fails the test instead of passing.

Dropped the rest: `RecordedRequest.toString`, `FakeResponse.headers` and its plumbing, and the `model:` params.

packages/genkit_openai/test/openai_plugin_compat_test.dart
adesinah: Done — the file-level `late server` is gone, each setUp group declares its own, and the three in `authentication` are locals.

testapps/openai_sample/lib/compatible_providers.dart
adesinah: Fixed the comment — it now just says where a provider-specific header goes, without claiming OpenRouter.

packages/genkit_openai/test/openai_plugin_compat_test.dart
adesinah: Filed the status mapping as #424 and pointed the test comment at it — a refused socket should be UNAVAILABLE, but fixing it is a behaviour change and likely belongs in the shared error mapping, not this PR.

packages/genkit_openai/test/fake_openai_server.dart
```diff
@@ -0,0 +1,313 @@
...
+    for (final r in requests)
+      if (r.path.endsWith('/chat/completions')) r.body,
+  ];
+
+  Future<void> _serve() async {
+    await for (final request in _server) {
+      try {
+        await _handle(request);
+      } catch (e, stackTrace) {
+        // Usually the client hung up or the test ended, which is harmless. A
```
cabljac, line 170: nit: this says a caught handler error is harmless, but `stop()` at :151 throws on any of them. One of the two is wrong; I'd keep the throw and reword the comment.

cabljac (approved):
Approving. Three nits.

Body still says the `list()` catalog gating was "fixed in its own commit" here; that commit is now `a536e67` in #414 and this PR is test-only, so the paragraph is stale.

Keyless compat hosts need a dummy key: `_resolveClientConfig` throws `INVALID_ARGUMENT` with no key on every generate, so Ollama and LM Studio (named in the sample at `compatible_providers.dart:46`) cannot run as documented. Worth one line in the README's "Two things to know" section, and the sample filter at `:121` skips a backend with no key, so `<NAME>_BASE_URL` alone registers nothing.

Also still keen on your take on the streaming tool calls question from my first review, since the fake host is now the place to pin it.

pavelgj (approved):
The loopback fake host is the right call here; it covers the baseUrl path MockClient injection never touched. LGTM.

PR #422, opened 2026-09-08
feat(genkit_openai)!: curated model catalog and typed refs

### LINKED ISSUE #368: feat(genkit_openai): curated model catalog and typed refs (Go compat_oai pattern)
Dart has zero curated OpenAI models - `model(String)` plus name-matching heuristics (`utils.dart:57-63`), no per-model typed refs, no per-model `supports`. JS curates ~35 chat models with per-ref capabilities (`compat-oai/src/openai/gpt.ts:279-315`); Go curates ~50 incl. dated aliases (`compat_oai/openai/openai.go:99`) with a `dynamicModelOptions` fallback (`:239`) and typed `ModelRef` (`:390`).

Follow the Go pattern (compiled-language standard): string-keyed curated map + per-family default supports + dynamic fallback for unknown names + typed refs for autocomplete. Target the union of current JS/Go catalogs, not a copy - they have diverged (Go: gpt-4.1 family, gpt-5.6-*; JS: gpt-4.5).

Scope note: Go deliberately limits its catalog to Chat Completions models, excluding image/TTS/STT (`openai.go:94`) - a defensible 1.0 scope for Dart too; embedder refs land with #361.

adesinah:
> **Stack** (merge bottom-up)
> 1. #414 — offline startup
> 2. #419 — compat-path coverage
> 3. **#422 — curated catalog** ← this PR
>
> Reviews the diff against #419, not `main`.

Capabilities came from name matching over the model id. The rules could not
express a chat model that takes no tools, could not tell a structured-outputs
model from one predating it, and had to be re-guessed for every family OpenAI
ships.

Replaced with per-model data, following the Go `compat_oai` pattern: a catalog
of 34 chat models (union of the JS and Go catalogs), each with a capability
preset, its dated snapshots, and a lifecycle stage. Resolution is exact name →
dated-suffix alias → generic defaults, so an uncurated id still works and a
later snapshot still lands on its alias. `OpenAIModels` gives a typed
`ModelRef` per live model.

`knownChatModels` becomes a view over the catalog, so listing and capability
lookup cannot disagree.

### Changes to #414's behaviour

- **`-pro` tiers are not curated.** They are Responses API only and this plugin
  speaks Chat Completions, so a curated entry would offer a name that cannot be
  served. Note this only keeps them out of the *offline* catalog — `o1-pro`,
  `o3-pro`, `gpt-5-pro` and `gpt-5.2-pro` still come back from discovery,
  classify as chat, and register with the generic defaults, exactly as on
  `main` and in Go. Filtering them would mean a name heuristic, which is what
  this PR removes.
- **Retired models resolve but are never listed.** `gpt-4.5`, `chatgpt-4o-latest`,
  `o1-mini`, `o1-preview`, the `gpt-4` previews, `gpt-4-32k` and
  `gpt-4-vision-preview` are staged `deprecated`: `resolve()` still serves them
  by name with honest capabilities, but they never reach a model picker.
- **`gpt-5-chat-latest` now listed.** The catalog can say what the heuristics
  could not: chat and vision, no function calling.
- **Curated metadata behind a compat backend** keeps the capabilities but drops
  OpenAI's deployment details (label, stage, versions), and the catalog is not
  added to that host's listing.
- **An uncurated id now defaults to `multimodalSupports`.** Previously an
  unknown name went through the vision/tools heuristics; it now reports
  `media: true` and full chat capability unless `getModelType` says it is
  another modality. That is the Go `dynamicModelOptions` behaviour, and it is
  why a model released after this version works without a plugin release — but
  it is more optimistic than what it replaces.
- **Vision version-comparison heuristic removed** as superseded. The bug it
  fixed stays fixed; `gpt-4.1` and the `gpt-5.x` line get media support from
  their catalog entries.

**BREAKING CHANGE:** `defaultModelInfo` and `oSeriesModelInfo` are removed.
They returned a family-heuristic guess; `modelInfoFor` now answers from the
catalog for every name.

### Testing

```bash
cd packages/genkit_openai && dart test
```

152 tests pass; `dart analyze` and `dart format` clean.

Fixes #368


packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -228,12 +227,22 @@ class OpenAIPlugin extends GenkitPlugin {
...
 
+  /// Capability metadata for [modelName] on this plugin instance.
+  ///
+  /// For the same reason the catalog itself is withheld from a compat backend,
+  /// its metadata is too. A proxy that happens to serve a colliding name is
+  /// still not OpenAI: the label, the lifecycle stage and the list of served
+  /// snapshots are all claims about OpenAI's deployment, so those models take
+  /// the generic defaults instead.
+  ModelInfo _infoFor(String modelName) =>
+      baseUrl == null ? modelInfoFor(modelName) : dynamicModelInfo(modelName);
```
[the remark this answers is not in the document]
adesinah: Good catch — this was a regression I introduced. Verified: via a proxy, `gpt-3.5-turbo` and `o3-mini` were reporting `media=true constrained=true`, worse than the heuristics this PR replaces.

Fixed in 98e888c, taking your split: curated `supports` carries over (a gateway serving `gpt-3.5-turbo` is serving that model), while label, stage and versions — which describe OpenAI's own deployment — do not. Extracted as `compatModelInfo` in `known_models.dart` so the rule is testable on its own.

Note the test that should have caught this used `gpt-4o`, whose curated supports are identical to the generic fallback; it now uses a text-only model.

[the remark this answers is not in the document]
adesinah: Correction to my reply above, prompted by the same point on #414: I said the generic default "skips prompt-injected schema instructions by claiming native constrained output". That's wrong — nothing in `package:genkit` reads `supports`, so the metadata is descriptive and the generate path is unaffected either way.

The fix still stands, just for the narrower reason: what's wrong is what the Dev UI picker and `listActions` report about a model. Comment in `known_models.dart` corrected to say so.

packages/genkit_openai/lib/src/known_models.dart
```diff
@@ -12,49 +12,548 @@
...
+      'gpt-3.5-turbo-0125',
+      'gpt-3.5-turbo-1106',
+      'gpt-3.5-turbo-0613',
+      'gpt-3.5-turbo-16k',
+      'gpt-3.5-turbo-16k-0613',
+    ],
+    // Azure drops the dot from the model name. Without these, `gpt-35-turbo`
+    // is uncurated and takes the multimodal defaults, which would have Genkit
+    // send images to a text-only model.
+    aliases: ['gpt-35-turbo', 'gpt-35-turbo-16k'],
```
cabljac: nit: covers `gpt-35-turbo` and `-16k`; `gpt-35-turbo-0125`, `-1106` and `-16k-0613` fall through to multimodal because `_datedSuffixPattern` (`:478`) only strips `-YYYY-MM-DD`. Azure deployment names carry those suffixes.

packages/genkit_openai/lib/genkit_openai.dart
```diff
@@ -139,3 +155,166 @@ class OpenAICompatPluginHandle {
...
+/// Typed [ModelRef]s for the OpenAI models curated by the `openai` plugin.
+///
+/// Each entry is equivalent to `openAI.model('<name>')`, which remains the
+/// escape hatch for models not listed here and for plugin instances registered
+/// under a custom namespace.
+///
+/// Only models OpenAI still serves get a ref. A curated model whose stage is
+/// [OpenAIModelStage.deprecated] is reachable by name — see
+/// [KnownOpenAIModel] — but is not offered for autocomplete.
+abstract final class OpenAIModels {
```
cabljac, line 173: nit: 29 statics for 29 non-deprecated enum entries today, and the test touches four. Nothing fails when an entry is added without a ref. A `static final List<ModelRef> all` asserted equal to the enum, or refs generated from it, would pin the parity.

packages/genkit_openai/lib/src/known_models.dart
```diff
@@ -12,49 +12,548 @@
...
+
+/// Curated capability metadata for known OpenAI chat models, keyed by bare
+/// model name (no plugin prefix).
 ///
-/// The plugin normally discovers models from `GET /models`, but that needs both
-/// connectivity and a valid key. This list is what gets listed when discovery
-/// is unavailable, and it is merged with the discovered set when it is.
+/// Derived from [KnownOpenAIModel]; other names still resolve, they just take
+/// [dynamicModelInfo] instead of a curated entry.
+final knownOpenAIModels = <String, ModelInfo>{
```
cabljac: nit: `knownOpenAIModels` and `knownChatModels` are exported mutable `final` collections while `info.supports` is `Map.unmodifiable`. `.add()` from user code corrupts global state; wrap both.

packages/genkit_openai/lib/src/known_models.dart
```diff
@@ -12,49 +12,548 @@
...
+const multimodalLegacySupports = <String, dynamic>{
+  ..._chatCore,
+  'tools': true,
+  'toolChoice': true,
+  'media': true,
+  'output': ['text', 'json'],
+};
+
+/// [multimodalLegacySupports] without image input.
+const textOnlyLegacySupports = <String, dynamic>{
```
cabljac, line 72: nit: `gpt-4` and `gpt-4-32k` advertise `output: ['text', 'json']` here and in Go, but JS `GPT_4_MODEL_INFO` says `['text']`, and gpt-4-0613 400s on `response_format`. Worth checking which is right before it ships as curated truth.

cabljac (changes requested):
Direction is right and the catalog itself reads well. Three things before this lands, then a few nits.

Uncurated ids now default to full multimodal (`dynamicModelInfo`, `known_models.dart:517`): `deepseek-chat`, a Groq llama name, `ft:gpt-4.1:org::id` all go from `media: false` on #414 to `media: true, constrained: true`, and uncurated o-series (`o5`, `o3-pro`) flip `systemRole` to true where curated `o3` is false. This matches Go's `dynamicModelOptions`, so I'm fine with the default, but it is not in the "Changes to #414's behaviour" list and a compat user with a `CustomModelDefinition` and no `info` now advertises vision on a text model. List it, or keep `media: false` for names that don't look like OpenAI's.

`gpt-4o-search-preview` and `-mini-search-preview` regress: `getModelType` says `search` (`utils.dart:87`), which `dynamicModelInfo:513` maps to `nonChatSupports`, so they go from usable chat on #414 (`multiturn`, `systemRole` true) to `multiturn: false`. OpenAI lists them as Chat Completions models with image input and structured outputs. Either curate them or stop classifying `search` as non-chat.

#414's vision fix is unpinned here. The `supportsVision` assertions for `gpt-4.1*` and `gpt-5*` are deleted and the startup catalog test asserts `isNotNull` where it asserted `isTrue`. I flipped `gpt41` and then `gpt5` to `textOnlyLegacySupports` and 155/155 still pass. One assertion per family would do.

README `:279-280` says the curated catalog "is not applied" for a compat `baseUrl`, but since `2c0fc86` `compatModelInfo` applies supports and drops only label, stage and versions. #419's README says the catalog is withheld too, so both paragraphs need the same rewrite. `:259` is a broken sentence.

The `-pro` exclusion only holds offline (`openai_plugin.dart:232`). Discovery is filtered by `getModelType` (`:205`), and `o1-pro`, `o3-pro`, `gpt-5-pro` classify as chat, so with a key `/models` lists them with tools and media. The body says they are dropped; either filter them in discovery too or say the exclusion is catalog-only.

Do you agree on the search-preview one? I'd like to confirm the classification before proposing where the fix goes.

packages/genkit_openai/lib/src/known_models.dart
adesinah: Right, my fix only covered the two names I'd listed. Confirmed `gpt-35-turbo-0125`, `-1106`, `-0613` and `-16k-0613` all came back `media=true`.

Dropped the enumerated aliases for a normalisation step instead: `gpt-(\d)(\d)-` → `gpt-$1.$2-`, tried before and after the dated-suffix strip. That maps the whole family, and any future Azure-dotted name, in one rule. Test covers all six.

packages/genkit_openai/lib/src/known_models.dart
adesinah: Checked, and JS is right — the `gpt-4` and `gpt-4-32k` aliases resolve to `-0613`, which predates `response_format` entirely. Both are on a new `textOnlyNoJsonSupports` preset with `output: ['text']` now.

Left the `-1106`/`-0125` previews and `gpt-4-turbo` on json, since that's the generation where it arrived. Test pins both sides.

packages/genkit_openai/lib/genkit_openai.dart
adesinah: Added `OpenAIModels.all` and a test asserting its names equal `knownChatModels`, so adding a catalog entry without a ref now fails.

Kept the statics hand-written rather than generating them — generated refs would lose the autocomplete, which is the point of having them.

packages/genkit_openai/lib/src/known_models.dart
adesinah: Fixed, both are `Map.unmodifiable` / `List.unmodifiable` now, with a test.

cabljac:
Staging looks right in design, and the `deprecated` gate at `known_models.dart:471` does the right thing. Four entries are staged wrong though, and they all get registered:

| id | stage | GET /v1/models/{id} |
|---|---|---|
| `chatgpt-4o-latest` | stable | 404 |
| `gpt-4-0125-preview` | legacy | 404 |
| `gpt-4-1106-preview` | legacy | 404 |
| `gpt-4-32k` | legacy | 404 |

All four pass the `!= deprecated` check, so they end up in `knownChatModels` and give the picker names that 404, which is the thing the comment at `:467` says the gate is there to prevent. `chatgpt-4o-latest` is the odd one since `stable` claims no announced shutdown. Moving all four to `deprecated` looks like the fix. Checked against one account's key, so worth confirming none of these are just tier-gated for me.

The rest of the catalog checks out: 33 top-level entries, the other 29 resolve, and the four already marked `deprecated` are correctly dead.

`test/openai_plugin_utils_test.dart:107`: the vision fix from #414 isn't pinned. I flipped `gpt41`/`gpt5` to text-only and the suite still passed 160/160.

`openai_plugin.dart:205`: `-pro` tiers still come back from discovery (o1-pro, o3-pro, gpt-5-pro, gpt-5.2-pro), though the body says they were dropped.

Also worth a line in the PR body: uncurated ids now default to `multimodalSupports`, so anything unknown gets `media: true` where it previously went through the heuristic.

On the exports at `genkit_openai.dart:26-49`: 14 have no caller outside lib/test. `compatModelInfo` has a lib caller, and the presets plus `knownOpenAIModelFor` seem reasonable as public API. The rest I'd keep in `src/` rather than commit to them in a breaking release.

nit: README:279-280 still says the catalog isn't applied when `baseUrl` is set, and :259 has a sentence that runs together ("This is the catalog listing falls back to").

adesinah:
All six done, thanks — the staging table was the useful one.

**The four entries.** Moved all to `deprecated`, so 25 register instead of 29. `chatgpt-4o-latest` I'd like a second key on before we trust it: `stable` was clearly wrong either way, but tier-gating and retirement look identical from one account, and a false positive means it drops out of every picker. It still resolves by name, so the downside is small if we're wrong.

**Vision fix unpinned.** Confirmed — I reproduced your mutation, 160/160. Added a test asserting the current multimodal families report `media: true`, plus the inverse for `gpt-4`/`gpt-3.5-turbo`/`o3-mini` so it can't be satisfied by making everything multimodal. Flipping `gpt41` now fails it.

**`-pro` tiers.** You're right, the body overclaimed. They're only absent from the curated catalog; discovery still surfaces and registers them, same as `main` and Go. Filtering would mean a name heuristic, which is the thing this PR removes, so I've corrected the body rather than the behaviour.

**Uncurated default.** Added to the body, called out as more optimistic than the heuristic it replaces.

**Exports.** Cut to your line: presets, catalog and lookups stay; `dynamicModelInfo`, `compatModelInfo`, `openAIModelAlias` and `openAIModelSpelling` moved back behind `src/` with a comment on why — they're resolution policy, not vocabulary. Tests import them from `src/`.

**README.** Both fixed. The `baseUrl` sentence was worse than stale — it predated the compat fix and said the opposite of what the code does now.

161 tests, analyze clean, in `6a32784`.

cabljac (approved):
lgtm

pavelgj (approved):
Catalog-over-heuristics is the right direction and the entries read well. Verified analyze + 161 tests green on the top of the stack. LGTM.

PR #425, opened 2026-09-10
fix(google_genai): map Gemini finish reasons to Genkit vocabulary

pavelgj:
## Summary

`fromGeminiCandidate` used to build the finish reason by lowercasing the raw Gemini enum string (`FinishReason(candidate.finishReason?.toLowerCase() ?? 'unspecified')`). That leaked non-canonical values into Genkit's vocabulary: `max_tokens`, `recitation`, `unspecified`, and any future enum value all flowed straight through instead of mapping onto the reasons Genkit actually understands (`stop`, `length`, `blocked`, `other`, `unknown`, ...).

This PR replaces that with an explicit `_toFinishReason` switch that mirrors the Go plugin's `translateCandidate` and the JS plugin's `fromGeminiFinishReason`:

- `STOP` -> `stop`
- `MAX_TOKENS` -> `length`
- `SAFETY` / `RECITATION` / `LANGUAGE` / `BLOCKLIST` / `PROHIBITED_CONTENT` / `SPII` / `IMAGE_SAFETY` / `IMAGE_PROHIBITED_CONTENT` / `IMAGE_RECITATION` -> `blocked`
- `MALFORMED_FUNCTION_CALL` / `UNEXPECTED_TOOL_CALL` / `TOO_MANY_TOOL_CALLS` / `NO_IMAGE` / `IMAGE_OTHER` / `MALFORMED_RESPONSE` / `MISSING_THOUGHT_SIGNATURE` / `OTHER` -> `other`
- absent / empty / `FINISH_REASON_UNSPECIFIED` / any unrecognized value -> `unknown`

Mapping absent and unspecified reasons to `unknown` (a result-carrying reason) rather than leaking `unspecified` means a truncated turn that names no reason still surfaces whatever text it produced. `MISSING_THOUGHT_SIGNATURE` (returned by Gemini 3 when thought signatures are missing from the request) is treated as a non-fatal `other`, matching the Go plugin.

## Also

- Handle content-less candidates: safety blocks and other early terminations arrive without `content`, so default the role to `model` instead of dereferencing a null `content` (previously `candidate.content!.role!` would throw).

## Tests

Adds `finish_reason_test.dart` covering the full mapping table, the absent-reason case, and the content-less candidate path.

### DIFF
--- a/packages/genkit_google_genai/lib/src/common_plugin.dart
+++ b/packages/genkit_google_genai/lib/src/common_plugin.dart
@@ -258,14 +258,53 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
 }
 
 (Message, FinishReason) fromGeminiCandidate(gcl.Candidate candidate) {
-  final finishReason = FinishReason(
-    candidate.finishReason?.toLowerCase() ?? 'unspecified',
-  );
   final message = Message(
-    role: Role(candidate.content!.role!),
+    role: Role(candidate.content?.role ?? 'model'),
     content: candidate.content?.parts?.map(fromGeminiPart).toList() ?? [],
   );
-  return (message, finishReason);
+  return (message, _toFinishReason(candidate.finishReason));
+}
+
+/// Maps a Gemini finish reason onto Genkit's vocabulary. An absent or
+/// `FINISH_REASON_UNSPECIFIED` reason becomes [FinishReason.unknown] (a
+/// result-carrying reason), never the non-canonical `unspecified`: a truncated
+/// turn that names no reason still carries whatever text it produced.
+/// Unrecognized future reasons also fall to `unknown` rather than leaking the
+/// raw enum string.
+FinishReason _toFinishReason(String? raw) {
+  switch (raw) {
+    case null:
+    case '':
+    case 'FINISH_REASON_UNSPECIFIED':
+      return FinishReason.unknown;
+    case 'STOP':
+      return FinishReason.stop;
+    case 'MAX_TOKENS':
+      return FinishReason.length;
+    case 'SAFETY':
+    case 'RECITATION':
+    case 'LANGUAGE':
+    case 'BLOCKLIST':
+    case 'PROHIBITED_CONTENT':
+    case 'SPII':
+    case 'IMAGE_SAFETY':
+    case 'IMAGE_PROHIBITED_CONTENT':
+    case 'IMAGE_RECITATION':
+      return FinishReason.blocked;
+    case 'MALFORMED_FUNCTION_CALL':
+    case 'UNEXPECTED_TOOL_CALL':
+    case 'TOO_MANY_TOOL_CALLS':
+    case 'NO_IMAGE':
+    case 'IMAGE_OTHER':
+    case 'MALFORMED_RESPONSE':
+    // Gemini 3 returns this when thought signatures are missing from the
+    // request; treat it as a non-fatal "other" like the Go plugin.
+    case 'MISSING_THOUGHT_SIGNATURE':
+    case 'OTHER':
+      return FinishReason.other;
+    default:
+      return FinishReason.unknown;
+  }
 }
 
 @visibleForTesting
--- a/packages/genkit_google_genai/test/finish_reason_test.dart
+++ b/packages/genkit_google_genai/test/finish_reason_test.dart
@@ -21,7 +21,10 @@ import 'package:test/test.dart';
 gcl.Candidate _candidate(String? finishReason) => gcl.Candidate(
   index: 0,
   finishReason: finishReason,
-  content: gcl.Content(role: 'model', parts: [gcl.Part(text: 'hi')]),
+  content: gcl.Content(
+    role: 'model',
+    parts: [gcl.Part(text: 'hi')],
+  ),
 );
 
 void main() {
@@ -55,13 +58,16 @@ void main() {
       expect(reason.value, isNot('unspecified'));
     });
 
-    test('content-less candidate does not throw and defaults role to model', () {
-      final (message, reason) = fromGeminiCandidate(
-        gcl.Candidate(index: 0, finishReason: 'STOP'),
-      );
-      expect(message.role, Role.model);
-      expect(message.content, isEmpty);
-      expect(reason.value, FinishReason.stop.value);
-    });
+    test(
+      'content-less candidate does not throw and defaults role to model',
+      () {
+        final (message, reason) = fromGeminiCandidate(
+          gcl.Candidate(index: 0, finishReason: 'STOP'),
+        );
+        expect(message.role, Role.model);
+        expect(message.content, isEmpty);
+        expect(reason.value, FinishReason.stop.value);
+      },
+    );
   });
 }
--- a/packages/genkit_google_genai/test/finish_reason_test.dart
+++ b/packages/genkit_google_genai/test/finish_reason_test.dart
@@ -1,0 +1,67 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:genkit/plugin.dart';
+import 'package:genkit_google_genai/src/common_plugin.dart';
+import 'package:genkit_google_genai/src/generated/generativelanguage.dart'
+    as gcl;
+import 'package:test/test.dart';
+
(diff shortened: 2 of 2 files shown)


packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -258,14 +258,53 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
...
+}
+
+/// Maps a Gemini finish reason onto Genkit's vocabulary. An absent or
+/// `FINISH_REASON_UNSPECIFIED` reason becomes [FinishReason.unknown] (a
+/// result-carrying reason), never the non-canonical `unspecified`: a truncated
+/// turn that names no reason still carries whatever text it produced.
+/// Unrecognized future reasons also fall to `unknown` rather than leaking the
+/// raw enum string.
+FinishReason _toFinishReason(String? raw) {
+  switch (raw) {
```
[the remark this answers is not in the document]
pavelgj: The finish reason is a protobuf enum and always arrives as SCREAMING_SNAKE_CASE; the Go (translateCandidate) and JS (fromGeminiFinishReason) plugins both match the uppercase literals exactly without normalizing case. The previous .toLowerCase() here was not handling lowercase input, it was just converting to Genkit's lowercase vocabulary. Keeping the exact-match switch to stay consistent with the other plugins, so declining the normalization.

packages/genkit_google_genai/test/finish_reason_test.dart
```diff
@@ -0,0 +1,67 @@
...
+      'SOME_FUTURE_REASON': FinishReason.unknown,
+    };
+
+    cases.forEach((raw, expected) {
+      test('$raw -> ${expected.value}', () {
+        final (_, reason) = fromGeminiCandidate(_candidate(raw));
+        expect(reason.value, expected.value);
+      });
+    });
+
```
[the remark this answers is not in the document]
pavelgj: Declining for the same reason as the normalization suggestion above: the API always returns uppercase enum values, so a lowercase input path is not something we need to test. Go and JS match the uppercase literals exactly without a case-insensitive path.

pavelgj:
@cabljac PTAL

packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -258,14 +258,53 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
...
+    case 'SAFETY':
+    case 'RECITATION':
+    case 'LANGUAGE':
+    case 'BLOCKLIST':
+    case 'PROHIBITED_CONTENT':
+    case 'SPII':
+    case 'IMAGE_SAFETY':
+    case 'IMAGE_PROHIBITED_CONTENT':
+    case 'IMAGE_RECITATION':
+      return FinishReason.blocked;
```
cabljac, line 293: Three I spotted that look like they belong here:

- `ESCALATION`: "Request was filtered by an escalation rule."
- `PUP_LIMITED_DISABLED`: account limited or disabled for Prohibited Use Policy violations.
- `MODEL_ARMOR`: "The model response was blocked by Model Armor."

First two are in live generativelanguage v1beta, third is Vertex aiplatform v1, which lands here via genkit_vertexai. All three get `unknown` today, so the abnormal guard doesn't fire.

packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -258,14 +258,53 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
...
+    case 'TOO_MANY_TOOL_CALLS':
+    case 'NO_IMAGE':
+    case 'IMAGE_OTHER':
+    case 'MALFORMED_RESPONSE':
+    // Gemini 3 returns this when thought signatures are missing from the
+    // request; treat it as a non-fatal "other" like the Go plugin.
+    case 'MISSING_THOUGHT_SIGNATURE':
+    case 'OTHER':
+      return FinishReason.other;
+    default:
```
cabljac, line 305: Broader version: any blocked-class reason they add later hits this arm and gets `unknown`, so parsing runs on an empty message. Go does the same, so matching it is fair enough.

Needs a decision either way: `other` here instead of `unknown`? Slightly wrong label, but abnormal, so parsing gets skipped.

packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -258,14 +258,53 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
...
-  final finishReason = FinishReason(
-    candidate.finishReason?.toLowerCase() ?? 'unspecified',
-  );
   final message = Message(
-    role: Role(candidate.content!.role!),
+    role: Role(candidate.content?.role ?? 'model'),
     content: candidate.content?.parts?.map(fromGeminiPart).toList() ?? [],
   );
-  return (message, finishReason);
+  return (message, _toFinishReason(candidate.finishReason));
```
cabljac, line 265: `candidate.finishMessage` is dropped. Go sets it (gemini.go:560), JS sets it (converters.ts:740), and `ModelResponse.finishMessage` exists (types.g.dart:2326). It's the only diagnostic left on a block now that these return cleanly instead of throwing. Streaming too: aggregation.dart:57 declares `finalFinishMessage`, `merge` never sets it.

packages/genkit_google_genai/lib/src/common_plugin.dart
```diff
@@ -258,14 +258,53 @@ abstract class CommonGoogleGenPlugin extends GenkitPlugin {
 }
 
 (Message, FinishReason) fromGeminiCandidate(gcl.Candidate candidate) {
-  final finishReason = FinishReason(
-    candidate.finishReason?.toLowerCase() ?? 'unspecified',
-  );
   final message = Message(
-    role: Role(candidate.content!.role!),
+    role: Role(candidate.content?.role ?? 'model'),
```
cabljac, line 262: Worth a line in the body: a candidate with no `content` and no `finishReason` now comes back as an empty `model` message with `unknown`. Go calls that malformed and returns INTERNAL "no valid candidates were found in the generate response" (gemini.go:569). JS does what this PR does.

packages/genkit_google_genai/test/finish_reason_test.dart
```diff
@@ -0,0 +1,73 @@
...
+  content: gcl.Content(
+    role: 'model',
+    parts: [gcl.Part(text: 'hi')],
+  ),
+);
+
+void main() {
+  group('fromGeminiCandidate finish reason mapping', () {
+    // Mirrors the Go plugin's translateCandidate switch.
+    final cases = <String, FinishReason>{
```
cabljac, line 33: Covers 12 of the 20 values in the checked-in discovery doc. Missing `LANGUAGE`, `BLOCKLIST`, `SPII`, `IMAGE_PROHIBITED_CONTENT`, `IMAGE_RECITATION`, `TOO_MANY_TOOL_CALLS`, `NO_IMAGE`, `IMAGE_OTHER`, `MALFORMED_RESPONSE`, and `''`. Body says full mapping table.

cabljac (commented):
Mapping matches Go (gemini.go:527-557) and JS (converters.ts:389-411) on everything it covers.

One gap I spotted and wanted to raise: three blocked-class reasons hit `default` and come back `unknown`. `unknown` isn't in core's `isAbnormal` set (generate.dart:268), so the parser still runs on the empty content (generate.dart:656):

    model failed to generate output matching expected schema: FormatException: No JSON object or array found

Reproduced on b6975ac via `generate(outputFormat: 'json')`.

Easy to miss from in here: generativelanguage-discovery.json has 20 finish reasons, live v1beta has 22. `ESCALATION` and `PUP_LIMITED_DISABLED` are only in the live one. Refreshing the snapshot would stop this reopening next time they add one.

Rest is minor, inline.

cabljac (approved):
No blockers here, left comments inline

packages/genkit_google_genai/lib/src/common_plugin.dart
pavelgj: Added `ESCALATION`, `PUP_LIMITED_DISABLED` and `MODEL_ARMOR` to `blocked`.

packages/genkit_google_genai/lib/src/common_plugin.dart
pavelgj: Unrecognized reasons now map to `other`, and absent or `UNSPECIFIED` reasons stay `unknown`. The enum lists across the discovery doc, protos and SDKs don't match, so the fallback has to be safe. This deliberately differs from Go/JS; there's a comment in the code.

packages/genkit_google_genai/lib/src/common_plugin.dart
pavelgj: Fixed on both paths. Aggregation now sets `finalFinishMessage` and passes it through.

packages/genkit_google_genai/lib/src/common_plugin.dart
pavelgj: Added a note to the body.

packages/genkit_google_genai/test/finish_reason_test.dart
pavelgj: Filled in the table, including `''`, plus a test that an unrecognized reason maps to `other`, and a `generate` JSON-output test for your reproduction.

Issue #327, opened 2026-07-13
feat(core): embedder capability metadata (dimensions, input modalities)


cabljac:
Part of #303. Blocks the metadata half of the gemini-embedding-2 work (see the follow-up googleai issue).

## Problem

Genkit core has no home for embedder capability metadata. `embedderMetadata()` / `Embedder` (`packages/genkit/lib/src/ai/embedder.dart`) only carry a label and custom options schema. There is no equivalent of:

- Go `ai.EmbedderOptions`: `Dimensions int`, `Supports.Input []string`
- JS `EmbedderInfo`: `dimensions`, `supports.input`

So plugin embedder registrations cannot advertise output dimensions or accepted input modalities (text / image / video), and the Dev UI cannot display them. This was flagged in #261 review and in the #302 discussion (closed by the model-side catalog work in #320/#323, which did not touch embedders).

## Concrete need

Go's `embedderConfig` values that currently have nowhere to go in Dart:

- `gemini-embedding-2`: dimensions 3072, input [text, image, video]
- `multimodalembedding`: input [text, image, video]
- `text-embedding-*` / `text-multilingual-embedding-002`: dimensions 768, input [text]

## Scope

1. Add embedder info to core (shape aligned with Go `EmbedderOptions` / JS `EmbedderInfo`): `dimensions`, `supports.input`, plus label/stage consistent with `ModelInfo`.
2. Thread it through `Embedder`, `embedderMetadata()`, and reflection metadata so the Dev UI surfaces it.
3. Migrate existing embedder registrations (googleai, vertexai, openai) to populate it where values are known.

API design should get review before implementation; this is public core surface.

cabljac:
Design decisions to settle before implementation, from comparing core, JS, Go, and the Dev UI:

**1. Type source: hand-written vs codegen.** Core types come from `genkit-tools/genkit-schema.json` via `_schema_generator`, but that schema has `ModelInfo` only - JS never exports `EmbedderInfoSchema` into it. So either we hand-write the type in core, or an upstream JS change exports it to the canonical schema and we regen. Hand-written unblocks this issue now; codegen keeps the cross-SDK contract but couples us to an upstream release.

[...]

PR #429, opened 2026-09-11
feat(genkit_openai): ship embedders with a curated per-model catalog

### LINKED ISSUE #361: feat(genkit_openai): embedders
Embedding models are discovered then filtered out (`openai_plugin.dart:89`, `utils.dart:204-206`). JS and Go both ship embedders with per-model dimensions. Partially blocked on core embedder metadata (#327) for dimensions metadata, but the embed path itself is not.

adesinah:
Embedding models were discovered and then thrown away. This ships them: `resolve()` builds an embedder for any name, `list()` routes embedding ids to embedder metadata instead of dropping them, and `openAI.embedder()` / `OpenAIEmbedders` give you refs.

The catalog is a `KnownOpenAIEmbedder` enum shaped like `KnownOpenAIModel` — id, label, vector length, and whether the model will return a shorter one. Three entries; checked `/v1/models` with a live key and that is still the whole lineup.

Dimensions metadata is the half #327 blocks. Until core has an `EmbedderInfo`, the values ride in the `model` metadata map that `embedderMetadata()` and `Embedder` already merge, under JS `EmbedderInfo`'s keys. That shape will change when #327 lands, which is why `embedderInfoFor` stayed in `src/` rather than becoming API.

Decisions worth naming: documents flatten to text and media parts are dropped (a document with no text at all is an error naming its index); more than 2048 documents split across requests sequentially rather than 400ing; a `dimensions` a curated model can't honour is rejected before the request goes out, but only against OpenAI's own host — a compat backend serving that name may not share the limit, so it answers for itself.

`getModelType` now matches `embed` rather than `embedding`, which is what makes `nomic-embed-text` and `mxbai-embed-large` land as embedders. A name carrying neither (`bge-m3`, `all-minilm`) is still missed, and nothing short of declaring it will catch that — a `CustomEmbedderDefinition` to match `CustomModelDefinition` is the follow-up. `resolve()` serves any of them by name meanwhile, so nothing is unreachable.

Tested live: vector lengths and `dimensions: 256` both hold against the API.

Closes #361.

### DIFF
--- a/packages/genkit_openai/lib/genkit_openai.dart
+++ b/packages/genkit_openai/lib/genkit_openai.dart
@@ -18,11 +18,25 @@ import 'package:genkit/plugin.dart';
 import 'package:http/http.dart' as http;
 
 import 'src/chat.dart' as chat;
+import 'src/embed.dart' as embed;
+import 'src/known_embedders.dart';
 import 'src/known_models.dart';
 import 'src/openai_plugin.dart';
 
 export 'src/chat.dart' show OpenAIChatOptions, OpenAIOptions;
 export 'src/converters.dart' show GenkitConverter;
+export 'src/embed.dart' show OpenAIEmbedderOptions;
+// The embedder catalog is public for the same reason the model catalog is.
+// `embedderInfoFor` and its compat variant are not: until core grows an
+// `EmbedderInfo` (#327) they hand back a raw map whose shape is expected to
+// change, and freezing that as API now would make the migration a breaking
+// one.
+export 'src/known_embedders.dart'
+    show
+        KnownOpenAIEmbedder,
+        knownEmbedderModels,
+        knownOpenAIEmbedderFor,
+        knownOpenAIEmbedders;
 // The catalog and the capability vocabulary are public: describing a model
 // the plugin does not know is a supported thing to do, and a caller doing it
 // should reach for the same presets the curated entries use.
@@ -157,6 +171,28 @@ class OpenAICompatPluginHandle {
     return modelRef(
       '$namespace/$name',
       customOptions: chat.chatModelOptionsSchema(),
+    );
+  }
+
+  /// Reference to an embedding model.
+  ///
+  /// Takes [namespace] for the same reason [model] does: it is the prefix the
+  /// embedder is looked up under (e.g. `openai/text-embedding-3-small`), and a
+  /// plugin registered with a custom name needs it passed.
+  ///
+  /// ```dart
+  /// final vectors = await ai.embed(
+  ///   embedder: openAI.embedder('text-embedding-3-small'),
+  ///   document: DocumentData(content: [TextPart(text: 'hello')]),
+  /// );
+  /// ```
+  EmbedderRef<embed.OpenAIEmbedderOptions> embedder(
+    String name, {
+    String namespace = defaultOpenAINamespace,
+  }) {
+    return embedderRef(
+      '$namespace/$name',
+      customOptions: embed.embedderOptionsSchema(),
     );
   }
 }
@@ -334,5 +370,36 @@ abstract final class OpenAIModels {
     gpt4Turbo,
     gpt4,
     gpt35Turbo,
+  ];
+}
+
+/// Typed [EmbedderRef]s for the OpenAI embedders curated by the `openai`
+/// plugin.
+///
+/// Each entry is equivalent to `openAI.embedder('<name>')`, which remains the
+/// escape hatch for embedders not listed here and for plugin instances
+/// registered under a custom namespace.
+abstract final class OpenAIEmbedders {
+  /// OpenAI text-embedding-3-small, 1536 dimensions.
+  static final EmbedderRef<embed.OpenAIEmbedderOptions> textEmbedding3Small =
+      openAI.embedder(KnownOpenAIEmbedder.textEmbedding3Small.id);
+
+  /// OpenAI text-embedding-3-large, 3072 dimensions.
+  static final EmbedderRef<embed.OpenAIEmbedderOptions> textEmbedding3Large =
+      openAI.embedder(KnownOpenAIEmbedder.textEmbedding3Large.id);
+
+  /// OpenAI text-embedding-ada-002, 1536 dimensions and no `dimensions`
+  /// option.
+  static final EmbedderRef<embed.OpenAIEmbedderOptions> textEmbeddingAda002 =
+      openAI.embedder(KnownOpenAIEmbedder.textEmbeddingAda002.id);
+
+  /// Every ref above, in catalog order.
+  ///
+  /// Exists so the statics cannot silently fall behind [KnownOpenAIEmbedder],
+  /// the same way `OpenAIModels.all` guards the model refs.
+  static final List<EmbedderRef<embed.OpenAIEmbedderOptions>> all = [
+    textEmbedding3Small,
+    textEmbedding3Large,
+    textEmbeddingAda002,
   ];
 }
--- a/packages/genkit_openai/lib/src/embed.dart
+++ b/packages/genkit_openai/lib/src/embed.dart
@@ -109,6 +109,10 @@ String _documentText(DocumentData document, {required int index}) {
 /// and what it would have returned. Only curated models are checked — an
 /// uncurated name has no claim to check against, so its request goes through
 /// and OpenAI has the last word either way.
+///
+/// The caller decides when the catalog is authoritative enough to reject on:
+/// it describes OpenAI's models, and a compatible host serving one of those
+/// names may not have the same limits.
 void validateEmbedderDimensions(String embedderName, int? dimensions) {
   if (dimensions == null) return;
   final curated = knownOpenAIEmbedderFor(embedderName);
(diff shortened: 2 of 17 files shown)


packages/genkit_openai/lib/src/known_embedders.dart
```diff
@@ -0,0 +1,193 @@
...
+/// alongside the `label` they write there — so the Dev UI has something to
+/// read and the values do not have to be invented twice later. The keys are
+/// JS `EmbedderInfo`'s: `label`, `dimensions`, `supports.input`.
+Map<String, dynamic> _info({String? label, int? dimensions, String? stage}) =>
+    Map.unmodifiable(<String, dynamic>{
+      'label': ?label,
+      'dimensions': ?dimensions,
+      'stage': ?stage,
+      'supports': Map<String, dynamic>.unmodifiable(_textInput),
+    });
```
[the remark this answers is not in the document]
adesinah: Null-aware map entries are valid Dart as of 3.10, which is this package's lower bound (sdk: ^3.10.0). `dart analyze` is clean and the suite passes on 3.13.2. The same syntax is throughout core's generated types — see packages/genkit/lib/src/types.g.dart. Leaving as is.

packages/genkit_openai/lib/src/embed.g.dart
```diff
@@ -0,0 +1,116 @@
...
+base class OpenAIEmbedderOptions {
+  /// Creates a [OpenAIEmbedderOptions] from a JSON map.
+  factory OpenAIEmbedderOptions.fromJson(Map<String, dynamic> json) =>
+      $schema.parse(json);
+
+  OpenAIEmbedderOptions._(this._json);
+
+  OpenAIEmbedderOptions({int? dimensions, String? user}) {
+    _json = {'dimensions': ?dimensions, 'user': ?user};
+  }
```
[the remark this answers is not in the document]
adesinah: Same as above — null-aware map entries are Dart 3.10 syntax and compile fine. This file is also generated by schemantic_builder, and core's types.g.dart emits exactly the same shape, so hand-editing it would just be undone on the next build.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -247,9 +264,39 @@ class OpenAIPlugin extends GenkitPlugin {
...
+  ) {
+    final base = embedderMetadata(
+      '$_pluginName/$embedderName',
+      customOptions: embed.embedderOptionsSchema(),
+    );
+    final metadata = {...base.metadata};
+    metadata['model'] = <String, dynamic>{
+      ...(metadata['model'] as Map).cast<String, dynamic>(),
+      ..._embedderInfoFor(embedderName),
+    };
```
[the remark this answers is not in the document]
adesinah: `embedderMetadata` sets 'model' unconditionally — packages/genkit/lib/src/ai/embedder.dart, same repo — so the key is always there. Tolerating null would hide it if that ever stopped being true rather than fail where the assumption broke.

packages/genkit_openai/lib/genkit_openai.dart
```diff
@@ -18,11 +18,25 @@ import 'package:genkit/plugin.dart';
...
 import 'src/openai_plugin.dart';
 
 export 'src/chat.dart' show OpenAIChatOptions, OpenAIOptions;
 export 'src/converters.dart' show GenkitConverter;
+export 'src/embed.dart' show OpenAIEmbedderOptions;
+// The embedder catalog is public for the same reason the model catalog is.
+// `embedderInfoFor` and its compat variant are not: until core grows an
+// `EmbedderInfo` (#327) they hand back a raw map whose shape is expected to
+// change, and freezing that as API now would make the migration a breaking
+// one.
```
cabljac: The comment here keeps `embedderInfoFor` and `compatEmbedderInfo` private because they hand back a raw map whose shape changes at #327, and freezing it as API now makes that migration breaking. Agreed on the reasoning, but the export two lines down gives the same shape away:

    KnownOpenAIEmbedder.info        // known_embedders.dart:100, public getter
    knownOpenAIEmbedders            // {id: embedder.info}

`info` returns the same `_curatedInfo` instance `embedderInfoFor` returns, so a caller reading `KnownOpenAIEmbedder.textEmbedding3Small.info['dimensions']` is depending on exactly the shape we're trying not to commit to, and #327 breaks them either way.

Two ways out and I don't mind which: drop `info` and `knownOpenAIEmbedders` from the export and keep the typed enum fields (`id`, `label`, `dimensions`, `dimensionsReducible`, `stage`), or accept the map as API and export `embedderInfoFor` too.

packages/genkit_openai/lib/src/embed.dart
```diff
@@ -0,0 +1,159 @@
...
+/// response model cannot parse, so offering it would only let a caller break
+/// the call.
+@Schema()
+abstract class $OpenAIEmbedderOptions {
+  /// Length of the returned vector.
+  ///
+  /// Only the `text-embedding-3-*` models accept this, and only to shorten:
+  /// they are trained so that a prefix of the vector is still a usable
+  /// embedding. Defaults to the model's full size.
+  int? get dimensions;
```
cabljac, line 41: nit: `dimensions` has no lower bound on the schema, and `validateEmbedderDimensions` is skipped whenever baseUrl is set (openai_plugin.dart:346), not just for uncurated names. So `dimensions: 0` or a negative reaches the wire for any model on a compat host. Go bounds it at the schema instead:

    // go/plugins/compat_oai/config.go:163
    Dimensions int `json:"dimensions,omitempty" jsonschema:"minimum=1"`

A `minimum: 1` on the annotation would cover the compat path and give the Dev UI the hint too.

packages/genkit_openai/lib/src/known_embedders.dart
```diff
@@ -0,0 +1,193 @@
...
+  Map<String, dynamic> get info => _curatedInfo[this]!;
+}
+
+/// The shape core will grow a type for in #327.
+///
+/// Until `Embedder` / `embedderMetadata()` carry an `EmbedderInfo` of their
+/// own, this rides in the `model` metadata map that both already merge —
+/// alongside the `label` they write there — so the Dev UI has something to
+/// read and the values do not have to be invented twice later. The keys are
+/// JS `EmbedderInfo`'s: `label`, `dimensions`, `supports.input`.
```
cabljac: nit: the doc comment says the keys are JS `EmbedderInfo`'s, but `stage` isn't one of them. JS `EmbedderInfoSchema` is label / supports.{input,multilingual} / dimensions (js/ai/src/embedder.ts:302), and Go and Python match it.

`stage` looks deliberate to me, it's the same vocabulary as JS `ModelInfo.stage` (js/ai/src/model-types.ts:97), so I'd keep the key and just reword the comment to say label/dimensions/supports come from `EmbedderInfo` and `stage` mirrors `ModelInfo`.

packages/genkit_openai/lib/src/known_embedders.dart
```diff
@@ -0,0 +1,193 @@
...
+
+  /// The second-generation embedder, superseded by the `-3-` pair.
+  ///
+  /// Predates Matryoshka training, so its vector length is fixed: it rejects
+  /// the `dimensions` parameter rather than truncating.
+  textEmbeddingAda002(
+    'text-embedding-ada-002',
+    'OpenAI text-embedding-ada-002',
+    dimensions: 1536,
+    dimensionsReducible: false,
```
cabljac, line 63: `dimensionsReducible` has no sibling equivalent. JS (compat-oai/src/openai/embedder.ts:22), Go (config.go:160) and Python (openai_plugin.py:400) all accept `dimensions` for every model and let OpenAI 400 on ada-002.

I think rejecting locally is the better behaviour, the API's 400 names only the parameter whereas we can name the model. But it leaves Dart the strict one of the four, and the validation is skipped for compat hosts anyway, so the divergence only shows on OpenAI's own host. Do you agree that's the right split, or is this something we'd want to push back to the other SDKs?

cabljac (changes requested):
Catalog and wire plumbing look right, and the tests hold up: the ordering test uses a reversed fake whose vectors encode their index, and the compat-host validation bypass is tested against a real baseUrl rather than mocked out. I checked the ids and sizes against JS, Go and Python and all three agree with the enum.

One thing I'd like changed before merge, on the public surface (inline). The rest are nits, plus one open question on dimensionsReducible.

packages/genkit_openai/lib/genkit_openai.dart
adesinah: Took the first option — `info` is private now and `knownOpenAIEmbedders` is off the export, so the typed enum fields are the only public way in. f25afeb.

packages/genkit_openai/lib/src/embed.dart
adesinah: Added `@IntegerField(minimum: 1)`, so it's bounded on the schema and the compat path is covered. f25afeb.

packages/genkit_openai/lib/src/known_embedders.dart
adesinah: Reworded — label/dimensions/supports are EmbedderInfo's, stage mirrors ModelInfo. f25afeb.

packages/genkit_openai/lib/src/known_embedders.dart
adesinah: Agreed, keeping it. Naming the model beats a 400 that only names the parameter, and it only bites on OpenAI's own host. Not worth pushing to the other SDKs over.

cabljac (approved):
lgtm

PR #430, opened 2026-09-11
feat(genkit_openai): support reasoning effort, verbosity and reasoning output

### LINKED ISSUE #239: [genkit_openai] Support reasoning_effort parameter for OpenAI models
### Description

  The genkit_openai plugin currently does not expose the reasoning_effort parameter, which is a key configuration option for these models.

### Current Behavior

  OpenAIOptions includes standard sampling parameters (temperature, topP, maxTokens, etc.), but there is no way to pass reasoning_effort through the
   plugin. The ChatCompletionCreateRequest in openai_plugin.dart is constructed without it:
```dart
  final request = sdk.ChatCompletionCreateRequest(
    model: options.version ?? modelName,
    messages: ...,
    temperature: options.temperature,
    topP: options.topP,
    maxCompletionTokens: options.maxTokens,
    // reasoning_effort is not passed
  );
```
###  Expected Behavior

  OpenAIOptions should support a reasoningEffort field (e.g., minimal, low, medium, high) so that users can control the reasoning depth of o-series
  models. The underlying openai_dart SDK already fully supports this via CreateChatCompletionRequest.reasoningEffort.

###  Suggested Change

  1. Add reasoningEffort to $OpenAIOptions in genkit_openai.dart:
```dart
  @StringField(enumValues: ['minimal', 'low', 'medium', 'high'])
  String? get reasoningEffort;
```
2. Pass it through in _createModel when building the request:
```dart
  final request = sdk.ChatCompletionCreateRequest(
    ...
    reasoningEffort: options.reasoningEffort != null
        ? sdk.ReasoningEffort.values.byName(options.reasoningEffort!)
        : null,
  );
```
###  Context

  - Reducing or Increasing reasoning effort results in faster or slower responses and fewer or more tokens used, which is important for cost/latency/quality optimization.
  - The `verbosity` parameter (controlling response conciseness) may also be worth exposing for the same reason.

adesinah:
Adds `reasoningEffort` and `verbosity` to `OpenAIChatOptions` and wires both into the request.

Which levels a model takes moves with the generation — `minimal` came with GPT-5, `none` replaced it in GPT-5.1, `xhigh` later — so every level is offered to every reasoning model and OpenAI decides whether the pair makes sense. The set itself is fixed by `openai_dart`, which models the parameter as an enum; a level OpenAI ships after this release needs an SDK bump to reach.

What is curated is a `reasons` flag on `KnownOpenAIModel`. Sending an effort to `gpt-4o` is rejected before the request goes out, naming the model — OpenAI's own 400 says `Unsupported parameter: 'reasoning_effort'`, which doesn't tell you which half is wrong, and an effort set once in a shared config and then inherited by whatever model the call picks is the realistic way to hit it. The check follows `version` when set, since that is the model that actually answers, and names the spelling that will be on the wire rather than the curated alias it resolved through. The level is validated first: `gpt-4o` with `'extreme'` has two things wrong with it, and the level is the half a caller can fix without knowing the catalog. Skipped for a model registered through `models:` — declaring it says more about what it accepts than the catalog does — and skipped off OpenAI's host, same rule the rest of the catalog follows.

That host test used to be `baseUrl == null`, which answered two questions at once: passing OpenAI's own URL explicitly meant "compat backend", withholding the curated catalog from `list()` and `_infoFor` and skipping `validateEmbedderDimensions` (#429). It is now an `_isOpenAIHost` getter comparing the host, so both spellings of the same host behave alike; Azure's `*.openai.azure.com` is a different API and deliberately does not match.

Also the response side, which the issue thread asked to bundle: `reasoning_content`, `reasoning` and OpenRouter's structured `reasoning_details` now become a `ReasoningPart` ahead of the answer, and stream as deltas instead of only landing in the final message. Worth knowing this is entirely a compat-backend path — OpenAI bills reasoning tokens and never returns the text; DeepSeek R1 and vLLM send one spelling, OpenRouter the others, and all three are read. Detail entries are slices of one trace so they are joined; `reasoning.encrypted` ones are skipped, since their payload is base64 only the provider can open.

`verbosity` gets no catalog gate. It is a GPT-5-family parameter, but nothing in the catalog says so, and asserting it per entry would be guessing at something less well documented than which models reason.

`reasons` still does not reach `ModelInfo`, so the Dev UI offers `reasoningEffort` on every model and the call then fails. That needs a `supports` key the other SDKs agree to — JS's is a closed `z.object` — so it is filed as #475 rather than done here.

Closes #239.

### DIFF
--- a/packages/genkit_openai/lib/src/chat.dart
+++ b/packages/genkit_openai/lib/src/chat.dart
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'package:genkit/plugin.dart';
 import 'package:openai_dart/openai_dart.dart';
 import 'package:schemantic/schemantic.dart';
 
@@ -57,6 +58,31 @@ abstract class $OpenAIChatOptions {
   /// Visual detail level for images ('auto', 'low', 'high')
   @StringField(enumValues: ['auto', 'low', 'high'])
   String? get visualDetailLevel;
+
+  /// How hard a reasoning model should think before answering.
+  ///
+  /// Accepted by the o-series and the GPT-5 family; a model that does not
+  /// reason rejects the parameter outright. Which of these levels a given
+  /// model takes moves with the generation — `minimal` arrived with GPT-5,
+  /// `none` replaced it in GPT-5.1, `xhigh` came later still — so every level
+  /// is offered to every reasoning model and OpenAI decides whether the pair
+  /// makes sense.
+  ///
+  /// The set itself is closed, and not by choice: `openai_dart` models
+  /// `reasoning_effort` as an enum, so a level newer than the SDK cannot be
+  /// put on the wire as anything but `unknown`. A level OpenAI ships after
+  /// this release needs an `openai_dart` bump to reach it.
+  @StringField(
+    enumValues: ['none', 'minimal', 'low', 'medium', 'high', 'xhigh', 'max'],
+  )
+  String? get reasoningEffort;
+
+  /// How much the model should say in its answer.
+  ///
+  /// A GPT-5-family parameter, and unrelated to [reasoningEffort]: this
+  /// shortens or lengthens the reply, not the thinking behind it.
+  @StringField(enumValues: ['low', 'medium', 'high'])
+  String? get verbosity;
 }
 
 /// Alias for [OpenAIChatOptions].
@@ -89,6 +115,40 @@ ResponseFormat? buildOpenAIResponseFormat(Map<String, dynamic>? schema) {
 /// Returns custom options schema for standard chat models.
 SchemanticType<ChatModelOptions> chatModelOptionsSchema() =>
     OpenAIChatOptions.$schema;
+
+/// Maps a [ChatModelOptions.reasoningEffort] string onto the SDK enum.
+///
+/// `ReasoningEffort.fromJson` answers `unknown` for a value it does not
+/// recognise, which would go on the wire as `reasoning_effort: "unknown"` and
+/// come back as a confusing 400. Better to name the value that was actually
+/// wrong — whether it is a typo or a level newer than the SDK, which is the
+/// one case where the caller is right and the plugin is behind.
+ReasoningEffort? toReasoningEffort(String? value) {
+  if (value == null) return null;
+  final effort = ReasoningEffort.fromJson(value);
+  if (effort == ReasoningEffort.unknown) {
+    throw GenkitException(
+      'Unknown reasoningEffort "$value". Known levels: '
+      '${ReasoningEffort.values.where((e) => e != ReasoningEffort.unknown).map((e) => e.toJson()).join(', ')}.',
+      status: StatusCodes.INVALID_ARGUMENT,
+    );
+  }
+  return effort;
+}
+
+/// Maps a [ChatModelOptions.verbosity] string onto the SDK enum, rejecting an
+/// unrecognised level for the same reason [toReasoningEffort] does.
+Verbosity? toVerbosity(String? value) {
+  if (value == null) return null;
+  final verbosity = Verbosity.fromJson(value);
+  if (verbosity == Verbosity.unknown) {
+    throw GenkitException(
+      'Unknown verbosity "$value".',
+      status: StatusCodes.INVALID_ARGUMENT,
+    );
+  }
+  return verbosity;
+}
 
 /// Parses chat-model options from action config.
 ChatModelOptions parseChatModelOptions(Map<String, dynamic>? config) {
--- a/packages/genkit_openai/lib/src/chat.dart
+++ b/packages/genkit_openai/lib/src/chat.dart
@@ -143,7 +143,8 @@ Verbosity? toVerbosity(String? value) {
   final verbosity = Verbosity.fromJson(value);
   if (verbosity == Verbosity.unknown) {
     throw GenkitException(
-      'Unknown verbosity "$value".',
+      'Unknown verbosity "$value". Known levels: '
+      '${Verbosity.values.where((v) => v != Verbosity.unknown).map((v) => v.toJson()).join(', ')}.',
       status: StatusCodes.INVALID_ARGUMENT,
     );
   }
--- a/packages/genkit_openai/lib/src/converters.dart
+++ b/packages/genkit_openai/lib/src/converters.dart
@@ -245,6 +245,12 @@ abstract final class GenkitConverter {
   static Message fromOpenAIAssistantMessage(sdk.AssistantMessage msg) {
     final parts = <Part>[];
 
+    // Reasoning comes before the answer it produced, so it leads the content.
+    final reasoning = reasoningTextOf(msg.reasoningContent, msg.reasoning);
+    if (reasoning != null) {
+      parts.add(ReasoningPart(reasoning: reasoning));
+    }
+
     // Handle refusal
     if (msg.refusal != null && msg.refusal!.isNotEmpty) {
       parts.add(TextPart(text: '[Refusal] ${msg.refusal}'));
@@ -274,6 +280,24 @@ abstract final class GenkitConverter {
     }
 
     return Message(role: Role.model, content: parts);
+  }
+
+  /// The reasoning text carried by an assistant message or a stream delta.
+  ///
+  /// Two spellings reach this plugin and neither is OpenAI's: `reasoning` is
+  /// what OpenRouter emits, `reasoning_content` what DeepSeek R1 and vLLM do.
+  /// OpenAI's own models never return their chain of thought on the chat API
+  /// at all - they bill it as reasoning tokens and keep it - so this is
+  /// entirely a compatible-backend path.
+  ///
+  /// Both are read because a gateway in front of several providers passes
+  /// through whichever its upstream used. When both are present they are the
+  /// same text under two names, so the first wins rather than being joined.
+  static String? reasoningTextOf(String? reasoningContent, String? reasoning) {
+    for (final candidate in [reasoningContent, reasoning]) {
+      if (candidate != null && candidate.isNotEmpty) return candidate;
+    }
+    return null;
   }
 
   /// Map OpenAI finish reason to Genkit FinishReason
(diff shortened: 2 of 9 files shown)


packages/genkit_openai/lib/src/chat.dart
```diff
@@ -90,6 +116,40 @@ ResponseFormat? buildOpenAIResponseFormat(Map<String, dynamic>? schema) {
...
+/// unrecognised level for the same reason [toReasoningEffort] does.
+Verbosity? toVerbosity(String? value) {
+  if (value == null) return null;
+  final verbosity = Verbosity.fromJson(value);
+  if (verbosity == Verbosity.unknown) {
+    throw GenkitException(
+      'Unknown verbosity "$value".',
+      status: StatusCodes.INVALID_ARGUMENT,
+    );
+  }
```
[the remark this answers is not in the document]
adesinah: Good call, done in b6c77a7.

packages/genkit_openai/test/openai_plugin_reasoning_test.dart
```diff
@@ -0,0 +1,340 @@
...
+        'created': 0,
+        'model': 'o4-mini',
+        'choices': [
+          {
+            'index': 0,
+            'message': {
+              'role': 'assistant',
+              'content': 'the answer',
+              'reasoning_content': ?reasoningContent,
+              'reasoning': ?reasoning,
```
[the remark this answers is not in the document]
adesinah: That syntax is fine — null-aware map entries landed in Dart 3.10 and this package is on `sdk: ^3.10.0`. It's not just allowed, it's required here: swapping in the collection-if version makes `dart analyze --fatal-infos` fail with two `use_null_aware_elements` infos, which is what CI runs. Same form is used throughout genkit's own generated types.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -277,6 +277,15 @@ class OpenAIPlugin extends GenkitPlugin {
       fn: (req, ctx) async {
         final modelRequest = req!;
         final options = chat.parseChatModelOptions(modelRequest.config);
+        // `version` overrides the resolved action's id, so it - not
+        // [modelName] - is the model that will answer, and the model the
+        // catalog has to be asked about.
+        final wireModel = options.version ?? modelName;
+        if (baseUrl == null) {
```
cabljac, line 284: The gate fires only when `baseUrl == null`, and `_infoFor` (:259-260) and `list()` (:234) test the same condition. So opting out of the reasoning gate also stops the curated catalog merging: with discovery empty, `list()` gives 0 against `https://api.openai.com/v1` and 25 against `null`, and `versions`/`stage` drop from metadata. Both spellings dial the same host. The compat behaviour is deliberate (known_models_test.dart:282-293); applying it to OpenAI's own URL is the part I would question, and this PR's live test uses that URL.

Second time this flag has done too much: on #429 it switched off `validateEmbedderDimensions`. Splitting "is this the provider's own host" from "should I skip my own checks" would settle both. Worth doing here, or its own change on top?

cabljac (commented):
Nothing blocking. Two worth changing, two that #443 already fixes and could come down here, and two nits.

The `baseUrl` flag carries four meanings at once, and opting out of the reasoning gate also drops the curated catalog. That is the one I would most like your read on.

Wire shape, streaming parity and the level set all check out. The seven values match the API reference exactly, `max` included.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -350,6 +361,33 @@ class OpenAIPlugin extends GenkitPlugin {
...
+  ///
+  /// Only curated models are judged. An uncurated name has no claim to judge
+  /// against, and a model released after this version of the plugin must not
+  /// be refused a parameter it may well accept.
+  ///
+  /// `verbosity` gets no equivalent guard, and its 400 is just as terse. It is
+  /// a GPT-5-family parameter, but the catalog carries no flag saying so, and
+  /// adding one would mean asserting per entry something less well documented
+  /// than which models reason. Left to the API rather than guessed at.
+  void _requireReasoningSupport(String modelName, String? reasoningEffort) {
```
cabljac, line 379: This consults `knownOpenAIModelFor(name)` only, so `models: [CustomModelDefinition(name: 'gpt-4o', info: ...)]` with `reasoningEffort: 'low'` is refused before any HTTP request, even though `_createModel` holds that `info` at :270-276. Registering a model explicitly says more about it than the catalog does.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -350,6 +361,33 @@ class OpenAIPlugin extends GenkitPlugin {
...
+    if (reasoningEffort == null) return;
+    final curated = knownOpenAIModelFor(modelName);
+    if (curated == null || curated.reasons) return;
+
+    throw GenkitException(
```
cabljac, line 384: Two things, both already fixed in #443.

`curated.id` here means `version: 'gpt-4o-2024-11-20'` reports "gpt-4o does not accept reasoningEffort"; feat/xAI-via-openai-client:498-506 uses `$modelName`. And the model guard at :284-288 precedes level validation at :327, so `gpt-4o` + `'extreme'` names the model, never the level. Nothing catches it earlier: chat.g.dart:298-300 parses by casting, so the schema's `enumValues` reject nothing. feat/xAI-via-openai-client:339-341 validates the level first.

Both are regressions until #443 lands.

packages/genkit_openai/lib/src/converters.dart
```diff
@@ -276,6 +282,24 @@ abstract final class GenkitConverter {
...
+  /// Two spellings reach this plugin and neither is OpenAI's: `reasoning` is
+  /// what OpenRouter emits, `reasoning_content` what DeepSeek R1 and vLLM do.
+  /// OpenAI's own models never return their chain of thought on the chat API
+  /// at all - they bill it as reasoning tokens and keep it - so this is
+  /// entirely a compatible-backend path.
+  ///
+  /// Both are read because a gateway in front of several providers passes
+  /// through whichever its upstream used. When both are present they are the
+  /// same text under two names, so the first wins rather than being joined.
+  static String? reasoningTextOf(String? reasoningContent, String? reasoning) {
```
cabljac, line 296: nit: `reasoning_details` is unread. openai_dart 8.1.0 has it on the message (chat_message.dart:388), the delta (chat_stream_event.dart:295) and the accumulator (:515); this function and openai_plugin.dart:411-414 read `reasoning_content`/`reasoning` only. An OpenRouter response carrying only that field yields no ReasoningPart on either path. It still reaches `response.raw`, and no sibling SDK reads it either.

packages/genkit_openai/lib/src/known_models.dart
```diff
@@ -425,6 +460,20 @@ enum KnownOpenAIModel {
...
+  /// A model that does not reason rejects the parameter outright, which is
+  /// the whole reason this is curated: setting an effort once in a shared
+  /// config and then switching models is an easy mistake, and a 400 naming
+  /// only the parameter does not say which of the two is wrong.
+  ///
+  /// Deliberately a flag and not a list of levels. Which levels a model takes
+  /// moves with the generation, and enumerating that per entry would be
+  /// guessing at values OpenAI can change under us; the value itself is left
+  /// for the API to judge.
+  final bool reasons;
```
cabljac, line 475: nit: `reasons` is read at one site in `lib/` (openai_plugin.dart:382) and never reaches `ModelInfo`, while `customOptions` advertises `reasoningEffort` on every model (:248, :275). So the Dev UI offers it on `gpt-4o` and the call then fails.

Not for this PR: Dart's `supports` is `Map<String, dynamic>` and would carry a new key, but JS's `ModelInfoSchema.supports` is a closed `z.object` (js/ai/src/model-types.ts:75-96), so a Dart-only key is off-spec cross-SDK. Probably an issue rather than a change here.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Done here. `baseUrl == null` was standing in for two questions; there is now an `_isOpenAIHost` getter that answers only the first, comparing the host so `https://api.openai.com/v1` and the null default are the same thing. The reasoning gate, `list()`, `_infoFor` and `validateEmbedderDimensions` all read it, so the embedders case from #429 is covered too. Azure's `*.openai.azure.com` is deliberately not matched. Test in known_models_test.dart pins that the explicit URL keeps label, stage and versions.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Fixed. A name in `models:` now skips the guard entirely - registering it says more than the catalog does, so it goes to the API like an uncurated name. Covered by 'a registered model outranks the catalog'.

packages/genkit_openai/lib/src/openai_plugin.dart
adesinah: Both fixed here rather than waiting on #443. The message names `$modelName`, so a `version` of `gpt-4o-2024-11-20` reports that spelling, and level validation now runs before the model guard - `gpt-4o` + `'extreme'` names the level, which is the half a caller can fix without knowing the catalog. Two tests pin each.

packages/genkit_openai/lib/src/converters.dart
adesinah: Added. `reasoningTextOf` takes `reasoningDetails` as a third source, after the flat fields since those need no assembly. Entries are slices of one trace so they are joined; `reasoning.encrypted` ones are skipped, the payload is base64 only the provider can open. Read on both the message and the delta, with tests for each.

packages/genkit_openai/lib/src/known_models.dart
adesinah: Agreed, leaving it. The cross-SDK bit is the blocker: a Dart-only `supports` key is off-spec while JS's is a closed object, so this wants an issue against the shared model spec rather than a change here. I'll open one.

adesinah: Opened #475 for it.

PR #439, opened 2026-09-14
feat(genkit_openai): add first-class DeepSeek support

### LINKED ISSUE #133: DeepSeek via OpenAI client

adesinah:
Stacked on #430, which this needs for the `reasoning_content` mapping.

Adds a `deepSeek()` handle beside `openAI()`, with a curated catalog and typed refs. Same plugin underneath — DeepSeek speaks the same protocol against a different host — so the differences sit in one `OpenAIProvider` value instead of spreading through the request builder.

The catalog lists `deepseek-flash` and `deepseek-v4-pro`, plus `deepseek-chat` and `deepseek-reasoner` as legacy: DeepSeek announced their discontinuation for 2026-07-24 but still serves both, routing them to Flash's non-thinking and thinking modes, so they stay listed until the names stop answering.

Nothing about thinking is curated. On DeepSeek it is a request-time mode, not a property of a name — `deepseek-chat` answers a `reasoningEffort` of `high` with reasoning tokens — so no level and no model is refused locally; the host decides.

The differences the provider value carries: `DEEPSEEK_API_KEY`, `max_tokens` rather than `max_completion_tokens`, `json_object` rather than `json_schema`, reasoning replayed on tool-carrying requests, and the `thinking` object. `reasoning_effort` goes out at the top level as it does everywhere, with `thinking: {type: enabled}` alongside it to say which mode it applies to; `none` is `{type: disabled}` with no effort. No level is refused locally — DeepSeek maps `minimal` to `low` and `medium`/`xhigh` to `high` — so the levels the option advertises are the levels that work.

339 unit tests, plus 8 live against DeepSeek covering thinking, effort levels, `max_tokens`, structured output, a tool loop and streaming.

Closes #133.


packages/genkit_openai/lib/src/chat_body_client.dart
```diff
@@ -0,0 +1,66 @@
...
+        !request.url.path.endsWith('/chat/completions')) {
+      return _inner.send(request);
+    }
+
+    final decoded = jsonDecode(request.body);
+    if (decoded is! Map) return _inner.send(request);
+
+    final rewritten = _rewrite(decoded.cast<String, dynamic>());
+    final replacement = http.Request(request.method, request.url)
+      ..followRedirects = request.followRedirects
```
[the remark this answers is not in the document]
adesinah: Good catch, both applied. The encoding one is belt-and-braces — the charset rides on the content-type header that's already copied verbatim — but it's right for any request, not just the ones the SDK builds.

packages/genkit_openai/lib/src/chat_body_client.dart
```diff
@@ -0,0 +1,66 @@
...
+  @override
+  Future<http.StreamedResponse> send(http.BaseRequest request) {
+    if (request is! http.Request ||
+        request.method != 'POST' ||
+        !request.url.path.endsWith('/chat/completions')) {
+      return _inner.send(request);
+    }
+
+    final decoded = jsonDecode(request.body);
+    if (decoded is! Map) return _inner.send(request);
```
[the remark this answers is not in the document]
adesinah: Applied. Only the SDK's own encoded bodies can reach this path, so it should be unreachable, but a decorator that throws on a surprise is worse than one that gets out of the way.

packages/genkit_openai/lib/src/provider.dart
```diff
@@ -0,0 +1,171 @@
...
+  defaultNamespace: defaultDeepSeekNamespace,
+  defaultBaseUrl: 'https://api.deepseek.com',
+  apiKeyEnvVar: 'DEEPSEEK_API_KEY',
+  infoFor: (model, {required compat}) =>
+      compat ? compatDeepSeekModelInfo(model) : deepSeekModelInfoFor(model),
+  catalogIds: knownDeepSeekChatModels,
+  reasonsFor: (model) => knownDeepSeekModelFor(model)?.thinks,
+  // No `medium`, `minimal` or `xhigh`.
+  // https://api-docs.deepseek.com/api/create-chat-completion
+  reasoningEfforts: const {'none', 'low', 'high', 'max'},
```
cabljac, line 141: The reference at the URL just above says the opposite: "For compatibility with existing software, `minimal` is accepted and mapped to `low`, and `medium` / `xhigh` are accepted and mapped to `high`." So `reasoningEffort: 'medium'` on `deepseek-flash` fails locally where DeepSeek would have run it at `high`. The check at openai_plugin.dart:310 is not gated by `_ownHost` either, so a `deepSeek(baseUrl: ...)` gateway user is refused too. deepseek_test.dart:366 pins the restriction, so that one needs inverting rather than adjusting.

packages/genkit_openai/lib/src/provider.dart
```diff
@@ -0,0 +1,171 @@
...
+  rewriteChatBody: deepSeekChatBody,
+  usesLegacyMaxTokens: true,
+  supportsJsonSchema: false,
+);
+
+/// Moves `reasoning_effort` into the `thinking` object DeepSeek reads.
+///
+/// OpenAI takes `reasoning_effort` at the top level; DeepSeek nests it, and
+/// derives the thinking toggle from it — `none` disables thinking, every other
+/// level enables it. A top-level `reasoning_effort` is simply not read there,
```
cabljac, line 152: "A top-level `reasoning_effort` is simply not read there" is the one claim here I can't source. The thinking-mode guide's OpenAI-format samples pass `reasoning_effort="high"` top-level alongside `extra_body={"thinking": {"type": "enabled"}}`, so both locations show up in the vendor's own examples. Go asserts the reverse of this file (deepseek.go:88-91), also with a unit test pinning its guess and nothing live that could separate them.

Not suggesting dropping the hook: the `none` -> `thinking: {type: disabled}` half is documented and load-bearing. Sending both, top-level effort plus the thinking object, is the shape the docs and both siblings agree on. Do you agree that's the right read, or have you seen top-level get ignored on the wire?

packages/genkit_openai/lib/genkit_openai.dart
```diff
@@ -336,3 +351,108 @@ abstract final class OpenAIModels {
...
+  }
+
+  /// Reference to a DeepSeek model.
+  ModelRef<chat.OpenAIChatOptions> model(
+    String name, {
+    String namespace = defaultDeepSeekNamespace,
+  }) {
+    return modelRef(
+      '$namespace/$name',
+      customOptions: chat.chatModelOptionsSchema(),
```
cabljac, line 418: This hands DeepSeek the full OpenAI level set, so the Dev UI offers `minimal`, `medium` and `xhigh` and the plugin then refuses them. Mostly falls away if the levels above are accepted. JS and Go both narrow per provider (`DeepSeekChatCompletionConfigSchema`, `deepseek.ChatConfig`), which keeps the advertised options and the accepted ones the same.

packages/genkit_openai/lib/src/known_deepseek_models.dart
```diff
@@ -0,0 +1,198 @@
...
+    deepSeekTextSupports,
+    thinks: true,
+  ),
+
+  /// The former chat alias, now `deepseek-flash` with thinking off.
+  deepseekChat(
+    'deepseek-chat',
+    'DeepSeek Chat',
+    deepSeekVisionSupports,
+    stage: OpenAIModelStage.legacy,
```
cabljac, line 83: By the vocabulary at known_models.dart:141-149, `legacy` is "still served, with a shutdown date announced" and `deprecated` is "no longer served". DeepSeek retired both on 2026-07-24 and the reference now lists only `deepseek-flash` and `deepseek-v4-pro`, so these read as `deprecated` on this PR's own terms. The live catalog test walks `stable` only, so it cannot notice either way.

Related: `deepseek-chat` takes the default `thinks: false`, so it refuses `none` as well as a real effort, and `none` is what that alias means.

packages/genkit_openai/test/deepseek_test.dart
```diff
@@ -0,0 +1,719 @@
...
+      expect(requests.single.headers['authorization'], 'Bearer ds-key');
+    });
+
+    test('reads DEEPSEEK_API_KEY, and says so when it is missing', () async {
+      final plugin = OpenAIPlugin(
+        provider: deepSeekProvider,
+        configVar: (name) => name == 'DEEPSEEK_API_KEY' ? 'from-env' : null,
+        httpClient: recordingClient([]),
+      );
+      expect(plugin.name, 'deepseek');
```
cabljac, line 221: nit: this half asserts `plugin.name`, which comes from `provider.defaultNamespace`, not from `configVar`. A typo in `apiKeyEnvVar` would still pass it. The second half does prove the name is consulted. Untested is that the value reaches `Authorization: Bearer from-env`, which :212 does for the direct key.

cabljac (changes requested):
One thing to change before merge. DeepSeek documents `minimal`, `medium` and `xhigh` as accepted and mapped to `low`/`high`, so the rejection at provider.dart:141 turns working requests into a local INVALID_ARGUMENT, and a test pins it. The doc comment cites the page that says the opposite.

The rest are smaller, inline, plus one open question on the `thinking` rewrite.

packages/genkit_openai/lib/src/provider.dart
adesinah: You're right on both counts, and the restriction is gone - `reasoningEfforts` and the guard that read it are deleted. DeepSeek maps what it has no native setting for (`minimal` to `low`, `medium` and `xhigh` to `high`), so refusing locally failed requests the host would have served, and it did so for gateway users too. deepseek_test.dart now asserts the opposite: `minimal`, `medium` and `xhigh` all go out.

packages/genkit_openai/lib/src/provider.dart
adesinah: Agreed - I had no wire evidence for that claim, only the shape of the `thinking` object. Both now go out: `reasoning_effort` stays at the top level where every host reads it, and `thinking: {type: enabled}` says which mode it applies to, which is what the docs' samples and both siblings do. `none` still maps to `{type: disabled}` with the effort dropped, since there is no effort to spend. Tests and README updated.

packages/genkit_openai/lib/genkit_openai.dart
adesinah: Falls away with the levels accepted - nothing is refused now, so the advertised set and the accepted set match without narrowing the schema.

packages/genkit_openai/lib/src/known_deepseek_models.dart
adesinah: Both fixed. `deepseek-chat` and `deepseek-reasoner` are `deprecated` now, which by this catalog's own rule also drops them from the listing while they stay resolvable with honest capabilities. And `none` is never refused, whatever the model claims - it is how a caller asks for no thinking, so the alias that means exactly that could not be the one thing that rejects it. A test covers `deepseek-chat` with `none`.

packages/genkit_openai/test/deepseek_test.dart
adesinah: Fair - that half proved nothing. It now runs a generate with `DEEPSEEK_API_KEY` in `configVar` and asserts `Authorization: Bearer from-env`, plus the reverse: `OPENAI_API_KEY` set and `DEEPSEEK_API_KEY` not is a keyless plugin.

packages/genkit_openai/lib/src/provider.dart
```diff
@@ -0,0 +1,179 @@
...
+/// has to know about: the key it reads, the token-limit field, the absence of
+/// `json_schema`, and the `thinking` object that carries its mode.
+final deepSeekProvider = OpenAIProvider(
+  defaultNamespace: defaultDeepSeekNamespace,
+  defaultBaseUrl: 'https://api.deepseek.com',
+  apiKeyEnvVar: 'DEEPSEEK_API_KEY',
+  infoFor: (model, {required compat}) =>
+      compat ? compatDeepSeekModelInfo(model) : deepSeekModelInfoFor(model),
+  catalogIds: knownDeepSeekChatModels,
+  reasonsFor: (model) => knownDeepSeekModelFor(model)?.thinks,
```
cabljac: This refuses `reasoningEffort` on `deepseek-chat` locally (`thinks: false`), but DeepSeek serves it: `deepseek-chat` with `reasoning_effort: high` returns `reasoning_content` (31 reasoning tokens on "Is 91 prime?"), and `low` does the same. Thinking is a request-time mode on DeepSeek, not a model property, so the same INVALID_ARGUMENT problem as last round: a working request fails before it leaves. deepseek_test.dart:547 pins the restriction and needs inverting. Simplest is to return `null` here and leave the level to the host, as the compat path already does.

packages/genkit_openai/lib/src/known_deepseek_models.dart
```diff
@@ -0,0 +1,204 @@
...
+  /// `deprecated` rather than `legacy`: DeepSeek retired both aliases on
+  /// 2026-07-24 and the reference now lists only `deepseek-flash` and
+  /// `deepseek-v4-pro`. Curated anyway, so that code written against the name
+  /// gets the right capabilities and a stage that says what happened, rather
+  /// than the dynamic defaults.
+  deepseekChat(
+    'deepseek-chat',
+    'DeepSeek Chat',
+    deepSeekVisionSupports,
+    stage: OpenAIModelStage.deprecated,
```
cabljac: I was wrong last round: these are still served. `deepseek-chat`, `deepseek-reasoner` and `deepseek-v4-flash` all answer today, routed to `deepseek-flash` (chat non-thinking, reasoner thinking). By known_models.dart:159-166, that is `legacy`, not `deprecated`. The updates page announces no shutdown for the first two and says the V4 Flash names are "temporarily routed to V4.1 Flash", so `legacy` fits there too. Doc comment at :80 and deepseek_test.dart:156 follow. Sorry for sending you the wrong way on it.

cabljac (changes requested):
Ran the live suite and a few direct calls against DeepSeek today; all 8 live tests pass and the effort question from last round is settled (top-level reasoning_effort is read, low vs high shows a clear reasoning_tokens gradient with the thinking object alongside). Two things to change before merge, both from the live run, and one of them reverses a claim I made last round: the retired aliases are still served. Details inline.

packages/genkit_openai/lib/src/provider.dart
adesinah: You're right, and the same mistake twice - thanks for testing it. `reasonsFor` returns `null` for DeepSeek now, so nothing is refused locally on that basis; the level goes out and the host decides. `thinks` is gone from the catalog entirely rather than just unused: it asserted a model property DeepSeek does not have, since thinking is a request-time mode. deepseek_test.dart:547 is inverted - `deepseek-chat` with `high` now goes out with `reasoning_effort: high` and `thinking: {type: enabled}`.

packages/genkit_openai/lib/src/known_deepseek_models.dart
adesinah: No apology needed - I took the change, so the correction is mine to carry. Both are back to `legacy` and listed again, with the doc comment saying why: announced, still answering, so `legacy` is exactly the state. The V4 Flash spellings stay snapshots of `deepseek-flash`, with the comment now noting they are temporarily routed rather than permanent names. Test and README follow.

Issue #428, opened 2026-09-10
Docs: running the Dev UI against a Flutter app on a physical device


cabljac:
There is no documented way to run the Dev UI against a Flutter app on a physical device. Once the `dart:io` dart-define fallback lands (#427), the working recipe is four commands and should be in the Flutter docs.

```
genkit start --experimental-reflection-v2 --write-env-file /tmp/genkit.env --port 4000
# read the two ports from /tmp/genkit.env, then
adb reverse tcp:<reflectionPort> tcp:<reflectionPort>
adb reverse tcp:<telemetryPort> tcp:<telemetryPort>
flutter run -d <device> \
  --dart-define=GENKIT_ENV=dev \
  --dart-define=GENKIT_REFLECTION_V2_SERVER=ws://localhost:<reflectionPort> \
  --dart-define=GENKIT_TELEMETRY_SERVER=http://localhost:<telemetryPort> \
  --dart-define=GENKIT_ENABLE_REALTIME_TELEMETRY=true
```

`adb reverse` makes the device's `localhost` reach the host, so the same URLs work on an emulator and on a phone over USB. Rewriting to `10.0.2.2` only works on the emulator.

The page should also say:

- `promptDir` should be null in a Flutter app; loading `.prompt` files walks the host filesystem, which has no meaning in an app sandbox.
- No API key is involved when the model is on-device; the Dev UI lists, invokes and traces flows with nothing but the two reverse-forwarded ports.
- `genkit start:flutter` cannot run `adb reverse` for you today, hence the manual steps.

Found during #384.

PR #442, opened 2026-09-14
fix(genkit): read genkit dev-mode config from dart-defines on io

### LINKED ISSUE #427: Dev UI: dart:io dev-mode config reads only Platform.environment, so genkit start:flutter cannot reach an app on a device
`genkit start:flutter` cannot reach a Flutter app running on an Android device or emulator, because the `dart:io` dev-mode config reads only `Platform.environment`.

**What happens.** `genkit start:flutter` starts the reflection-v2 and telemetry servers and passes `GENKIT_ENV`, `GENKIT_REFLECTION_V2_SERVER`, `GENKIT_TELEMETRY_SERVER` and `GENKIT_ENABLE_REALTIME_TELEMETRY` to `flutter run` as `--dart-define`s. On a device the app resolves to the `dart:io` variants of three files, and all three read `Platform.environment` only:

- `packages/genkit/lib/src/utils_io.dart` (`isDevEnv`)
- `packages/genkit/lib/src/core/reflection/reflection_io.dart` (reflection URL)
- `packages/genkit/lib/src/o11y/telemetry/telemetry_io.dart` (telemetry URL)

An Android app is forked from the zygote and has no process environment, so every define lands in the binary and is ignored. The app never registers with the Dev UI. The web variants of the same three files already read `String.fromEnvironment`.

**Fix.** Each io site falls back to its `--dart-define` when the environment has no value. The define key has to be a compile-time literal, so the lookup cannot go through `getConfigVar(name)`; each site declares its own `const`. 32 added lines across the three files; the existing tests pass.

Verified on an Android emulator and a Galaxy S23+ over USB with `adb reverse` on the two ports: the Dev UI lists the device's actions, runs them, and receives traces with the model span.

**Not covered by this fix.** `genkit start:flutter` picks its ports and cannot set up `adb reverse` itself, so the device flow is still four commands rather than one. That is a docs item (separate issue) and possibly a CLI follow-up.

Found during #384. Patch ready, PR to follow with tests.

cabljac:
`genkit start:flutter` passes the dev-mode settings to `flutter run` as `--dart-define`s, but the three `dart:io` readers only looked at `Platform.environment`. Nothing can add to an app's environment on Android or iOS, so every define was dropped and the app never registered with the Dev UI. Desktop was unaffected, because the CLI also puts the same variables into the process environment.

Each setting now resolves through `resolveDevConfig(envValue, defineValue)`: environment wins, define is the fallback, empty counts as unset. Each key is declared once as a `const` and used for both channels.

Fixes #427

Before:
image

after:
image

Tested two ways: the precedence rules directly, and the wiring by running a probe app in a child process with real defines and a scrubbed environment. The second is needed because a define is baked in at compile time, so nothing in-process can check which key feeds which setting.

Also stops `ReflectionServerV1` starting on Android and iOS. It serves unauthenticated `runAction` on loopback, which a define made reachable on a device for the first time. A host Dev UI cannot reach a device's loopback anyway.

Not fixed here: on a physical device the app still needs two `adb reverse` rules, see #428. `GENKIT_ENABLE_REALTIME_TELEMETRY` still has no reader, and `reflection_v1.dart:399` still reads `GENKIT_RUNTIME_ID` from the environment only.


packages/genkit/test/dev_config_dart_define_test.dart
```diff
@@ -0,0 +1,244 @@
...
+
+Future<void> _waitFor(bool Function() condition, String describeFailure) async {
+  final deadline = DateTime.now().add(_waitLimit);
+  while (!condition()) {
+    if (DateTime.now().isAfter(deadline)) {
+      fail('$describeFailure within ${_waitLimit.inSeconds}s');
+    }
+    await Future<void>.delayed(const Duration(milliseconds: 50));
+  }
+}
```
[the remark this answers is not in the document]
cabljac: Taken in 9d076e7.

packages/genkit/test/dev_config_dart_define_test.dart
```diff
@@ -0,0 +1,244 @@
...
+
+    final stderrText = StringBuffer();
+    process.stderr.transform(utf8.decoder).listen(stderrText.write);
+    final ready = process.stdout
+        .transform(utf8.decoder)
+        .transform(const LineSplitter())
+        .firstWhere((line) => line.trim() == 'probe-ready', orElse: () => '')
+        .then((line) {
+          if (line.isEmpty) fail('probe exited before ready:\n$stderrText');
+        });
```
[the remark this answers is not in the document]
cabljac: Taken in 9d076e7. `onDone` completes with an empty line rather than an error, so the existing "exited before ready" failure path still fires with the captured stderr.

packages/genkit/test/dev_config_dart_define_test.dart
```diff
@@ -0,0 +1,244 @@
...
+    final server = await HttpServer.bind(InternetAddress.loopbackIPv4, 0);
+    final fake = _FakeReflectionServer._(server);
+    server.listen((request) async {
+      final socket = await WebSocketTransformer.upgrade(request);
+      socket.listen((data) {
+        if (data is String) {
+          fake.messages.add(jsonDecode(data) as Map<String, dynamic>);
+        }
+      }, onError: (Object _) {});
+    });
```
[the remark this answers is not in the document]
cabljac: Leaving this one. The fake only ever receives the probe's own connection, so a non-upgrade request means the probe built the wrong request. Letting that surface as an unhandled error fails the test loudly; a 400 response would swallow it and the test would time out instead.

packages/genkit/lib/src/core/reflection/reflection_io.dart
```diff
@@ -14,14 +14,34 @@
...
+  final v2ServerUrl = resolveDevConfig(
+    io.Platform.environment['GENKIT_REFLECTION_V2_SERVER'],
+    _v2ServerDefine,
+  );
+  final runtimeId =
+      resolveDevConfig(
+        io.Platform.environment['GENKIT_RUNTIME_ID'],
+        _runtimeIdDefine,
+      ) ??
+      '';
```
demolaf: Can we declare each key once and reuse it for both channels?

```suggestion
const String _v2ServerKey = 'GENKIT_REFLECTION_V2_SERVER';
const String _runtimeIdKey = 'GENKIT_RUNTIME_ID';

const String _v2ServerDefine = String.fromEnvironment(_v2ServerKey);
const String _runtimeIdDefine = String.fromEnvironment(_runtimeIdKey);

/// Starts the reflection server the Developer UI talks to.
///
/// Both settings come from the process environment, falling back to the
/// `--dart-define` of the same name for an app whose environment no launcher
/// can set. An empty `GENKIT_REFLECTION_V2_SERVER` selects the v1 server, as an
/// unset one does.
ReflectionServerHandle startReflectionServer(Registry registry, {int? port}) {
  final v2ServerUrl = resolveDevConfig(
    io.Platform.environment[_v2ServerKey],
    _v2ServerDefine,
  );
  final runtimeId =
      resolveDevConfig(
        io.Platform.environment[_runtimeIdKey],
        _runtimeIdDefine,
      ) ??
      '';
```

packages/genkit/lib/src/o11y/telemetry/telemetry_io.dart
```diff
@@ -14,12 +14,20 @@
...
-String? genkitTelemetryServerUrl() {
-  final server = io.Platform.environment['GENKIT_TELEMETRY_SERVER'];
-  return (server == null || server.isEmpty) ? null : server;
-}
+/// same configuration produces the same instrumentation decision across
+/// platforms.
+String? genkitTelemetryServerUrl() => resolveDevConfig(
+  io.Platform.environment['GENKIT_TELEMETRY_SERVER'],
+  _telemetryServerDefine,
+);
```
demolaf, line 34:
```suggestion
const String _telemetryServerKey = 'GENKIT_TELEMETRY_SERVER';
const String _telemetryServerDefine = String.fromEnvironment(
  _telemetryServerKey,
);

/// The Genkit telemetry server base URL from `GENKIT_TELEMETRY_SERVER` in the
/// process environment, falling back to the `--dart-define` of the same name,
/// or `null` when neither is set.
///
/// An empty value is treated as unset (matching the web implementation) so the
/// same configuration produces the same instrumentation decision across
/// platforms.
String? genkitTelemetryServerUrl() => resolveDevConfig(
  io.Platform.environment[_telemetryServerKey],
  _telemetryServerDefine,
);
```

packages/genkit/lib/src/utils_io.dart
```diff
@@ -14,10 +14,19 @@
...
 
-final bool isDevEnv = io.Platform.environment['GENKIT_ENV'] == 'dev';
+const String _devEnvDefine = String.fromEnvironment('GENKIT_ENV');
+
+/// Whether Genkit runs in dev mode, from `GENKIT_ENV` in the process
+/// environment, falling back to the `--dart-define` of the same name for an app
+/// whose environment no launcher can set.
+final bool isDevEnv =
+    resolveDevConfig(io.Platform.environment['GENKIT_ENV'], _devEnvDefine) ==
+    'dev';
```
demolaf:
```suggestion
const String _devEnvKey = 'GENKIT_ENV';
const String _devEnvDefine = String.fromEnvironment(_devEnvKey);

/// Whether Genkit runs in dev mode, from `GENKIT_ENV` in the process
/// environment, falling back to the `--dart-define` of the same name for an app
/// whose environment no launcher can set.
final bool isDevEnv =
    resolveDevConfig(io.Platform.environment[_devEnvKey], _devEnvDefine) ==
    'dev';
```

demolaf (commented):
Looks good, couple of nits but not priority.

packages/genkit/lib/src/core/reflection/reflection_io.dart
cabljac: Applied in 4aa2e52, across all three files.

You were right and my reason for not doing it was wrong. The key has to be a compile-time constant, not a literal, so a `const String` works fine. `reflection_web.dart:19-23` already relies on that. I checked it under AOT as well, since that is where a bad key would fail silently: `const String _key = 'GENKIT_ENV'; const v = String.fromEnvironment(_key);` compiled with `dart compile exe --define=GENKIT_ENV=dev` resolves to `dev`.

Also removed the claim from the `dev_config.dart` doc comment, which asserted the same wrong thing.

packages/genkit/lib/src/o11y/telemetry/telemetry_io.dart
cabljac: Applied in 4aa2e52. See the reflection_io thread for why your version is correct.

packages/genkit/lib/src/utils_io.dart
cabljac: Applied in 4aa2e52.

packages/genkit/lib/src/core/reflection/reflection_io.dart
```diff
@@ -14,25 +14,88 @@
...
+        runtimeId: runtimeId,
+      );
+      server.start();
+      return ReflectionServerHandle(server.stop);
+    case ReflectionTransport.v1:
+      final server = ReflectionServerV1(registry, port: port);
+      server.start();
+      return ReflectionServerHandle(server.stop);
+    case ReflectionTransport.none:
+      _logger.severe(
```
pavelgj: I'm debating whether this should be `severe` or `warning`... I'm leaning towards warning.

```suggestion
      _logger.warning(
```

pavelgj (approved):

PR #443, opened 2026-09-14
feat(genkit_openai): add first-class xAI support

### LINKED ISSUE #134: xAI via OpenAI client

adesinah:
Stacked on #439, which introduced the provider dialect this builds on.

Adds an `xAI()` handle with a `KnownXaiModel` catalog and `XaiModels` refs. Of the curated providers it is the closest to OpenAI — same request fields, same `json_schema` structured outputs — so it needed no new machinery, just a catalog and a provider instance. That was the point of the dialect, and this is the first evidence it holds.

xAI's reasoning levels are `none`, `low`, `medium`, `high`, `xhigh` — no `minimal` or `max`. xAI documents the accepted set as varying per model, so the plugin checks the union and leaves the pairing to the API.

Two fixes fall out of this beyond the preset:

- `getModelType` only recognised `sora` as video, so xAI's `grok-imagine-video-*` would have been listed as chat models. Matched narrowly, since a bare `video` substring runs ahead of every chat rule and would swallow a video-understanding chat model on some compat host.
- `_ownHost` compared base URLs as raw strings, so `deepSeek(baseUrl: 'https://api.deepseek.com/v1')` — the spelling in DeepSeek's own docs, and the one our sample used — silently fell into compat mode and lost its catalog. Now normalised for case, trailing slash and the `/v1` suffix.

Closes #134.


packages/genkit_openai/lib/src/converters.dart
```diff
@@ -235,7 +236,13 @@ abstract final class GenkitConverter {
...
     // OpenAI requires parameters to be a valid JSON Schema object
     // If no schema is provided, use an empty object schema
-    var parameters = tool.inputSchema;
+    // Flattened first: schemantic emits a generated class's schema as a
+    // `$ref` into `$defs`, and hosts disagree about resolving one. OpenAI
+    // accepts it; xAI answers `tool parameter root must be an object type
+    // (root schema is a $ref)`. Inlining it costs nothing and is the shape
+    // every host documents. The response format is flattened for the same
+    // reason - see `buildOpenAIResponseFormat`.
+    var parameters = tool.inputSchema?.flatten().cast<String, dynamic>();
```
cabljac, line 245: `flatten()` throws `FormatException('Circular reference detected: #/$defs/<Name>')` on a self-referential schema (packages/schemantic/lib/src/flatten.dart:50).

A tool input type that references itself now fails. Examples: a tree node, a nested comment. The catch at openai_plugin.dart:419-438 wraps it as `GenkitException('OpenAI API error: ...')` with no status, and core returns `finishReason: failed`. Nothing reached the wire, so the message blames the API for a local schema problem. One call site at :355, so every provider is affected.

At #439 the same schema went out as `$ref` and `$defs`. OpenAI accepts that shape as far as I can tell; the only source is the comment this replaced.

The reason given in the comment is correct for xAI. Two options: catch the error and send the unflattened schema, or wrap it as INVALID_ARGUMENT. Add an OpenAI-side test either way. The only `$ref` test on the tool path is xai_test.dart:213. openai_plugin_chat_test.dart:74 covers `$defs` for the response format. Neither uses a recursive schema.

packages/genkit_openai/lib/src/openai_plugin.dart
```diff
@@ -55,6 +55,34 @@ class OpenAIPlugin extends GenkitPlugin {
...
+  }
+
+  static String _normalizeHost(String url) {
+    var normalized = url.toLowerCase();
+    while (normalized.endsWith('/')) {
+      normalized = normalized.substring(0, normalized.length - 1);
+    }
+    // The version segment is the one path suffix providers publish both with
+    // and without; anything else in the path is a different backend.
+    if (normalized.endsWith('/v1')) {
```
cabljac, line 80: This makes `https://api.x.ai` and `https://api.x.ai/v1` the same host for the catalog check. But the SDK sends the string as given. `request_builder.dart:45-67` only normalizes the trailing slash and merges query params. It never adds `/v1`.

xAI publishes `/v1` only. `POST https://api.x.ai/chat/completions` answers 404. `POST https://api.x.ai/v1/chat/completions` answers 400. So `xAI(baseUrl: 'https://api.x.ai')` gets the curated catalog, and then every request answers 404.

DeepSeek answers 401 for both spellings. The `/v1` strip was correct for #439. It is wrong for xAI.

Fix: when `_isOwnHost` matches, send `provider.defaultBaseUrl` instead of the caller's string. No existing test asserts the wire URL for an alternate spelling, so that should not break the suite.

packages/genkit_openai/lib/src/chat.dart
```diff
@@ -138,25 +138,26 @@ ResponseFormat? buildOpenAIResponseFormat(
...
+/// somewhere, so if a message already says it there is nothing to add — but a
+/// schema has nowhere else to go, and a prompt that merely mentions JSON
+/// ("extract the fields from this JSON log line") does not carry one. Dropping
+/// the schema because the word happened to appear would hand the model a
+/// free-form JSON request and fail validation on the way back.
 String? jsonObjectInstruction(
   List<Message> messages,
   Map<String, dynamic>? schema,
 ) {
+  if (schema != null) {
```
cabljac, line 151: The reason given above is correct, and the word test it replaced was worse. But this code now adds the schema every time, and one path sends the schema twice.

Core adds its own instruction first. The JSON formatter writes "Output should be in JSON format and conform to the following schema" plus the schema into the last system message, or the last user message when there is none (formatters.dart:174-176). A raw `GenerateActionOptions(instructions: true)` request therefore already carries the schema. This code adds a second copy. The reflection API passes `/util/generate` input through untouched (reflection_v1.dart:243-245), so any caller of it can set `instructions: true`. `ai.generate` cannot ask for the formatter's default text: its `outputInstructions` is a `String?`.

Core marks its own part with `metadata.purpose == 'output'` (formatters.dart:171). I checked that the metadata survives into `ModelRequest.messages`, which is what :381 hands to this function. Skip the schema when a part has that marker. The two paths then cannot both run, and no word test is needed.

This affects DeepSeek only. xAI sets `supportsJsonSchema: true` and never calls this function. So it is a fix to #439 in the xAI PR.

packages/genkit_openai/lib/src/known_xai_models.dart
```diff
@@ -0,0 +1,150 @@
...
+  ///
+  /// The only curated xAI model that takes no reasoning effort at all.
+  grok420NonReasoning(
+    'grok-4.20-0309-non-reasoning',
+    'xAI Grok 4.20 (non-reasoning)',
+  ),
+
+  /// Grok 4.20 multi-agent, where several agents research in parallel.
+  grok420MultiAgent(
+    'grok-4.20-multi-agent-0309',
```
cabljac, line 61: Go omits this model on purpose. Chat completions rejects it: "Multi Agent requests are not allowed on chat completions" (xai.go:234-237).

If that is still true, the Dev UI lists a model that answers 400 when selected. I have not confirmed this against the API. Do you agree? Or have you seen it answer through /v1/chat/completions?

packages/genkit_openai/test/xai_test.dart
```diff
@@ -0,0 +1,338 @@
...
+
+part 'xai_test.g.dart';
+
+/// A generated class, whose schema is a `$ref` into `$defs` before flattening.
+@Schema()
+abstract class $CityQuery {
+  String get city;
+}
+
+MockClient recordingClient(
```
cabljac, line 34: nit: `recordingClient` and `chatBodyOf` copy deepseek_test.dart:37 and :101, differing only in the canned strings. `$CityQuery` is generated three times across xai_test.dart:30, xai_integration_test.dart:37 and deepseek_integration_test.dart:41. test/fake_openai_server.dart already holds shared builders.

cabljac (changes requested):
One blocker. `flatten()` throws on a self-referential tool schema. A recursive tool schema that went out as `$ref`/`$defs` at #439 now fails on every provider, OpenAI included. The request never leaves the process. Details inline.

Two more items are fixes to #439 behaviour, and their tests land in deepseek_test.dart. The catalog also lists one model Go omits on purpose.

The xAI ids and context sizes match docs.x.ai.

PR #445, opened 2026-09-15
feat(genkit): tag ai.run and agent runTurn spans with action type

pavelgj:
Part 1 of 4 of the OpenTelemetry GenAI instrumentation stack (was #435).

Tags `ai.run(...)` steps and agent `runTurn` spans with an action type so
downstream instrumentation can classify them. `ai.run` steps get the new
`ActionType.flowStep`; agent turns get their existing type. No behavior change
beyond span metadata.

Foundational and independent: the `genkit_otel` package in later layers reads
`SpanMetadata.actionType`, but does not compile-depend on this change.

Stack:
1. **#445 span tagging (this PR)**
2. #446 genkit_otel instrumentation package
3. #447 ContentCapturingMode
4. #448 otel_jaeger sample + tooling


packages/genkit/lib/src/core/action.dart
```diff
@@ -74,6 +74,9 @@ extension type const ActionType(String value) {
   /// A utility action (e.g. the built-in `generate` action).
   static const ActionType util = ActionType('util');
 
+  /// An ad-hoc step inside a flow (e.g. `ai.run(name, fn)`).
+  static const ActionType flowStep = ActionType('flowStep');
```
MichaelDoyle, line 78: Something that's always bugged me. You can use `ai.run` outside of a flow, right?

e.g. I think we would recommend in middleware to create spans to do `ai.run(...)`?

If you want to standardize on this, let me know, and we can get dev-ui to understand general purpose genkit spans. Whatever name we choose.

MichaelDoyle (approved):

packages/genkit/lib/src/ai/agents/agent.dart
```diff
@@ -524,106 +524,109 @@ class SessionRunner<State> {
...
+            lastGoodState = session.getState();
+            lastGoodStateVersion = session.getVersion();
 
-          _notifyEndTurn(snapshotId, finishReason);
-          return false;
-        }, input: input);
+            _notifyEndTurn(snapshotId, finishReason);
+            return false;
+          },
+          actionType: ActionType.agent.value,
```
huangjeff5: If I `send('hi')` to `supportBot`, I get two spans. The root is the agent. Under it is `runTurn-1`. Next message in the same session is `runTurn-2`, still under that agent. One agent, many turns.

A flow is the same shape. The root is the flow. Under it are the `ai.run` steps. One flow, many steps.

This feels weird to me because this line stamps the turn with `ActionType.agent.value`, so after `send('hi')` both `supportBot` and `runTurn-1` come back `genkit:type = agent`. The description says agent turns get their existing type. That existing type is already on the root. The child reads like another agent, not like a turn under one.

I feel like the right experience should be: the root is the agent, each turn is the thing underneath, same as a flow and its steps. `flowStep` is the under-type we already have, so I would use that here for now. That is also what the other SDKs stamp on `runTurn-N`. If we want this child to be `agent` instead, we should change those at the same time, not only Dart.

packages/genkit/lib/src/core/action.dart
huangjeff5: I agree with Michael here.

`ai.run` is the helper we tell people to use for a named span. Inside a flow, `flowStep` reads fine: one flow, many steps. The same helper in middleware has no flow parent. `runTurn-1` sits under an agent, not a flow. One agent, many turns. Same shape as the flow case, different parent.

I feel like we need one under-type that works in all of those places: under a flow, under an agent, and on a standalone `ai.run`. `flowStep` is the string we have, and it sounds like "child of a flow," so it feels out of place anywhere the parent is not a flow.

This PR assigns `flowStep` to `ai.run` and `agent` to the turn. Those are the two stamps in the diff. The turn case is that awkward name showing up as a different type: we needed something under `agent`, and `flowStep` does not sound like it belongs there.

If we rename, I would want a word that means "child in a one-to-many," not "child of a flow":
- `step` (under a flow or an agent)
- `run` (the helper they type: `ai.run`)

Happy to keep `flowStep` until that rename. I would not use `agent` as the under-type. The root is already `agent`. The children need the shared under-name.

packages/genkit/lib/src/ai/agents/agent.dart
pavelgj: Agreed, switched the turn to `flowStep`. You're right that stamping it `agent` made the child read like a second agent under the agent root, instead of a turn under it. `flowStep` gives us the same root/child shape as a flow and its steps.

Confirmed this matches JS too: `runTurn-N` there goes through the `run` helper (ai/src/agent.ts), and `run` hardcodes `SPAN_TYPE_ATTR: 'flowStep'` (core/src/flow.ts). So Dart now emits the same type on turns. Dart only had a choice here because the agent calls `runInNewSpan` directly rather than routing through `ai.run`; passing `flowStep` lines it back up with JS.

packages/genkit/lib/src/core/action.dart
pavelgj: I agree on the name, but keeping `flowStep` for now as you suggested. Coordinating a rename is bit too much for my limited brain capacity at the moment. I don't think there's pressure at the moment, we can revisit.

PR #446, opened 2026-09-15
feat(genkit_otel): add OpenTelemetry GenAI instrumentation package

pavelgj:
Part 2 of 4 of the OpenTelemetry GenAI instrumentation stack (was #435).

Adds the new `genkit_otel` package: an `Instrumentation` provider that emits
OpenTelemetry telemetry following the GenAI semantic conventions (traces,
attributes, GenAI metrics, message mapping) on top of the
`dartastic_opentelemetry` SDK. Includes package scaffold, tests, workspace
wiring (root `pubspec.yaml`, README table row), and a `telemetry.dart` doc note.

Message-content capture in this layer is off; the `ContentCapturingMode` control
lands in #447.

Stack:
1. #445 span tagging
2. **#446 genkit_otel instrumentation package (this PR)**
3. #447 ContentCapturingMode
4. #448 otel_jaeger sample + tooling


packages/genkit_otel/lib/src/genai_instrumentation.dart
```diff
@@ -0,0 +1,510 @@
...
+    final split = splitModelName(metadata.name);
+    final provider = deriveProviderName(split.prefix);
+    final request = metadata.input is ModelRequest
+        ? metadata.input as ModelRequest
+        : null;
+
+    final attrs = <String, Object>{
+      GenAiAttr.operationName: GenAiOperation.chat,
+      GenAiAttr.requestModel: split.model,
+      GenAiAttr.providerName: ?provider,
```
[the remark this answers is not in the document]
pavelgj: This compiles and \`dart analyze\` passes clean. \`?provider\` is a null-aware map element (Dart 3.10+, and this package targets \`sdk: ^3.10.0\`): the entry is included only when \`provider != null\`, which is exactly equivalent to the suggested collection-\`if\`. Keeping the null-aware form. No change needed.

packages/genkit_otel/lib/src/genai_instrumentation.dart
```diff
@@ -0,0 +1,510 @@
...
+      kind: otel.SpanKind.client,
+      attributes: otel.OTel.attributesFromMap(attrs),
+    );
+    _maybeWarnNotRecording(span);
+
+    // Base metric attributes shared by both histograms: low cardinality only.
+    final metricAttrs = <String, Object>{
+      GenAiAttr.operationName: GenAiOperation.chat,
+      GenAiAttr.requestModel: split.model,
+      GenAiAttr.providerName: ?provider,
```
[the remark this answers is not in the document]
pavelgj: Same as above: \`?provider\` is a valid Dart 3.10+ null-aware map element (equivalent to the suggested collection-\`if\`), and \`dart analyze\` is clean. No change needed.

packages/genkit_otel/lib/src/genai_instrumentation.dart
```diff
@@ -0,0 +1,510 @@
...
+      _metrics.recordTokenUsage(
+        baseAttributes: baseAttrs,
+        inputTokens: usage.inputTokens?.toInt(),
+        outputTokens: usage.outputTokens?.toInt(),
+      );
+    }
+
+    _metrics.recordDuration(stopwatch.elapsedMicroseconds / 1e6, {
+      ...baseAttrs,
+      GenAiAttr.errorType: ?errorType,
```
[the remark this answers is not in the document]
pavelgj: Same as above: \`?errorType\` is a valid Dart 3.10+ null-aware map element, equivalent to the suggested collection-\`if\`. Analyzes clean. No change needed.

packages/genkit_otel/lib/src/genai_instrumentation.dart
```diff
@@ -0,0 +1,510 @@
...
+
+  @override
+  Future<O> runInNewSpan<O>(
+    SpanMetadata metadata,
+    Future<O> Function([SpanContext? span]) next,
+  ) {
+    switch (metadata.actionType) {
+      case 'model':
+        return _runModelSpan(metadata, next);
+      case 'tool' when emitToolSpans:
```
huangjeff5: The action type for tools is `tool.v2` so we need to branch against that, not `tool`.

`emitToolSpans: true` plus a real tool (`ActionType.tool` → `tool.v2`) falls through to a generic `{genkit.action.type: tool.v2}` span. Same leftover as the flag off. The test that pins `execute_tool` builds `actionType: 'tool'`, which no production site sends.

I suggest matching `tool.v2` so a real tool gets the `execute_tool` span when the flag is on.

packages/genkit_otel/lib/src/genai/gen_ai_attributes.dart
```diff
@@ -0,0 +1,202 @@
...
+    case 'length':
+      return 'length';
+    case 'blocked':
+      return 'content_filter';
+    case 'interrupted':
+      // No exact spec value; treat an interrupted turn as a normal stop.
+      return 'stop';
+    case 'other':
+    case 'unknown':
+    default:
```
huangjeff5, line 179: If a model returns `finishReason: failed` or `aborted` (no throw), this path always passes `failed: false`, those reasons hit `default`, and the chat span gets `gen_ai.response.finish_reasons = [stop]`. Anyone treating `stop` as success will count that turn as fine.

I suggest stamping `error` (or anything that is not `stop`) when the response itself said the turn failed.

MichaelDoyle:
A few high level items.

1. There are some missing things from tools that we might want to support
    - `gen_ai.tool.call.id`
    - `gen_ai.tool.description`
    -  When output collection is enabled: `gen_ai.tool.call.arguments` & `gen_ai.tool.call.result`
2. I think we should consistently collect `genkit.action.type`, `genkit.action.name`--even for model and tool spans. I think its too inconsistent not to, and its not 100% duplication. These have semantic meaning throughout the entire span tree.
3. Dynamic metadata added via `SpanContext.setMetadata()` uses colons: `genkit:metadata:<key>` (e.g., `genkit:metadata:context`, `genkit:metadata:agent:sessionId`. We should probably be consistent accross everything (all dots or all colons).
4. Possibly as a follow up, we should consider Agent spans. https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md

packages/genkit_otel/lib/src/genai/gen_ai_attributes.dart
MichaelDoyle: ➕

packages/genkit_otel/lib/src/genai_instrumentation.dart
MichaelDoyle: ➕ 

I was going to say `tool || tool.v2` but if its not possible then `tool.v2` is the right one.

pavelgj: Good catch. Matching to `ActionType.tool` instead, which is correct. There's no need match for `'tool'` because in dart it's not used (cannot be used).

packages/genkit_otel/lib/src/genai/gen_ai_attributes.dart
pavelgj: Good catch. Done.

pavelgj:
> A few high level items.
> 
> 1. There are some missing things from tools that we might want to support
>    
>    * `gen_ai.tool.call.id`
>    * `gen_ai.tool.description`
>    * When output collection is enabled: `gen_ai.tool.call.arguments` & `gen_ai.tool.call.result`
> 2. I think we should consistently collect `genkit.action.type`, `genkit.action.name`--even for model and tool spans. I think its too inconsistent not to, and its not 100% duplication. These have semantic meaning throughout the entire span tree.
> 3. Dynamic metadata added via `SpanContext.setMetadata()` uses colons: `genkit:metadata:<key>` (e.g., `genkit:metadata:context`, `genkit:metadata:agent:sessionId`. We should probably be consistent accross everything (all dots or all colons).
> 4. Possibly as a follow up, we should consider Agent spans. https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/gen-ai-agent-spans.md

1. added capturing of input/output, but id/description will need more work, deferring
2. done
3. done.
4. good idea, I thought about it, but definitely follow up.

huangjeff5 (approved):

PR #452, opened 2026-09-16
fix(genkit_google_genai): send seed in GenerationConfig

### LINKED ISSUE #363: fix(genkit_google_genai): seed accepted in config but silently dropped
`seed` (`model.dart:54`) is never mapped into `GenerationConfig` (`common_plugin.dart:260-294`). The binding already exists (`generated/generativelanguage.dart:583`), so the fix is one line in `toGeminiSettings`.

fileSearch split out to #451: it needs a new binding and a provisioned File Search store, which shouldn't hold up a one-line fix.

cabljac:
`seed` is declared on both `GeminiOptions` and `GeminiTtsOptions`, so a request config containing it passes schema validation, but neither `toGeminiSettings` nor `toGeminiTtsSettings` passed it to `gcl.GenerationConfig`. The value was validated and then silently dropped, and the request went out with no `seed` field.

Both mappers now pass `options.seed` straight through. The generated `GenerationConfig` binding already had the field and serialised it when non-null, so no regeneration was needed. `seed` is a plain nullable int and gets no emptiness handling, matching `temperature`, `topK` and the other scalar fields around it.

Two wire-level tests in `generation_config_wire_test.dart` drive a request through the plugin with a mock HTTP client and assert `generationConfig.seed` is on the captured body, one for a standard Gemini model and one for a TTS model. Both fail on `main` with `Actual: <null>` and pass with the fix. The full `genkit_google_genai` suite, `dart analyze` and `dart format` are clean.

Fixes #363


adesinah:
Live-tested this against the Gemini API, so the "not verified against the live API" caveat can go. Same seed twice gives byte-identical output, and unseeded runs at the same temperature differ, so the seed is doing the work rather than the prompt just being deterministic.

One trap worth recording for whoever tests this next: my first run failed and the fix was fine. I was generating against `gemini-flash-latest`, which is a moving alias and routed the two calls to different backends — indistinguishable from the seed being dropped. Pinning to `gemini-3.5-flash` it passes every time. A seed check also needs temperature above 0 plus an unseeded control, otherwise it passes on main too, where the field never reaches the wire.

Unrelated, in case you hit it on a live run: `should use tools` fails on this branch because the model answers `56,088` and the test looks for `56088`. Already fixed in #438 (6015e30), just not on main yet.

adesinah (approved):

PR #453, opened 2026-09-16
feat(genkit)!: simulate constrained generation when a model lacks native support

### LINKED ISSUE #433: Core: no constrained-generation fallback; supports.constrained is inert
Dart core has no equivalent of JS's `simulateConstrainedGeneration`, so `supports.constrained` changes no runtime behaviour.

**How JS does it**
- `js/ai/src/model.ts:373-384`: if a model's `supports.constrained` is absent or `'none'`, core pushes `simulateConstrainedGeneration()`. A third case, `'no-tools'`, simulates only when the request carries tools.
- `js/ai/src/model/middleware.ts:557-584`: that middleware injects the schema into the prompt as instructions, then resets `constrained` to false and clears `format`, `contentType` and `schema` (:573-577) before the plugin sees the request.
- JS's `jsonFormatter` sets `defaultInstructions: false` (`js/ai/src/formats/json.ts:25`), so it injects nothing on its own. The middleware is the entire mechanism.

**Dart today**
- `packages/genkit/lib/src/ai/middleware/` contains only `retry.dart`. `_resolveMiddleware` (`generate.dart:90-138`) installs no defaults; only explicit `use:` entries run.
- Nothing reads `supports.constrained` at runtime. A whole-repo sweep finds only writers (`model.dart:102-110`, `remote_model.dart:70-77`, `types.dart:185`, plus plugin declarations). The reflection API serialises it for Dev UI display (`reflection_v1.dart:162`), which is the only consumer.
- `formatters/json.dart:27` sets `constrained: true` unconditionally and nothing downstream strips it. A model declaring no `constrained`, or `constrained: false`, still receives `output.constrained == true`, the schema, and zero instruction parts.
- Declarations are inconsistent, since nothing exercises them: google_genai, vertexai and anthropic declare `constrained`; openai (`utils.dart:48-53`), chrome (`chrome_model.dart:26-31`) and firebase_ai declare nothing. Core's own defaults set it true (`model.dart:108`, `remote_model.dart:76`).

**Impact**
- No plugin has a fallback of any kind. openai (`chat.dart:79-86`) and firebase_ai (`genkit_firebase_ai.dart:742`) send the native schema whenever `output.schema` is set and never read `output.constrained`. google_genai does read it (`common_plugin.dart:81`, `:99`). chrome ignores `output` entirely and returns prose.
- So a caller passing `constrained: false` has no effect on most plugins, and a model without native support has no degraded path. Failure mode varies: a provider rejection for the hosted plugins, a client-side parse failure for chrome. Neither is tested.
- `genkit_anthropic` carries the only fallback, a forced `return_output` tool, and #402 gates it behind `structuredOutputs` such that no model reaches it.

**Suggested shape**
- Port the JS pair: a `simulateConstrainedGeneration` middleware plus the `supports.constrained` check that installs it.
- The port must strip `schema`, not just `constrained`, since openai and firebase_ai key off `schema` alone.
- Audit declared `supports.constrained` across plugins and core's own defaults in the same change, since today's values were never exercised.

**Out of scope**
- The JSON parser Adesina mentioned from Go is a separate ticket.

adesinah:
`supports.constrained` was inert. Every plugin wrote it, nothing read it, and
`formatters/json.dart` set `constrained: true` on every schema request anyway.
So a model with no native constrained generation received the flag, the schema
and zero instructions — a provider rejection for the hosted plugins, a
client-side parse failure for chrome, and no degraded path anywhere.

Ports the JS pair: `simulateConstrainedGeneration` plus the
`supports.constrained` check that installs it.

- `SimulateConstrainedGenerationMiddleware` injects the schema into the
  conversation as instructions and clears `constrained` and `schema` before the
  plugin sees the request. Clearing `schema` is the point — openai and
  firebase_ai key off it and never read the flag.
- The strip and the injection are one decision. Instructions attach to the last
  system or user message, and if the schema cannot be placed there the request
  goes through untouched rather than losing both. It also appends alongside a
  caller's own `outputInstructions` instead of being suppressed by them, which
  is why it does not route through `injectInstructions`.
- `format` and `contentType` are kept, unlike JS. Plugins read that pair to
  enable native JSON mode, which guarantees the response parses and is
  independent of schema constraint.
- `needsConstrainedSimulation` is the JS predicate: absent, `false` and
  `'none'` simulate, `'no-tools'` simulates only with tools in the request.
  Absent means simulate — wrongly simulating costs a longer prompt, wrongly
  not simulating costs a rejection.
- Wired into `Model` itself, not just the `generate` loop, so a direct action
  call — `registry.lookupAction`, the Dev UI's `runAction` — gets it too, not
  only calls that go through `generate`. Deciding inside the model call also
  means a middleware that adds the schema or the tools on the way down is not
  skipped. Parsing is untouched.
- `outputNoInstructions` is not honoured on the simulated path — the flag lives
  on a type `generate` has already discarded by the time this middleware sees
  the request, and threading it through means changing a wire type JS and Go
  share too. JS has the identical gap, so this is parity, not a new hole;
  documented in `simulate_constrained_generation.dart`.

Marked breaking (`feat(genkit)!`): `modelMetadata`'s default for an undeclared model used to imply `constrained: true`; it now withholds the claim. Every in-repo caller passes `modelInfo` explicitly, so the blast radius is third-party plugins and `remoteModel`, which lose native schema support silently unless they add `info`.

Declaring `constrained` now decides behaviour, so the declarations are audited
in the same change. Core's undeclared-model defaults claimed it and no longer
do; `genkit_firebase_ai` declared nothing and now declares the Gemini
capabilities it has, with its native `responseSchema` gated on `constrained`
like google_genai's. chrome and openai were already right. `defineModel` gains
`ModelInfo? info`, since a model defined through it could not declare anything.

`genkit_anthropic` keeps its uncurated fallback on `baseClaudeSupports`: the
forced `return_output` tool is the only native path it has, and not every
Claude name accepts a forced `tool_choice` (`claude-fable-5-1` answers it with
a 400). Curation is the only signal for which names do, so an uncurated name
makes no claim and core simulates. Curated models claim `'no-tools'` rather
than `true`: that tool pins `tool_choice`, so a request carrying the caller's
own tools could never reach them and the tool loop would silently never fire.
With tools, the schema goes through the prompt instead and `tool_choice` stays
free; without them, the native path applies and outranks a caller `toolChoice`,
which would otherwise unpin the tool on a request core did not simulate for.
That override is gated on `constrained` as well as the schema: a request that
explicitly opted out keeps its `toolChoice`, since nothing was stripped from it
and the tool is offered rather than forced. Both pinned by wire tests. Narrowing further, if
`structuredOutputs` ever needs to mean something more specific than "the tool
trick works," is for #402.

Instructions attach to the *first* system message (last user message when
there is none), matching JS. A provider that folds the conversation's system
messages into one field can keep only the first - genkit_anthropic does - so
instructions on a later one vanish after the schema has already been stripped.
`injectInstructions` in `formatters.dart` had the same divergence and is
aligned in the same change.

The audit also turned up #462: `gpt-4` and `gpt-4-32k` are sent a
`response_format` their snapshots reject. Pre-existing, unchanged by this PR —
they 400 on `json_schema` today and on `json_object` after it — and the fix
needs a call about compat hosts that belongs with #415. A related but distinct
gap, #466: an unrecognised model name on a non-OpenAI compat host inherits
OpenAI's own optimistic "assume the latest capabilities" default, including
`constrained: true`, which doesn't hold for an arbitrary third-party backend.

`dart analyze` clean; genkit 732, openai 175, google_genai 69, anthropic 107,
vertexai 26, middleware 53 pass. `genkit_firebase_ai` needs a newer Flutter SDK
than the one installed in this dev environment, so its tests (including a new
one pinning `firebaseAiModelInfo`'s `constrained` claim) were verified by
review rather than run locally, and are left to CI.

`genkit_anthropic/test/test_live.dart` gains live coverage of the simulated
path against an uncurated model, buffered and streamed, asserting the name is
genuinely uncurated first so it cannot quietly start exercising the native
path. The suite also skips with a clear message when `ANTHROPIC_API_KEY` is
unset, instead of failing every test on an empty response.

Closes: #433


packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
```diff
@@ -0,0 +1,102 @@
...
+    );
+
+    return next(
+      ModelRequest(
+        messages: messages,
+        config: request.config,
+        tools: request.tools,
+        toolChoice: request.toolChoice,
+        docs: request.docs,
+        output: OutputConfig(constrained: false),
```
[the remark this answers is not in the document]
adesinah: Good catch, taken in 01e26e9. Checked it against the plugins first: `genkit_google_genai` builds `isJsonMode` from exactly `output.format == 'json' || output.contentType == 'application/json'` (common_plugin.dart:68-70), and `genkit_openai` sends `json_object` off the same signal since #415, so clearing both really did cost native JSON mode on the simulated path.

Worth saying why it's right rather than just smaller: JSON mode is orthogonal to schema constraint. It guarantees the response parses, which leaves the injected instructions responsible only for the shape. And the instructions name JSON, which is the precondition OpenAI puts on `json_object`.

This does part company with JS, which clears all four in middleware.ts:573-577, so I've said so in the doc comment and pinned it with a test rather than leaving it to be re-derived later.

`schema` still goes, for the reason already in the comment.

cabljac (commented):
Two worth settling before merge: the middleware is installed in the generate loop rather than on the model action, so a direct model-action call still reaches the plugin with `constrained: true` and the schema intact; and `outputNoInstructions: true` no longer suppresses anything on the simulated path. The firebase_ai gate is fine, its dartdoc reason is not. Rest inline.

One more with no diff line to hang it on, `genkit_anthropic/lib/src/plugin_impl.dart:247`. Reproduced at the wire with a MockClient: `claude-3-5-sonnet-20241022` on `main` sends `TOOL_CHOICE: {type: tool, name: return_output}`; on this branch it sends `TOOLS: null` and the schema block in the prompt, parsed back out by core's `extractJson` (`formatters/json.dart:39`). The PR body defers this to #402, so the question is whether the deferral is right, not whether it was noticed. Every Claude name outside `KnownClaudeModel` loses forced structured output the moment this merges. Declare `constrained` on `baseClaudeSupports` now, or land #402 first?

Analyze and every suite are green at this ref, including `flutter test` in genkit_firebase_ai (23 pass, Flutter 3.47.2), so the PR body's note about needing a newer SDK can come out.

Needs a decision: `feat(genkit)!` on the squash title?

packages/genkit/lib/src/ai/generate.dart
```diff
@@ -512,7 +524,19 @@ Future<GenerateResponseHelper> _runGenerateLoop(
     );
   }
 
-  final composedModel = resolvedMiddleware.reversed.fold(
+  // Appended rather than prepended, so it is the last hop before the model:
+  // caller middleware observes the request the caller actually made, schema
+  // and all, and only the plugin sees the stripped one. Always installed, and
+  // it decides for itself once it sees the request — middleware ahead of it
+  // can add the schema or the tools the decision turns on.
+  final modelMiddleware = [
```
cabljac: Installed here, so only `ai.generate` gets it. I resolved the model action and called it directly with `constrained: true` plus a schema against a model with no `info`, and the plugin received both intact with no instructions. Dev UI `runAction` takes the same route (`reflection_v1.dart:133`), though I did not stand one up. JS attaches the equivalent at define time (`js/ai/src/model.ts:366-387`, wired in at `:192`), so a direct call is covered there, and Go rejects it outright: `model %q does not support native constrained output, but constrained output was requested. Request: %+v` (`go/ai/model_middleware.go:267`). Attaching at `Model` construction, or porting Go's `validateSupport`, closes it.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
```diff
@@ -0,0 +1,176 @@
...
+    ActionFnArg<ModelResponseChunk, ModelRequest, void> ctx,
+    Future<ModelResponse> Function(
+      ModelRequest request,
+      ActionFnArg<ModelResponseChunk, ModelRequest, void> ctx,
+    )
+    next,
+  ) {
+    final output = request.output;
+    final schema = output?.schema;
+    if (output?.constrained != true || schema == null) {
```
cabljac, line 91: `outputNoInstructions: true` with an `outputSchema`, on a model that does not declare `constrained`, now produces the full "Output should be in JSON format..." block. On `main` it produced none. The suppression lives on `GenerateActionOutputConfig.instructions` and is dropped when `generate.dart:493-506` builds the `OutputConfig` the middleware sees, which has no `instructions` field. JS has the same gap (`js/ai/src/model/middleware.ts:557-583`), and suppression still works on models that declare `constrained`, so it holds on the native path and is dropped on the simulated one.

Two ways out: thread `instructions` through to `OutputConfig` so the middleware can honour it, or accept JS parity and pin it with a test. `formats_test.dart:302` passes no schema, so it short-circuits and covers neither.

packages/genkit_firebase_ai/lib/genkit_firebase_ai.dart
```diff
@@ -719,12 +745,24 @@ fai.Schema _toGeminiSchemaInternal(Map<String, dynamic> json) {
   }
 }
 
+/// Maps Genkit's request onto Firebase AI's generation config.
```
cabljac, line 752: The gate is pinned by the test diff, so the behaviour is intended. It is the dartdoc that needs correcting. `:752` says core has already injected those instructions, and it has not: `jsonFormatter` sets `defaultInstructions: false` (`formatters/json.dart:28`) and the middleware returns early unless `constrained == true` (`simulate_constrained_generation.dart:78`), so `generate(outputSchema: X, outputConstrained: false)` arrives here with the schema set and zero `purpose: 'output'` parts, and the gate then drops `responseSchema` as well, leaving nothing describing the shape.

Narrow scope: an ordinary `generate(outputSchema: X)` still carries `constrained: true` from `applyFormat`. `genkit_google_genai/lib/src/common_plugin.dart:292` and `:332` already do the same. Does that match your reading?

packages/genkit_openai/lib/src/known_models.dart
```diff
@@ -23,10 +23,20 @@ import 'utils.dart';
 // them advertise no constrained generation. That is the only difference
 // between the `*Supports` presets and their `*LegacySupports` counterparts.
 //
-// Descriptive, not load-bearing: nothing in `package:genkit` reads `supports`,
-// so this changes what the Dev UI and `listActions` report, not what the
-// generate path does. The plugin sends `response_format` off the request's own
-// output config either way.
+// `constrained` is load-bearing since #433: `generate` simulates constrained
```
cabljac, line 26: Adjacent to this comment rather than in it. `dynamicModelInfo` (`:548-560`) falls through to `multimodalSupports`, which declares `'constrained': true`, and `compatModelInfo` (`:589-592`) reuses it via `openai_plugin.dart:260`. On `openAI(baseUrl: ...)`, a name `getModelType` cannot classify (`mistral-large`, `deepseek-chat`) keeps `constrained: true`, skips simulation, and `buildOpenAIResponseFormat` (`chat.dart:94-120`) emits `json_schema`, which a host implementing only `json_object` rejects. Names containing `instruct`, `davinci` or `babbage` are classified `completion` and are simulated, so the gap is narrower than it first looks.

Pre-dates the PR, but `_unclaimedModelInfo` makes the default load-bearing, so the same argument applies here: drop `constrained` from the dynamic fallback, or file a follow-up if that is wider than you want in this PR.

packages/genkit/lib/src/ai/model.dart
```diff
@@ -83,6 +83,29 @@ class Model<CustomOptions>
   }
 }
 
+/// Capability metadata for a model that supplied none.
```
cabljac, line 154: `modelMetadata`'s default drops `constrained`, so `listActions` output changes for every undeclared model and any third-party plugin building a `Model` without the flag silently stops receiving `output.schema`. `remoteModel` loses its own fallback (`remote_model.dart:64-70`). Every in-repo caller passes `modelInfo` explicitly, so the blast radius is third-party plugins and `remoteModel`. 426ced1 is `fix(genkit_openai)!: send json_object for schemaless JSON output (#415)` for a narrower change, so `feat(genkit)!:` on the squash title looks right.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
```diff
@@ -0,0 +1,176 @@
...
+import '../generate_middleware.dart';
+
+/// Renders [schema] as the instructions a model is given in place of native
+/// constrained generation.
+///
+/// Deliberately not `jsonFormatter`'s wording: that formatter sets
+/// `defaultInstructions: false`, so it never injects anything of its own and
+/// this middleware is the whole mechanism. Keeping the string here means the
+/// simulated path reads the same whatever format asked for a schema.
+String simulatedConstrainedInstructions(Map<String, dynamic> schema) {
```
cabljac, line 32: nit: the dartdoc at `:25` says "Deliberately not `jsonFormatter`'s wording", but the two strings are byte-identical, 254 bytes (`formatters/json.dart:33-35`). `carriesInstructions` dedups on exact `p.text == instructions` (`:133`), so that equality is what prevents double injection on the `GenerateActionOptions(instructions: bool(true))` path, which `formats_test.dart:174` already exercises. Changing `should` to `MUST` in one literal fails exactly one test, on a `contains` assertion at `formats_test.dart:181`; prefixing the string instead leaves all 729 green while emitting two instruction parts. A shared constant, or matching on `metadata['purpose'] == 'output'`, would pin it properly.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
```diff
@@ -0,0 +1,176 @@
...
+/// instructions this middleware injected on an earlier turn from instructions
+/// the caller wrote themselves: the first must not be duplicated, the second
+/// does not describe the schema and must not suppress it. Matching on the
+/// rendered text distinguishes them, so a caller's own `outputInstructions`
+/// now sit alongside the schema rather than silently replacing it.
+///
+/// Returns null when there is no system or user message to append to, which
+/// the caller reads as "cannot simulate".
+List<Message>? _withInstructions(List<Message> messages, String instructions) {
+  bool carriesInstructions(Message m) =>
```
cabljac, line 151: nit: this scans every message regardless of role. An echo of the instruction text on a model or tool turn suppresses injection while `schema` and `constrained` are still stripped, routing around the guard at `:90`. Less severe than it sounds, since the matched text is the schema render itself, so the model still sees the shape, just attributed to a model turn. Restricting the scan to system and user messages costs nothing.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
```diff
@@ -0,0 +1,176 @@
...
+      m.content.any((p) => p.isText && p.text == instructions);
+  if (messages.any(carriesInstructions)) return messages;
+
+  var targetIndex = messages.lastIndexWhere((m) => m.role == Role.system);
```
cabljac: nit: `lastIndexWhere` picks the last system message, matching Dart's own `formatters/formatters.dart:174` but not JS (`js/ai/src/formats/index.ts:96`, `findIndex`) or Go (`go/ai/format.go:146-151`, loop with `break`), which take the first. All three fall back to the last user message, so this only shows with more than one system message. Might be worth fixing in one place rather than two.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
```diff
@@ -0,0 +1,176 @@
...
+  if (targetIndex < 0) return null;
+
+  final target = messages[targetIndex];
+  return [
+    ...messages.sublist(0, targetIndex),
+    Message(
+      role: target.role,
+      content: [
+        ...target.content,
+        TextPart(text: instructions, metadata: {'purpose': 'output'}),
```
cabljac, line 174: nit: `injectInstructions` (`formatters/formatters.dart:184-195`) swaps an existing `{purpose:'output', pending:true}` placeholder in place; this appends and leaves it standing. Latent, nothing in the repo writes that metadata today, and the one upstream producer (`dotprompt`'s `{{section}}`) lands as a `DataPart`, which the reader's `isText` check excludes. JS is not a clean reference here either: `js/ai/src/formats/index.ts:109` uses `partIndex > 0`, so a placeholder at index 0 is appended after rather than replaced. Mirroring the placeholder swap keeps the two Dart injectors in step.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
```diff
@@ -0,0 +1,176 @@
...
+/// from exactly them, and `genkit_openai` sends `json_object` off the same
+/// signal — and that mode is orthogonal to schema constraint: it guarantees the
+/// response parses, leaving the injected instructions to supply only the shape.
+/// Clearing them would throw that away and make the simulated path parse worse
+/// than it needs to. The instructions name JSON, which is the precondition
+/// OpenAI puts on `json_object`.
+///
+/// Parsing is unaffected: `generate` parses the response against the format
+/// it resolved before middleware ran, so the caller still gets typed output.
+class SimulateConstrainedGenerationMiddleware extends GenerateMiddleware {
```
cabljac, line 71: nit: `SimulateConstrainedGenerationMiddleware` is not exported from `genkit.dart` the way `retry.dart` is (line 83), so the test imports it through `package:genkit/src/...`. JS also exposes an `instructionsRenderer` hook (`js/ai/src/model/middleware.ts:539-542`) for providers that want a different preamble. Intentional?

packages/genkit_firebase_ai/lib/genkit_firebase_ai.dart
```diff
@@ -195,6 +219,7 @@ class _FirebaseGenAiPlugin extends GenkitPlugin {
   Model _createModel(String modelName) {
     return Model(
       name: 'firebaseai/$modelName',
+      metadata: {'model': firebaseAiModelInfo.toJson()},
```
cabljac, line 226: nit: this is what `generate.dart:97-101` reads `constrained` out of, so it keeps firebase_ai on the native path, and deleting it leaves all 23 tests and `flutter analyze` green. `firebaseAiModelInfo` has no other reference, so the line going orphans the whole constant silently. Something like `expect(plugin.resolve(ActionType.model, 'gemini-...')!.metadata['model']['supports']['constrained'], isTrue)` would pin it.

packages/genkit_anthropic/test/test_live.dart
```diff
@@ -38,6 +39,15 @@ void main() {
   // Check if API key is available
   final apiKey = Platform.environment['ANTHROPIC_API_KEY'];
 
+  // Without this the suite still runs, and every call comes back as an empty
```
cabljac: nit: the early return leaves no registered tests, so `dart test test/test_live.dart` without `ANTHROPIC_API_KEY` exits 79 with "No tests ran". `group(..., skip: apiKey == null ? 'ANTHROPIC_API_KEY is not set' : null)` reports skips instead. Not a CI issue, melos filters on `dirExists: test` and the default glob never loads the file.

packages/genkit/lib/src/ai/generate.dart
adesinah: Good catch, you're right that a direct action call skipped this entirely. Fixed in d3c2622 by wrapping `fn` inside `Model`'s own constructor instead of only in `generate`'s middleware chain, so `registry.lookupAction` / Dev UI's `runAction` get the simulation too. `generate` no longer installs its own copy, it just relies on the model call at the bottom of the chain.

Added a test that calls the action directly (bypassing `generate`) and confirms the schema still gets simulated - it fails without the fix.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
adesinah: Fixed in f3909ec, scoped the match to system/user messages.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
adesinah: You're right, they're byte-identical. Fixed the comment in f3909ec to say what's actually true: it matches the wording on purpose (that's what lets the dedupe key off exact text), it's just not a call into jsonFormatter.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
adesinah: Went with the second option. Threading `instructions` through `OutputConfig` means changing a wire type JS and Go share too, that felt like a bigger call than this PR should make on its own. Documented the gap in f3909ec instead (JS parity, cited). Didn't add the pinning test you mentioned since formats_test.dart's existing one already doesn't exercise this path and I didn't want to touch that test's intent - happy to add a dedicated one if you'd rather have it pinned than just documented.

packages/genkit_firebase_ai/lib/genkit_firebase_ai.dart
adesinah: Right, core doesn't inject anything for `constrained: false` either (jsonFormatter's defaultInstructions: false, and the middleware only fires when constrained is true). Fixed the dartdoc in 91a8aba to say that instead.

packages/genkit_firebase_ai/lib/genkit_firebase_ai.dart
adesinah: Added in f87a6d0. Couldn't run it here (this environment's Flutter is 3.0.5, package needs 3.12+), but it's a one-line expect on the constant so I'm confident in it - let me know if it doesn't pass on your end.

packages/genkit_anthropic/test/test_live.dart
adesinah: Fixed in 166f8ee, group-level skip instead of the early return.

packages/genkit_openai/lib/src/known_models.dart
adesinah: That's a real gap and a separate one from #462 - filed #466 and pointed to it from the comment in 2324822. Didn't change dynamicModelInfo/compatModelInfo's actual behavior here since the optimistic default is deliberate and correct for real OpenAI (per the dartdoc above it), the question is only whether compatModelInfo should inherit it, and that's a design call for that function, not something #433 forces.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
adesinah: Leaving this alone - it matches formatters.dart's existing lastIndexWhere, and fixing the last-vs-first divergence from JS/Go means touching both places at once, which is a formatters.dart question more than a this-PR one. Only shows up with more than one system message anyway.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
adesinah: Leaving this too, since as you note nothing produces that metadata today - didn't want to write handling for a case that can't happen yet. If a producer shows up this needs reconciling with injectInstructions' swap-in-place behavior then, agreed.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
adesinah: Intentional, not exporting it. retry.dart's RetryMiddleware is opt-in, you attach it yourself. This one isn't meant to be attached by hand anymore - it's wired into every Model at construction now, so exporting it as a use: option would be misleading (adding it again would just be a no-op on top of itself). A swappable instructionsRenderer like JS's is a real idea but a separate feature, didn't want to speculate it into this PR.

adesinah:
Covered the rest of your review body:

- Anthropic (`plugin_impl.dart:247`): went with declaring `constrained` now rather than waiting on #402. `commonModelInfo` claims `structuredClaudeSupports` instead of `baseClaudeSupports` (245d410) - the forced `return_output` tool is ordinary tool calling gated on `output.schema` alone, nothing in the request builder actually checks curation or Anthropic's Structured Outputs list, so withholding the claim for uncurated names was strictly worse than `main`. Updated the two live tests that assumed an uncurated model would need simulation - none do anymore, so replaced them with one live check that the catalog-wide claim holds, and pointed to the unit suite for actual middleware coverage.
- Squash title: added the `!` - `feat(genkit)!: ...` - and a short paragraph in the body on what breaks (`modelMetadata`'s undeclared-model default).
- Flutter SDK note: reworded rather than dropped, since I still can't run genkit_firebase_ai here (3.0.5 vs the 3.12+ it needs) - now says so plainly instead of implying it's untested everywhere.

Let me know if anything above still needs another pass.

adesinah:
Ran a review pass over the branch; two real bugs came out of it, both mine.

**`remoteModel` ignored its own `constrained` declaration** (c910613). When I moved the simulation into `Model`, I read `supports.constrained` in the constructor's initializer list — but `defineRemoteModel` builds the `Model` with no metadata and fills it by cascade afterwards, so the wrapper captured `null` and simulated unconditionally. `defineRemoteModel(..., modelInfo: ModelInfo(supports: {'constrained': true}))` advertised native support and then had its schema stripped on every call. The existing test only asserted the metadata, never what the remote received, which is how it got through. `Model` now redirects through a private constructor so one map instance backs both the field and the lookup, and the claim is read per call — that also fixes it for any third-party `Model(...)..metadata[...] = ...`. Added a test asserting the wire body.

**`constrained: true` on Claude broke the tool loop** (c35222f). The forced `return_output` tool pins `tool_choice`, so `generate(tools: [...], outputSchema: ...)` left the caller's tools on the wire but unreachable — Claude could only answer with `return_output`. `true` skips simulation regardless of tools; `'no-tools'` is the value for exactly this, and it's what the tier declares now. Native path when nothing competes, prompt instructions when the caller brought tools. Wire test both ways. Thanks for pushing on declaring this now rather than deferring — deferring would have shipped the narrower `true` and this would have been much harder to spot later.

Not changing, want your read: the same argument was raised for `firebaseAiModelInfo`, that `responseSchema` + `tools` is rejected by Gemini and it should be `'no-tools'` too. I can't verify that here — unlike the Claude case the breakage would be in Google's API rather than our own code, `genkit_google_genai` carries the identical `constrained: true` pre-existing, and changing only firebase_ai would split the two on an unproven claim. If you know that combination 400s, it's a real bug in both and worth its own issue rather than a quiet change in this PR.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -29,7 +29,17 @@ final _logger = Logger('genkit_anthropic');
...
+/// Claims [structuredClaudeSupports], not [baseClaudeSupports]: the structured
+/// output this plugin sends today is the forced `return_output` tool below
+/// (`:252-262`), gated on `output.schema` alone, not on a model's presence in
+/// [KnownClaudeModel] or Anthropic's Structured Outputs list. Withholding
+/// `constrained` here would have core simulate for every uncurated name -
+/// every Claude model but the ten this plugin curates - trading a tool call
+/// this plugin already knows how to make for prompt instructions. #402 adds a
+/// second, more selective mechanism; this claim may need to narrow once that
+/// lands.
+final commonModelInfo = ModelInfo(supports: structuredClaudeSupports);
```
cabljac: This assumption does not hold for every Claude name. Checked against the live API:

```
claude-fable-5-1   tool_choice {"type":"tool"}  -> 400  tool_choice: type "tool" and "any" are not supported for this model.
claude-fable-5-1   tool_choice {"type":"any"}   -> 400  same
claude-fable-5     -> 200 tool_use
claude-sonnet-4-5  -> 200 tool_use
claude-opus-5      -> 200 tool_use
claude-opus-4-8    -> 200 tool_use
claude-haiku-4-5   -> 200 tool_use
```

`claude-fable-5` is in `KnownClaudeModel`, `claude-fable-5-1` is not, so on the previous head it took `baseClaudeSupports`, was simulated, and worked. With `commonModelInfo` claiming `'no-tools'`, a schema-only request to it now forces `return_output` and gets a 400. Every curated model is fine, so this is confined to uncurated names on models that refuse forced tools.

Separately, `needsConstrainedSimulation` keys `'no-tools'` on `request.tools` alone (`simulate_constrained_generation.dart:94-99`), while `:272` here forces `tool_choice` and `:275-282` lets a caller `toolChoice` overwrite it. A schema-only request with `toolChoice: none` is therefore not simulated yet loses the forced choice, leaving the model a tool it may not call and no textual schema. With `auto` it can still choose the tool, so that case is weaker rather than broken.

packages/genkit_anthropic/lib/src/known_models.dart
```diff
@@ -31,13 +31,24 @@ const baseClaudeSupports = <String, dynamic>{
   'output': ['text'],
```
cabljac, line 35: nit: `baseClaudeSupports` is now unreachable from `lib/`. No `KnownClaudeModel` entry sets `structuredOutputs: false` and `commonModelInfo` no longer uses it, so only tests and dartdoc reference it. Worth dropping, or saying it stands as the tier for a future curated model that cannot do the tool trick.

cabljac (commented):
Re-reviewed at `c35222fe`. Attach point, dedup scoping, the firebase_ai dartdoc and the `!` on the title are all addressed, and moving the read of `supports.constrained` to call time rather than construction is a better fix than the one I suggested.

The `'no-tools'` change also fixes a bug that was live before this PR: the old head sent `tools=[lookup, return_output]` with `tool_choice` pinned to `return_output`, so a caller's own tool could never fire.

One new problem from the widening, inline.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
```diff
@@ -0,0 +1,195 @@
...
+  // schema text (in a transcript, a tool result, ...) without it having come
+  // from this middleware, and matching that would suppress injection while
+  // still stripping `schema`/`constrained` below - the model would end up
+  // with neither the native path nor the instructions.
+  bool carriesInstructions(Message m) =>
+      (m.role == Role.system || m.role == Role.user) &&
+      m.content.any((p) => p.isText && p.text == instructions);
+  if (messages.any(carriesInstructions)) return messages;
+
+  var targetIndex = messages.lastIndexWhere((m) => m.role == Role.system);
```
cabljac: i'm wondering if this can cause schema loss - with two system messages, caller tools and output schema , simulation appends instructions to the last system message and strips output.schema 

Anthropic's _buildCreateRequest forwards only the first system message and filters others out.

I think changing lastIndexWhere to indexWhere matches both upstream injectors and fixes the case i repro'd. Preserving all system messages, as Go does, would address Dart’s broader conversion issue..

cabljac (changes requested):

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Good catch, reverted the fallback to `baseClaudeSupports`. Curation is the only signal for which names take a forced `tool_choice`, so an uncurated name makes no claim and core simulates - a longer prompt instead of a 400.

On the `toolChoice` point: the forced `return_output` choice now outranks the caller's `toolChoice`, so a schema request can no longer lose it. Pinned by a wire test.

packages/genkit_anthropic/lib/src/known_models.dart
adesinah: Reachable again now that `commonModelInfo` is back on it. Doc says what it is: the tier for a name we can't vouch for.

packages/genkit/lib/src/ai/middleware/simulate_constrained_generation.dart
adesinah: You're right, switched to `indexWhere` for the system message (last user message unchanged), which matches JS. Added a two-system-message test. `injectInstructions` in `formatters.dart` had the same divergence, so I aligned it too.

packages/genkit_anthropic/lib/src/plugin_impl.dart
```diff
@@ -262,7 +270,12 @@ class AnthropicPluginImpl extends GenkitPlugin {
       toolChoice = sdk.ToolChoice.tool(toolName);
     }
 
-    if (req.toolChoice != null) {
+    // The forced `return_output` tool is the whole mechanism behind the
+    // `constrained` claim, so it outranks the caller's `toolChoice`: core only
+    // routes a schema request here when the model claims native support, and
+    // honouring a `none` here would leave the model neither the tool nor the
+    // instructions core stripped.
+    if (req.toolChoice != null && req.output?.schema == null) {
```
cabljac: This fixes the constrained case I raised. One edge case: the override also applies with explicit `outputConstrained: false`, where simulation is skipped. With a schema, a `lookup` tool and `toolChoice: 'lookup'`, the outgoing request now forces `return_output`. Could we preserve the caller’s choice when constraint is explicitly disabled?

cabljac (approved):
approved to merge once one remaining comment is addresssed

packages/genkit_anthropic/lib/src/plugin_impl.dart
adesinah: Good catch. The override is now gated on `output.constrained == true` as well as the schema: opting out of constrained output opts out of the mechanism behind it, and core stripped nothing, so there is no reason to overrule you. `return_output` is still offered in that case, just not forced. Test added.

Issue #456, opened 2026-09-17
feat(genkit): carry and honour Retry-After from provider errors


cabljac:
Nothing in `packages/` reads `Retry-After` - zero hits repo-wide. `RetryMiddleware._calculateDelay` (`packages/genkit/lib/src/ai/middleware/retry.dart:236-240`) is pure exponential backoff with jitter, so a server-suggested delay is ignored even when sent.

JS splits this: core carries the metadata, the plugin parses the header. `ErrorResponseMetadata` (`js/core/src/error.ts:38-44`) holds `retryAfterMs` and raw `headers`, deliberately not serialized to the wire - in-process only, for middleware to make retry decisions. The Anthropic plugin parses both delay-seconds and HTTP-date forms per RFC 7231 section 7.1.3 (`js/plugins/anthropic/src/models.ts:48-58,210-215`).

Go is narrower and different: `go/plugins/googlegenai/errors.go:49` puts `retryAfterMs` in the error's details map with a typed `RetryDelay(err)` accessor, reading `google.rpc.RetryInfo` from 429s rather than a header. Dart's `GenkitException.details` is a `String?`, so that shape does not transfer.

Scope:

1. A response-metadata carrier on `GenkitException` for `retryAfterMs` and raw headers, not serialized.
2. `genkit_anthropic` parses `Retry-After` and populates it. The SDK already exposes headers: `ApiException.responseMetadata.headers` is a `Map<String, String>` (`anthropic_sdk_dart-8.0.0/lib/src/errors/exceptions.dart:38,98`).
3. `RetryMiddleware` uses the suggested delay when present, falling back to backoff.

One issue rather than three: items 1 and 2 change no observable behaviour alone, and item 3 is what makes them testable.

Depends on #355 - a 529 must be retryable before the delay matters.

PR #459, opened 2026-09-17
fix(genkit_anthropic): map 529 overloaded to UNAVAILABLE

### LINKED ISSUE #355: fix(genkit_anthropic): 529 (overloaded) maps to UNKNOWN, so overload errors are never retried
`StatusCodes.fromHttpStatus` has no 529 case (`packages/genkit/lib/src/exception.dart:129-158`), so it falls through to `default: return StatusCodes.UNKNOWN`. The Anthropic plugin maps through it at `packages/genkit_anthropic/lib/src/plugin_impl.dart:200`.

The effect is not cosmetic. `RetryMiddleware.defaultRetryStatuses` (`packages/genkit/lib/src/ai/middleware/retry.dart:132-138`) is UNAVAILABLE, DEADLINE_EXCEEDED, RESOURCE_EXHAUSTED, ABORTED, INTERNAL. UNKNOWN is not in it, so an Anthropic 529 overload is never retried.

JS maps 529 alongside 503 to UNAVAILABLE, and does it in the plugin rather than core (`js/plugins/anthropic/src/models.ts:205-207`). 529 is Anthropic-specific, so core's canonical mapping is the wrong place - it would apply to every plugin.

Fix: map 529 to UNAVAILABLE in `genkit_anthropic`. Test: a mocked 529 is retried by the default retry middleware.

Retry-After parsing split out to #456.

cabljac:
Anthropic returns HTTP 529 when the API is overloaded. Core's canonical HTTP mapping, `StatusCodes.fromHttpStatus`, has no 529 case, so the plugin surfaced the error as `UNKNOWN`. `UNKNOWN` is not in `RetryMiddleware.defaultRetryStatuses`, so an overload was never retried - the one error class that most deserves a retry was the one that got none.

The fix maps 529 to `UNAVAILABLE` inside `genkit_anthropic`, not in core. 529 is Anthropic-specific, and core's mapping applies to every plugin. This mirrors the JS plugin, which maps 529 alongside 503 in `js/plugins/anthropic/src/models.ts:205-207`. Every other status still goes through `fromHttpStatus` unchanged.

Tests are in `packages/genkit_anthropic/test/error_mapping_test.dart`. The behavioural one drives `genkit.generate` with the default retry middleware against a mocked 529 followed by a 200, and asserts two HTTP calls and a successful result. It fails on unfixed code (1 call, no retry). Two narrower tests pin the 529 to `UNAVAILABLE` mapping and the unchanged 400 to `INVALID_ARGUMENT` mapping.

Caveats: Retry-After header parsing is deliberately out of scope and stays with #456.

Fixes #355


adesinah (approved):
Changes look good to me.

pavelgj (approved):

